PR #48, opened 2024-03-05
feat: added comboxes aria labels

bashleigh:
# Pull request checklist

fixes: #

### DIFF
--- a/src/components/card/card-components.tsx
+++ b/src/components/card/card-components.tsx
@@ -70,50 +70,44 @@ export interface CardProps extends HTMLAttributes<HTMLDivElement> {
   isSelected?: boolean // Does the card have the blue selected border
 }
 
-export const handleToggleContextMenu = (
-  contextMenuOpen: boolean,
-  setContextMenuOpen: Dispatch<SetStateAction<boolean>>,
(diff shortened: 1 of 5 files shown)


src/components/card/card-components.tsx
```diff
@@ -70,50 +70,44 @@ export interface CardProps extends HTMLAttributes<HTMLDivElement> {
   isSelected?: boolean // Does the card have the blue selected border
 }
 
-export const handleToggleContextMenu = (
```
willmcvay, line 73: Weird formatting - can you run `yarn lint --fix` just to check please

willmcvay (changes requested):
Not sure if it's your editor but loads of files seem to have changed with no changes - can you just check yarn lint is working please

src/components/card/card-components.tsx
bashleigh: Have done. No change 👍🏼

willmcvay: If you look in your node_modules, which version of prettier do you have? I get 3.2.5 - maybe you need to do a yarn install?

willmcvay (approved):
Nice one on working it out, LGTM

PR #49, opened 2024-03-05
feat: unstyled anchor element

bashleigh:
# Pull request checklist

fixes: #


src/styles/unstyled.anchor.ts
```diff
@@ -0,0 +1,16 @@
+import { styled } from '@linaria/react'
+
+export const unstyledAnchor = styled.a`
```
willmcvay, line 3: Firstly would need to be `ElUnstyledAnchor` as it creates a class and all our classes are `el-` prefix.

Secondly, we don't want them to inherit as that comes from the parent. You are using this in combination with another class so might override, might not. Depends on the specificity and when the CSS is loaded. 

Think in this case just easier to copy-paste the styles in TBH, being carful not to override anything already there. Only really being applied to a few components - not worth making re-usable. 

I'd just add to breadcrumb, accordion etc the things you need to stop the anchor froma adding styles and be done with it!

src/components/typography/typography.tsx
```diff
@@ -44,47 +44,49 @@ export interface TypographyProps extends HTMLAttributes<HTMLElement> {
...
-      hasMargin && elHasMargin,
-      hasNoMargin && elHasNoMargin,
-      hasSectionMargin && elHasSectionMargin,
-      hasCenteredText && elHasCenteredText,
-      hasDisabledText && elHasDisabledText,
-      hasCapitalisedText && elHasCapitalisedText,
-      hasUpperCasedText && elHasUpperCasedText,
-      intentClass,
-    ),
+const propsToClass =
```
willmcvay, line 47: Can you run `yarn lint --fix` over the project please? There's some weird formatting going on

src/components/breadcrumb/__styles__/index.ts
```diff
@@ -1,6 +1,7 @@
+import { unstyledAnchor } from '../../../styles/unstyled.anchor'
 import { styled } from '@linaria/react'
 
-export const ElBreadCrumbItem = styled.div`
+export const ElBreadCrumbItem = styled(unstyledAnchor)`
```
willmcvay, line 4: Just do styled.a and override any weird anchor behaviour

src/components/accordion/__styles__/index.ts
```diff
@@ -12,15 +13,14 @@ export const ElAccordionItem = styled.div`
...
-  cursor: pointer;
   padding: 0.875rem 0;
   border-bottom: 1px solid var(--color-grey-100);
 
   &:has(+ .${elIsActive}) {
     border-bottom: none;
   }
 `
-export const ElAccordionTitle = styled.div`
+export const ElAccordionTitle = styled(unstyledAnchor)`
```
willmcvay, line 23: See my comment on unstyledAnchor. I'd just do styled.a and add any overides for the pseudo classes you need

willmcvay (changes requested):

src/styles/unstyled.anchor.ts
```diff
@@ -0,0 +1,16 @@
+import { styled } from '@linaria/react'
+
+export const ElUnstyledAnchor = styled.a`
```
willmcvay, line 3: Can this be removed now?

src/components/accordion/__styles__/index.ts
```diff
@@ -12,15 +12,14 @@ export const ElAccordionItem = styled.div`
...
-  cursor: pointer;
   padding: 0.875rem 0;
   border-bottom: 1px solid var(--color-grey-100);
 
   &:has(+ .${elIsActive}) {
     border-bottom: none;
   }
 `
-export const ElAccordionTitle = styled.div`
+export const ElAccordionTitle = styled.a`
```
willmcvay, line 22: The anchor should be on the accordion item, not on the title - the item is the clickable, not the title

willmcvay (changes requested):
Couple of additional bits, can you also run `yarn lint --fix` there are some weird formatting issues

src/styles/unstyled.anchor.ts
bashleigh: I did wonder why you wanted it if it weren't to be used? Am sure you asked for a component we could extend from?

src/components/typography/typography.tsx
bashleigh: Done. Made no change so will sort manually

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -4,31 +4,32 @@ import { Icon } from '../icon'
...
 
 export const BreadCrumb: FC<BreadCrumbProps> = ({ items, defaultActiveIndex = 0, ...rest }) => {
   const [active, setActive] = useState<number>(defaultActiveIndex)
 
   return (
     <ElBreadCrumbContainer {...rest} aria-label="Breadcrumb">
       {items.map(({ onClick, text }, index) => {
-        if (index > active) return null
+        // Why is this line here and why is it causing a problem now?
+        // if (index > active) return null
```
bashleigh, line 32: Not sure why you added this condition? https://github.com/reapit/foundations/blame/master/packages/elements/src/components/breadcrumb/breadcrumb.tsx#L31

willmcvay: Because if the items are 

Item 1 | item 2 | item 3

And my position on a page is item 2 and therefore the active value is 2, I don't want to render item 3 because that's not where my current navigation position is.

willmcvay: It's probably the story that needs fixing ie set a default active value

willmcvay (changes requested):
Can you please also look at your local linting setup - the workflow run says there are 138 issues in the file https://github.com/reapit/elements/actions/runs/8172238552/job/22342134683

willmcvay (commented):
Just needs a changelog entry then LGTM

willmcvay (approved):
Thanks

Issue #65, opened 2024-03-26
Design Tokens via Style Dictionary and Tokens Studio


lilykirkwood:
**Is your feature request related to a problem? Please describe.**
Design tokens could be leveraged within Elements through the use of [Style Dictionary](https://amzn.github.io/style-dictionary/#/). This could provide a single source of truth for both Designers and Developers, as it can often be difficult to keep both Figma and code in sync. Organising and aliasing variables can become disorganised where they are mixed amongst other code. Additionally, currently within Elements if an intent colour needs to be changed, it must be changed in both the CSS and JS variables.

**Describe the solution you'd like**
My suggestion is that Designers hold the responsibility for creating and maintaining the large majority of tokens within Figma. They could use a tool like [Tokens Studio](https://tokens.studio/) to export these into a JSON format and push to version control. Style Dictionary within the Elements build would then use this/these JSON file(s) to create the CSS & JS variables used within the rest of the code. Style Dictionary allows tokens to be organised in as many files as you wish, within any folder structure, allowing you to choose whatever organisation works best for your team.

**Describe alternatives you've considered**
None. These are the only tools I have considered. There are others out there, but Style Dictionary and Tokens Studio are well known favourites within the Design Tokens community.

**Additional context**
I have a working version of Style Dictionary available in an Elements branch which I can push if this is of interest. Currently it is set up to produce all of the CSS and JS variables already in use within Elements.
I have not experimented with Tokens Studio, but I have read good things in my research, and they are [collaborating with Style Dictionary for v4](https://amzn.github.io/style-dictionary/#/version_4?id=from-the-folks-at-tokens-studio), so I have confidence in the longevity of this pairing of tools as a solution. I understand that Andrei in the design team has successfully used Tokens Studio in a commercial setting previously as well.

kurtdoherty:
**praise:** @lilykirkwood +1 to the idea of using Style Dictionary to enable us to build out multiple formats of the design tokens. Style Dictionary looks valuable to us, so thanks for raising it with the wider group in our previous meetings. Also, your presentation on tokens and Style Dictionary was great; thanks for the sharing that via Teams 🙏 

[...]

PR #73, opened 2024-03-28
feat: style dictionary

lilykirkwood:
fixes: #65 

## Style Dictionary POC

This POC implements Style Dictionary into the Elements build, to produce all of the existing CSS and JS variables from source JSON files, as well as an example of component-level tokens in the Breadcrumb component.

**Walkthrough:**
* Token sources are JSON files, either in `src/tokens/**/*.json` or `src/**/tokens.json`. This is set inside the config file `sd.config.json`.
* `globals.ts` no longer contains tokens, instead:
    * CSS variables are output into `tokens.css`.
    * JS colours are output into `tokens.ts` – imports of these have therefore been updated.
* `yarn build` now builds the dictionary as its first step.
* To build only the dictionary, run `style-dictionary build --config ./sd.config.json` in the project root.
* `base.json` produces all of the variables that were previously in `globals.ts`.
    * Quick evidence of this can be found in [this diff](https://www.diffchecker.com/Ia7lgcZE/) of the alphabetised variables from the old `globals.ts` (left) and the new `tokens.css` (right).
    * I would expect these tokens to be organised into logical groupings within their own files, e.g. color, font, etc.
* The breadcrumb component now has its own `tokens.json` file for its component-level token declarations. I have tokenised all of its properties that I considered should be customisable.
* I acknowledge that there are now a significant number of new JS variables being produced. If this is considered an issue, it can be resolved when the tokens are organised into separate JSON files, by pointing the config for the JS variables at only specific JSON files.


kurtdoherty:
@lilykirkwood has this PR picked up a bunch of other unrelated changes since you first created the branch? There's 524 files here... 😵 Any chance you can do a rebase or something so we can see just the changes relevant to your POC?

kurtdoherty:
@lilykirkwood thanks for the all work you did on this 🙏 

Closing this in favour of #120.

PR #120, opened 2024-06-03
feat: #65 add style dictionary

willmcvay:
# Pull request checklist

fixes: #65 

Based on @lilykirkwood 's initial PR, pulls in tokens from Figma, parses and injects variables into the Linaria globals object.


kurtdoherty:
**question:** @willmcvay are you waiting for this to be reviewed or are there still some things you need to finish off before assigning reviewers?

**note:** I'm happy to review it if you'd like; I'll just need to set aside a block of time to do so (it's a pretty large PR 😅). Feel free to point me at specific files if you think some of the changes don't need to be specifically reviewed 🙏

tokens/build.mjs
```diff
@@ -0,0 +1,84 @@
+import StyleDictionaryPackage from 'style-dictionary'
+const { fileHeader, formattedVariables } = StyleDictionaryPackage.formatHelpers
+
+const themes = ['Reapit', 'PayProp']
+
+const getStyleDictionaryConfig = (theme) => {
+  const lowerCasedTheme = theme.toLowerCase()
+  return {
+    source: ['./tokens/tokens.json'],
+    format: {
```
kurtdoherty, line 10: **suggestion:** It seems like we could we simplify this and just use the built-in CSS format to get a CSS file, e.g. tokens.css, then import that file using a CSS `@import` statement in `globals.ts` 🤔

**note:** It's quite possible I'm missing something with Linaria that prevents this from working; though it seems @import statements are permitted in the global.ts as we're using it for the Source Code Pro font.

tokens/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+          }
+        },
+      },
+    ],
+    platforms: {
+      linaria: {
+        buildPath: `tokens/${lowerCasedTheme}/`,
+        files: [
+          {
+            destination: 'linaria-tokens.ts',
```
kurtdoherty, line 61: **note:** If my suggestion above works and gets implemented, we could avoid the `linaria-` and `typescript-` prefixes on the generated files and just have `tokens.css` and `tokens.ts` files.

tokens/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+      },
+    },
+    parsers: [
+      {
+        pattern: /\.json$/,
+        parse: ({ contents }) => {
+          try {
+            const object = JSON.parse(contents)
+            // Bit fragile but I only want the primitives, semantic and component variables for each theme
+            // Figma exports a "mode 1" that we don't need - confirmed with Andrei
```
kurtdoherty, line 44: **praise:** It may be a bit fragile, but at least is well-contained to this file 👌 👏

package.json
kurtdoherty, line 1: **suggestion:** This can be done separate to this PR, but while I think of it, we'll want to be able to import the typescript tokens from Elements, but I can't see them being exported. I notice that Elements isn't yet using [package exports](https://nodejs.org/api/packages.html#package-entry-points), but adding a package export to allow us to do `import * as tokens from "@reapit/elements/tokens.ts"` should be sufficient for us without bloating the main entry point with stuff that no one else is likely to need.

kurtdoherty (approved):
**praise:** Nice work on this @willmcvay 👏 Will be great to get this in and have our CSS derived from the DS in Figma 🥳 

I just had two suggestions, both of which could be deferred to separate PRs if you'd prefer.

tokens/build.mjs
willmcvay: Nice suggestion, yes, this works - have implemented and was able to get rid of the linaria parse function.

tokens/build.mjs
willmcvay: Agreed, have re-named :-)

willmcvay:
@kurtdoherty Thanks for reviewing, if you could take a look again, would be good to get this merged in. Thanks

package.json
kurtdoherty: Yep, it's only useful for the TS tokens and only while Console Cloud is still using MUI components alongside Elements. Hopefully, there'll be a day in our future where we'll have zero MUI in Console Cloud and no longer need the TS tokens export 🤞

kurtdoherty:
@willmcvay everything looks good to me 👌

willmcvay:
> @willmcvay everything looks good to me 👌 Let's 🚢 it (assuming the "Codacy Static Code Analysis" issues don't need to be dealt with).

Was just playing around with Codacy, it's not a thing we are supporting currently, have removed from the repo. Cheers for the review.

Issue #124, opened 2024-06-24
Spike: V5 format proposal


willmcvay:
**Background context or User story:**

_As discussed with Kurt and Karmen, proposing that we adjust the structure of how the project is presented in Storybook for v5. This ticket is a spike so we have something to discuss but if we like, can use as the model going forward_

**Specification or Acceptance Criteria:**

Broadly, we identified 3 categories of story and the proposal is to group and structure by these flavours:

-  Basic Components (should agree a name, not sure if Atomic is right?) - Presentational wrapped Linaria tags that generate vanilla HTML and CSS, suitable for use as CSS classes in any app (inc non-React). Composable, more un-opinionated - allows maximum flexibility if required.
- Complex Components (Molecules maybe?) - React first composed components - much more opinionated, typically prop driven with advanced functionality driven by React. Underlying markup is often the same, but provides a shorthand abstraction for the happy path in most React apps.
- Patterns - Composed collections of components that validate the design system and provide developers with off the shelf solutions to common UI problems. No new functionality in this section, rather acting as a developer guide on how to use the components effectively.

Currently the Storybook is a jumble of all 3 - suggest we make these distinctions clearer and this spike as a proposal we can discuss, refine and potentially adopt

kurtdoherty:
@willmcvay Thanks for putting this together along with the demonstration in #125 👏 I haven't looked too closely at the file structure in #125 yet, as I'd like to focus first on the definition of "basic components" you've used above.

> Basic Components [snip] - Presentational wrapped Linaria tags...

When I think about designing a component that is made up of smaller parts—like accordions that may consist of an AccordionGroup, Accordion, AccordionHeader and AccordionContent—I want the flexibility to use context, or dynamically determine what HTML element is rendered, within each of those smaller parts so that I provide a simple, minimal and intuitive API for developers to use.

Perhaps I'm missing some aspect of Linaria's API, but it does seem like this would be possible if we were exporting Linaria `styled` components as these "smaller parts" for use by developers. Firstly, Linaria `styled` components seem to be explicitly tied to a specific HTML element (e.g. `styled.button`) and, secondary, don't seem to allow for the styled component to consume from a React context unless one does `styled(AccordionHeader)` (which I _wouldn't_ be in favour of).

As such, I'm concerned about us equating Linaria `styled` components with the "basic" or "atomic" components that we would be exporting for developers to use directly.

[...]

PR #125, opened 2024-06-24
feat: #124 start on spike to get structure

willmcvay:
# Pull request checklist

fixes: #124 

See the above issue for details - the initial refactored components produce the below:

Screenshot 2024-06-24 at 16 41 57


src/components/tabs/composed/tabs.mdx
```diff
@@ -21,42 +18,39 @@ Props available for the React version of the component;
...
+## Tabs - Controlled
 
 Simple example demonstrating how to transition content using tabs. In practice, you may want to dispatch to a router or similar.
 
 The underlying radio buttons are uncontrolled by default so the `isChecked` is used to set the `defaultChecked` in this example. This is to allow the radio to behave nativele i.e. it will update the state using the `onChange` event but not vice versa.
 
-<Canvas of={TabsStories.TabsReactExampleControlled} />
-<RenderHtmlMarkup of={TabsStories.TabsReactExampleControlled} />
+<Canvas of={TabsStories.Controlled} />
+import {BasicUsage} from '../basic/tabs.stories'
```
willmcvay, line 56: Note to myself to remove this line - autocomplete IDE error

adrian-reapit:
I think keeping the docs for a component in one place would be better and make separate usage files. Could have separate Docs as well but can combine. I would normally go straight to the usage example
 
image

willmcvay:
> I think keeping the docs for a component in one place would be better and make separate usage files. Could have separate Docs as well but can combine. I would normally go straight to the usage example
> 
> image

Yep, I think we all agreed that on the call. Will try and find some time for an alternative later in the week

willmcvay:
Screenshot 2024-07-11 at 16 55 36

@kurtdoherty @adrian-reapit have reverted the structure changes based on feedback but kept the division of components, hooks and patterns as above.

Have wrestled with naming conventions around basic vs atoms vs composed vs react for the docs and just reverted to:

"Basic Usage" - these are our atomic components, max flexibility, presentational only
"React Usage" - composed components, happy path, batteries included.

Not very exciting but I honestly can't think of anything more meaningful!

Would appreciate a review on this - assuming happy with basic structure, will press on with getting the specs on the board as a number of the v5 components are good to go in the DS. 

Cheers :-)

willmcvay:
@kurtdoherty Based on the chat in #124, I pushed an additional commit around naming conventions and removing the `__styles__` dir. Would appreciate a review.

src/components/accordion/accordion.tsx
```diff
@@ -1,57 +1,18 @@
...
-}
-
-export const AccordionTitleContent: FC<AccordionBaseProps> = ({ children, ...rest }) => {
-  return <ElAccordionTitleContent {...rest}>{children}</ElAccordionTitleContent>
-}
-
-export const AccordionContent: FC<AccordionBaseProps> = ({ children, ...rest }) => {
-  return <ElAccordionContent {...rest}>{children}</ElAccordionContent>
-}
+} from './styles'
```
kurtdoherty, line 15: **question:** @willmcvay what was your thinking behind leaving these styled components in use instead of the new "atoms" from `accordion.atoms.tsx`? Were you just trying to minimise the amount of changes in the PR?

src/components/accordion/accordion.atoms.tsx
kurtdoherty, line 1: **question:** Do you think grouping the atoms in a single file like this is what we should always do, or do you think its fine for a contributor to separate the atoms out into their own files in cases where they are more than just wrappers around the Linaria styled components?

kurtdoherty (approved):
Looks good to me 👏 Just a couple of questions, mostly for my own clarity on whether you're being prescriptive or not with the arrangement/use of the atoms.

src/components/accordion/accordion.stories.tsx
kurtdoherty, line 1: **question:** Will we always have to group React-based usage examples into a single story? The Console FE team tends to demonstrate key aspects of a component via separate stories (see screenshot below from our storybook). Is this something we'll be able to do when its helpful?

Screenshot 2024-07-17 at 1 29 10 PM

src/components/accordion/accordion.atoms.tsx
willmcvay: Typically the components are pretty simple because they are just wrappers around linaria. It's not something I feel super strongly about, I guess 1 component per file just gets a bit noisy for no real benefit but I don't have a problem if people feel differently.

src/components/accordion/accordion.stories.tsx
willmcvay: So, it's just a preference, but I would encourage use of controls for atomic components to allow people to play with the component in real time.

Example using a button story:

```ts
// Writing a story like this will render a story called intent but I CAN'T use controls
export const Intent = {
  render: ({}) => <Button intent="primary">Button Text</Button>,
}
```

```ts
// Whereas using the args object, I can edit away in real time using any of the props
export const Intent = {
  args: {
    intent: 'primary',
    children: 'Button Text',
  },
}
```
In the screenshot below, I used the second story structure and I get a nice drop down to play with my component, adjusting any prop as a I choose. 

Screenshot 2024-07-17 at 15 13 26

I guess my point is we can have fewer stories if we use controls and allow people to adjust props as they like.

The nuance - this falls down when the story is not a single atom like a ButtonGroup, containing multiple Buttons, or when you pass in an array of objects as a prop, like an Accordion. In this case, you will need to use the `render` function for the story and yes, I would then use your above approach to show the various states.

kurtdoherty: Yes, absolutely agree on the use of controls. There's a few reasons we split out specific stories for our components:

1. We use individual stories to demonstrate the same examples Design show for components in Figma. This helps us ensure we've covered all the same variants/behaviour etc that Design have specified;
2. We use individual stories as the basis for our unit tests. This allows us to avoid duplicating the various prop combinations/scenarios across the story file and our unit tests. Instead, our unit test imports a story and then exercises it (either via a simple snapshot, or through interaction);
3. We use the built-in "docs page" feature that Storybook comes with (i.e. not MDX pages). We loose the flexibility to format the docs page however we like, but it gives us immediate consistency across all docs pages for little effort, and still allows us to provide specific descriptions per story that can further explain the different behaviours/examples.

Feel free to take a look at our [Storybook](https://app.console.com.au/client-apps/design-system/react-ui-core/?path=/docs/forms-button--docs) to see what I mean by (1) and (3). In practice, we actually haven't done (2) particularly well, but it still represents our intention.

Issue #136, opened 2024-07-30
v5 component refactor: TextArea


kurtdoherty:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_.

The existing `TextArea` component in Elements does not currently meet all aspects of the Design System's TextArea specification.

**Specification**

- Is new component or existing? **Existing**
- Is the component being re-named? **No**
- Is the component design signed off by product and design? **Unknown**
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-9227&t=3FvWMbhNKUniX3TP-4)
- [Link to Elements Audit in Figma](https://www.figma.com/design/J6EneNGfpk2GcUuymiOKW4/Elements-Gaps-%2B-Discrepancies?node-id=319-6548&t=rBEuqtRqTtxyBUpo-4)
- Additional behavioural specification: **Not available yet**
- Accessibility requirements: **Not applicable, because we're using the standard `<textarea>` HTML element**

**Developer Checklist**

- [x] Styles alignment between Design System and Elements
- [x] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [x] Align with accessibility standards / spec as per above
- [x] If relevant, break down component into `Styles Only` and `React` component structures
- [x] Ensure all variants of components are documented as appropriate 
- [x] Ensure unit test coverage is adequate for component
- [x] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

**note:** One of the (currently undocumented) behaviours the DS expects of a text area is the ability for it to resize itself between some defined minimum and maximum number of rows. #134 is seeking clarification on whether we are willing to conditionally leverage an experimental CSS property, `field-sizing`, to enable this resizing behaviour to be present for CSS-only consumers on specific browsers.

willmcvay:
@kurtdoherty I am totally fine on using modern CSS features - we have auto-prefixer in the build to handle browser quirks - the only issue will be our friend Safari - we have a lot of mobile iOS users so as long as there is a graceful fallback, it's all good.

[...]

PR #150, opened 2024-08-01
refactor: #136 TextArea - Add CSS resizing

kurtdoherty:
### Context

- #136 describes updates needed for the TextArea component to align with the Design System;
- The key breaking change is that the TextArea can no longer be manually resized; rather, it should grow/shrink based on the amount of content entered.
- I am leveraging an experimental CSS property, [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) to achieve the resizing behaviour for CSS-only and React consumers on Chrome and Edge browsers, with a fallback to a JS-based resizing solution for React-only consumers on other browsers.
- This work is split across two PRs:
  - 👉 **Part 1: Add CSS resizing**
  - #151 

### This PR (Part 1)

- Moves `__styles__/index.ts` to `styles.ts` per the [project structure conventions](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide#project-structure-conventions);
- Updates CSS for `ElTextArea` styled component to:
  - Use the newly available design tokens
  - Expand the invalid state styling too be used either when either the `elTextAreaHasError` class name is applied, or the `:invalid` or `:user-invalid` pseudo-classes are active;
  - Provide styling for `:disabled` and `:read-only` states;
  - Provide max and min height constraints based on CSS variables that CSS-only consumers can easily override;
- Updates `ElTextArea` to leverage experimental field-sizing property if supported;
- Updates Storybook docs to demonstrate basic usage, invalid state, and resizing behaviour;
- Updates TextArea and its prop interface to include minRows, maxRows and rows.

textarea p1

**note:** The design guidelines have not be finalised for this component, so there may be some additional changes once they are, particularly the treatment of default min/max rows values or the TextArea being resizable by default as opposed to fixed size by default.

**note:** This PR is on the large size, so if you'd like me to split it up further I can. In hindsight, the Storybook updates could have been left until later 😅


src/components/form-layout/form-layout.stories.tsx
```diff
@@ -77,7 +77,7 @@ export const BasicForm = {
         </InputWrap>
         <InputWrapFull>
           <InputGroup>
```
kurtdoherty, line 79: **note:** `type` is no longer an accepted prop for TextArea, so fixing up these stories.

src/components/input-group/__styles__/index.ts
```diff
@@ -2,7 +2,7 @@ import { styled } from '@linaria/react'
 import { ElIcon } from '../../icon/__styles__'
 import { ElLabel } from '../../label/__styles__'
 import { ElInput, elHasInputError } from '../../input/__styles__'
```
kurtdoherty, line 4: **note:** Fixing up to import from the root `textarea` folder instead of relying on internal file structures.

src/components/textarea/__styles__/index.ts
kurtdoherty, line 1: **note:** Moved to `textarea/styles.ts`

src/components/textarea/__tests__/__snapshots__/textarea.test.tsx.snap
kurtdoherty, line 1: **note:** The current approach of snapshotting the `render` result is not ideal. It means we are capturing a host of React Testing Library's implementation details which are irrelevant to our component unit tests. As you'll see in the unit test file, I've updated the unit tests to snapshot the document fragment that contains our rendered component.

**aside:** It may be helpful to add this as a task to all the v5 refactor issues so that devs can remember to update the tests.

src/components/textarea/__tests__/textarea.test.tsx
```diff
@@ -1,9 +1,19 @@
-import { render } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 import { TextArea } from '..'
```
kurtdoherty, line 3: **note:** I've updated the unit tests to avoid the generic "should match a snapshot". While this kind of unit test provides _some_ value, it's missing out (IMO) on one of the most important aspects of unit tests: _documentation_. As such, I've replaced it with three tests that cover some of the important aspects we would want to document about our TextArea, though one is currently skipped (see code comment).

**aside:** I would not say that these unit tests are hitting the bar of "necessary and sufficient" by themselves, but the addition of the visual snapshots helps get us pretty close as they provide coverage of the invalid, disabled and read-only states, plus the initial sizing of text areas based on the `minRows`, `maxRows` and `rows` props.

I can provide more info on what the terms "necessary and sufficient" mean separately (they are terms we recently introduced to the Console FE team).

src/components/textarea/styles.ts
```diff
@@ -0,0 +1,79 @@
...
+import { styled } from '@linaria/react'
+
+export const elTextAreaHasError = css``
+
+export interface TextAreaCSSProperties {
+  '--textarea-max-rows': number
+  '--textarea-min-rows': number
+}
+
+export const ElTextArea = styled.textarea`
```
kurtdoherty, line 11: **note:** The code comment below highlights an important decision I made in the implementation of this component. Instead of relying on Linaria's ability to convert `${({ maxRows }) => ...}` into a randomly named CSS variable for us, I've opted to handle it manually in order to provide well-named CSS variables that CSS-only consumers can leverage.

I would have preferred to rely on data attributes on the textarea, but there's currently no way for CSS to use the value of a data attribute. See [attr()](https://developer.mozilla.org/en-US/docs/Web/CSS/attr) for more information.

src/components/textarea/styles.ts
```diff
@@ -0,0 +1,79 @@
...
+  // NOTE: These are public CSS variables that we use to allow CSS-only consumers
+  // to define the maximum and minimum rows the text area should resize between.
+  // We do this ourselves instead of leveraging Linaria's dynamic styles because
+  // that approach results in randomly-named CSS variables, which would not provide
+  // a friendly interface for CSS-only consumers.
+  --textarea-max-rows: infinity;
+  --textarea-min-rows: 2;
+
```
kurtdoherty, line 19: **note:** The code comment below is another important decision. While I strongly believe there should be _zero_ developer-defined, component-specific design tokens in Elements, I don't think that should prevent us from leveraging CSS variables in a "private" manner to aide the definition of our styles.

In order to try and communicate they are "private"/"implementation detail" CSS variables, I've prefixed them with the standard double underscore, `__`. Not sure if that's something we want to do or not.

If we'd prefer not to do this at all, I can remove these and just use the real design tokens directly in all the places they're referenced.

src/components/textarea/styles.ts
```diff
@@ -0,0 +1,79 @@
...
+  resize: none;
+  width: 100%;
+
+  // NOTE: field-sizing property is currently experimental
+  // @see https://drafts.csswg.org/css-ui/#field-sizing
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing
+  @supports (field-sizing: content) {
+    field-sizing: content;
+  }
+
```
kurtdoherty, line 45: **note:** The code comment below highlights an important design decision I've made with the max and min height constraints for the text area. Since we want CSS-only consumers to get the correct resizing behaviour, we need to facilitate the max/min constraints in CSS, not JS.

I've used the min and max `block-size` CSS properties as they account for the consumer's [writing-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode).

src/components/textarea/styles.ts
```diff
@@ -0,0 +1,80 @@
...
+  // We use CSS to calulate the max/min block size of the text area based on the max/min number of rows
+  // that have been specified. We use block-size instead of height because the latter is not writing-mode aware.
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/max-block-size
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/min-block-size
+  max-block-size: calc(
```
kurtdoherty, line 51: **note:** We use the new `lh` unit to calculate the length of the rows, then add on the padding and border. See [MDN's docs on CSS values and units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units) for more information on `lh`, but the TL;DR is:

> `lh` and `rlh` are relative lengths units similar to `em` and `rem`. The difference between `lh` and `rlh` is that the first one is relative to the line height of the element itself, while the second one is relative to the line height of the root element, usually `<html>`.

src/components/textarea/styles.ts
```diff
@@ -0,0 +1,80 @@
...
+  );
+  min-block-size: calc(
+    1lh * var(--textarea-min-rows) + 2 * (var(--__textarea-padding-y) + var(--__textarea-border-width))
+  );
+
+  &:focus {
+    outline: none;
+    border-color: var(--outline-text_input-focus);
+  }
+
```
kurtdoherty, line 62: **note:** the original Elements text area did not have styles for disabled or read-only states. These have been added here. The Design System does not currently differentiate visually between disable and read-only; this may change in future.

src/components/textarea/styles.ts
```diff
@@ -0,0 +1,80 @@
...
+  &:read-only,
+  &:disabled {
+    background-color: var(--fill-default-lightest);
+  }
+
```
kurtdoherty, line 67: **note:** The original text area only applied the invalid state styles when the `elTextAreaHasError` class was present. This has been expanded to include the `:invalid` and `:user-invalid` pseudo-classes.

src/components/textarea/textarea.mdx
kurtdoherty, line 1: **note:** Some minor updates here to include the extra stories that have been added.

**aside:** I'm keen to understand why Elements prefers MDX files over Storybook's autogenerated DocsPage functionality. What I'm doing here (e.g. the use of the `Description` component to render the JS Doc comments from the component or a specific story) is all handled by Storybook's DocsPage feature, so I'm not sure what value we are getting from doing this ourselves 🤔

src/components/textarea/textarea.stories.tsx
kurtdoherty, line 1: **note:** I've expanded the number of stories being demonstrated here, as well as applied Storybook's types to the exports. The JS Doc for each story is picked up by Storybook's `Description` component, which is now in use in `textarea.mdx`.

src/components/textarea/textarea.tsx
```diff
@@ -1,22 +1,70 @@
...
 import { cx } from '@linaria/core'
+import { ElTextArea, elTextAreaHasError, type TextAreaCSSProperties } from './styles'
+import { forwardRef } from 'react'
 
-export interface TextAreaProps extends InputHTMLAttributes<HTMLTextAreaElement> {
+import type { CSSProperties, TextareaHTMLAttributes } from 'react'
+
+// NOTE: We omit the `cols` prop because our text area should always grow to the width of its container.
+interface RestrictedTextareaHTMLAttributes extends Omit<TextareaHTMLAttributes<HTMLTextAreaElement>, 'cols'> {}
+
```
kurtdoherty, line 9: **note:** I've updated how the component's prop interface is defined for two reasons:

1. To add the new `minRows` and `maxRows` props; and,
2. To use the `TextareaHTMLAttributes` instead of `InputHTMLAttributes` (the latter one is for `<input>` elements).

src/components/textarea/textarea.tsx
```diff
@@ -1,22 +1,70 @@
...
-  )
-})
+/**
+ * An (almost) standard HTML/JSX `<textarea>` for use in forms.
+ *
+ * Can automatically resize itself between a minimum and/or maximum number of lines of text (rows). This
+ * resizing behaviour is available for CSS-only consumers on Chrome and Edge. For browsers that do not yet
+ * support the [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) property, we
+ * fallback to a JS-based resizing solution that is only available to React-based consumers.
+ */
```
kurtdoherty, line 47: **note:** I've opted to use forwardRef's generic type slots instead of the original `TextAreaWrapped` type. This makes for a simpler approach and ensures our forwarded ref is correctly typed.

src/components/textarea/textarea.tsx
```diff
@@ -1,22 +1,70 @@
...
+      <ElTextArea
+        {...rest}
+        className={cx(hasError && elTextAreaHasError, className)}
+        defaultValue={defaultValue}
+        style={
+          {
+            // NOTE: `rows` takes precedence. If it's defined, the text area is effectively fixed-height.
+            // If it's not defined, we allow resizing to occur between the min and max rows.
+            '--textarea-max-rows': rows ?? maxRows,
+            '--textarea-min-rows': rows ?? minRows,
```
kurtdoherty, line 60: **note:** I'm using `satisfies TextAreaCSSProperties` (a type which comes from our `styles.ts` module) in order to ensure we don't provide more styles here than what we explicitly permit through the `TextAreaCSSProperties`. I then assert this to be the `CSSProperties` type in order to satisfy React/Linaria's expectations (they don't consider the CSS variables to be valid property names).

src/components/form-layout/form-layout.stories.tsx
willmcvay: Yep, agreed, there's no need

src/components/textarea/__tests__/__snapshots__/textarea.test.tsx.snap
```diff
@@ -1,72 +1,19 @@
...
-  "queryByLabelText": [Function],
-  "queryByPlaceholderText": [Function],
-  "queryByRole": [Function],
-  "queryByTestId": [Function],
-  "queryByText": [Function],
-  "queryByTitle": [Function],
-  "rerender": [Function],
-  "unmount": [Function],
-}
+exports[`can set custom min/max rows 1`] = `
```
willmcvay, line 3: Much cleaner without the React internals :-)

src/components/textarea/__tests__/textarea.test.tsx
willmcvay: If you have some docs on this, would be more than happy to add to the wiki

src/components/textarea/styles.ts
willmcvay: Yep, makes sense - appreciate the comments in here too

src/components/textarea/styles.ts
willmcvay: It does look a bit odd syntactically but the comment explains fine. I would suggest as long as they are outside the root scope it's probably fine.

src/components/textarea/styles.ts
willmcvay: I did just check this out in Safari just because Safari and couldn't see any issues FYI

src/components/textarea/textarea.stories.tsx
```diff
@@ -1,19 +1,98 @@
+import { Meta, StoryObj } from '@storybook/react'
 import { TextArea } from './index'
 
 export default {
   title: 'TextArea',
```
willmcvay, line 5: Can you add the `Components` prefix here i.e. `Components/TextArea` please so it moves to the relevant folder in Storybook

willmcvay (approved):
Hi @kurtdoherty so looks good to me - have tested locally in Safari, FF & Chromium because of some of the edgier properties and works as expected which is great - will check out the JS fallback next.

Comments are appreciated for the more non-vanilla implementations - I appreciate, it's a tricky bit of code.

Just that small issue with the stories file to move to the new structure in SB but happy if you want to move that to the second PR.

src/components/textarea/styles.ts
kurtdoherty: @willmcvay I agree the syntax looks weird. We could choose to use `private` as a prefix? i.e. `--private-textarea-border-width` 🤔

src/components/textarea/textarea.stories.tsx
kurtdoherty: - [ ] Add `Components/` prefix to TextArea story title

kurtdoherty:
> With the other PR this is quite an update on a normal TextArea and if its got bespoke 'tricky' code it might better to make a new component. MaterialUI has a TextareaAutosize component, name makes clear what you are getting https://mui.com/base-ui/react-textarea-autosize/

@adrian-reapit I did consider this originally, but decided to take the approach I've presented in these PRs because:
- We've previously discussed as a DS working group that we want to _aim_ for a 1-1 equivalency between Elements and the Design System. While we shouldn't blindly adhere to that in every case, I do think it makes sense for TextArea.
- I view the JS-based fallback for resizing to be (relatively) short-lived. Once `field-sizing` support is rolled out for Safari (it's [already](https://webkit.org/css-status/?#?search=field-sizing) [in progress](https://github.com/WebKit/WebKit/commit/bcca5fc0fa1715d7da3e05cac22fb460a12539f0)) and Firefox (its [on their radar](https://bugzilla.mozilla.org/show_bug.cgi?id=1832409)), all the "tricky" code will disappear.
- Once the JS-based resizing logic is removed, the idea of having two text area components whose only difference is the `field-sizing` CSS property doesn't seem ideal.
- I really value the DX of seeing a design use a TextArea component and being able to reach for a component with the same name and the same props.

willmcvay:
@kurtdoherty @adrian-reapit Just to resolve the slight log-jam, why not take the "Styles Only" usage approach and export a wrapped version of `ElTextArea` like we do with the `Accordion` atoms. Then people can choose between batteries included or simple HTML eg https://github.com/reapit/elements/blob/main/src/components/accordion/accordion.atoms.tsx#L12

kurtdoherty:
@willmcvay we already are exporting the basic styled component, `ElTextArea`. You can see this happening in `TextArea/styles.ts` and `TextArea/index.ts`. I'm not sure there's much value in exporting another component that simply wraps it? 🤔

kurtdoherty:
**note:** I have a meeting with Andrei today to talk through the component interface, particularly around what design would like to see as the properties for their TextArea in Figma and how that works for us, so there may be some minor changes land after that conversation.

willmcvay:
> @willmcvay we already are exporting the basic styled component, `ElTextArea`. You can see this happening in `TextArea/styles.ts` and `TextArea/index.ts`. 👍
> 
> I'm not sure there's much value in exporting another component that simply wraps it? The atom would want to be called TextArea, but that's the name of the "main" component too. This isn't a problem with the Accordion atoms because they're all uniquely named 🤔

Will leave up to you. IMO, I am in favour of the implementation you presented - was just a suggestion to alleviate any concerns Adrian had. Will leave you to guys to agree a preference. If the auto-resizing behaviour is what the design team want, I am very much on board with delivering this.

kurtdoherty:
**question:** @adrian-reapit just checking in to see if your concerns about the approach taken in this PR and #151 are "blocking", or if you're okay for me to proceed? If they're blocking, I'd like to line up a video call for us to chat through our options as soon as possible so we can get this work finished off and merged in 🙏

kurtdoherty:
From @adrian-reapit via a Teams chat:

> Its not blocking but iIs a noticeable change.. I would just give the choice until it had full browser support and didn't need a fallback. We have 12 uses of textarea in our application which will need to be checked so that min/max are set correctly and our site is all mobile usage so the 'fallback' is the component. Everyone updating to v5 will need to do the same on top of all other changes. testers would need to know that the desktop version is not the same as mobile. Happy to discuss if you want

A few comments:
- The Design System wants text area's to have a minimum of 3 rows but no maximum number of rows, by default. This means those 12 text areas would largely behave the same as they do now, only they'll resize themselves, not allow the user to do that themselves. Conceptually, this seems like a pretty safe change for those 12 text areas, but I acknowledge there may be nuances to those 12 uses that I'm not aware of that make this change more concerning.
- Off the back of a conversation with Andrei, I need to tweak the component interface by adding a new prop, `fieldSizing` with two values `fixed` and `content`. While `content` would be the default value for this prop, we could add a third, deprecated option, `manual`, that would restore the manual resizing behaviour the v4 text area currently has. This would mean v4 consumers that want to maintain this existing behaviour would just need to add `fieldSizing="manual"`. Would this be reasonable?

kurtdoherty:
From @adrian-reapit via Teams chat:

> That suggestion would be ok. Thanks

PR #151, opened 2024-08-02
refactor: #136 TextArea - Add JS fallback for resizing

### LINKED ISSUE #136: v5 component refactor: TextArea

kurtdoherty:
### Context

- #136 describes updates needed for the TextArea component to align with the Design System;
- The key breaking change is that the TextArea can no longer be manually resized; rather, it should grow/shrink based on the amount of content entered.
- I am leveraging an experimental CSS property, [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) to achieve the resizing behaviour for CSS-only and React consumers on Chrome and Edge browsers, with a fallback to a JS-based resizing solution for React-only consumers on other browsers.
- This work is split across two PRs:
  - #150 
  - 👉 **Part 2: Add JS fallback for resizing**

### This PR (Part 2)

Resolves #136 

- Adds a JS-based fallback for resizing the text area that we use when the [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) CSS property is not supported;
- Adds a new styled component, `ElShadowTextArea`, that is used to calculate the height that the visible text area should be given;
- Adds two new hooks that take care of resizing the text area:
  - `useResizeTextAreaEffect` is an effect that handles resizing on initial mount of the component, and whenever its `value` prop changes (i.e. when the text area is a controlled component);
  - `useResizeTextAreaOnChange` provides a higher-order function that is used to wrap the text area's `onChange` handler. This is necessary when the text area is an uncontrolled component.
- Updates `TextArea` to conditionally render a "shadow" text area and wires up the new hooks to both text area elements so that we can "sync" the height from the shadow text area to the primary text area;
- Adds a new `mergeRefs` utility to allow multiple refs to be attached to the text area (a forwarded ref and an internal ref).

textarea p2

**note:** The GIF above shows Storybook being used via Safari, which does not currently support the `field-sizing` CSS property. Despite that, you can see that the resizing behaviour is still functioning as expected.

**note:** The design guidelines have not be finalised for this component, so there may be some additional changes once they are, particularly the treatment of default min/max rows values or the TextArea being resizable by default as opposed to fixed size by default.

### DIFF
--- a/src/components/textarea/__tests__/__snapshots__/textarea.test.tsx.snap
+++ b/src/components/textarea/__tests__/__snapshots__/textarea.test.tsx.snap
@@ -1,5 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`content field-sizing a shadow text area is present when the CSS \`field-sizing\` property is not supported 1`] = `
+<DocumentFragment>
+  <mock-styled.textarea
+    classname=""
+    data-field-sizing="content"
+    style="--textarea-max-rows: Infinity; --textarea-min-rows: 3; block-size: 0px;"
+  />
+  <mock-styled.mock-styled.textarea
+    aria-hidden="true"
+    data-field-sizing="content"
+    style="--textarea-max-rows: Infinity; --textarea-min-rows: 3;"
+  />
+</DocumentFragment>
+`;
+
 exports[`content field-sizing can set custom min/max rows 1`] = `
 <DocumentFragment>
   <mock-styled.textarea
--- a/src/components/textarea/__tests__/textarea.test.tsx
+++ b/src/components/textarea/__tests__/textarea.test.tsx
@@ -1,7 +1,14 @@
+import isCSSContentFieldSizingSupported from '../is-css-content-fieldsizing-supported'
 import { render, screen } from '@testing-library/react'
 import { TextArea } from '../textarea'
 
+jest.mock('../is-css-content-fieldsizing-supported', () => jest.fn())
+
 describe('content field-sizing', () => {
+  beforeEach(() => {
+    jest.mocked(isCSSContentFieldSizingSupported).mockReturnValue(true)
+  })
+
   test('default min/max rows are 3 -> Infinity', () => {
     const { asFragment } = render(<TextArea fieldSizing="content" />)
     expect(asFragment()).toMatchSnapshot()
@@ -9,6 +16,13 @@ describe('content field-sizing', () => {
 
   test('can set custom min/max rows', () => {
     const { asFragment } = render(<TextArea fieldSizing="content" maxRows={10} minRows={3} />)
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  test('a shadow text area is present when the CSS `field-sizing` property is not supported', () => {
+    jest.mocked(isCSSContentFieldSizingSupported).mockReturnValue(false)
+
+    const { asFragment } = render(<TextArea fieldSizing="content" />)
     expect(asFragment()).toMatchSnapshot()
   })
 })
@@ -39,7 +53,14 @@ describe('manual field-sizing', () => {
 
 // TODO: This test is currently skipped because our Linaria styled global mock
 // is changing the tag name.
-test.skip('is accessible via the textbox role', () => {
-  render(<TextArea fieldSizing="content" maxRows={10} minRows={3} />)
+test.skip('is always accessible via the textbox role', () => {
+  jest.mocked(isCSSContentFieldSizingSupported).mockReturnValue(true)
+
+  const { rerender } = render(<TextArea fieldSizing="content" maxRows={10} minRows={3} />)
+  expect(screen.getByRole('textbox')).toBeDefined()
+
+  jest.mocked(isCSSContentFieldSizingSupported).mockReturnValue(false)
+
+  rerender(<TextArea fieldSizing="content" maxRows={10} minRows={3} />)
   expect(screen.getByRole('textbox')).toBeDefined()
 })
--- a/src/components/textarea/styles.ts
+++ b/src/components/textarea/styles.ts
@@ -78,6 +78,15 @@ export const ElTextArea = styled.textarea<ElTextAreaProps>`
     }
   }
 
+  &[data-field-sizing='fixed'] {
+    // NOTE: field-sizing property is currently experimental
+    // @see https://drafts.csswg.org/css-ui/#field-sizing
+    // @see https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing
+    @supports (field-sizing: fixed) {
+      field-sizing: fixed;
+    }
+  }
+
   &[data-field-sizing='manual'] {
     resize: both;
   }
@@ -116,4 +125,14 @@ export const ElTextArea = styled.textarea<ElTextAreaProps>`
     font-family: var(--font-sans-serif);
     font-size: var(--font-size-small);
   }
+`
+
+export const ElShadowTextArea = styled(ElTextArea)`
+  position: absolute;
+  height: 0;
+  left: 0;
+  overflow: hidden;
+  top: 0;
+  transform: translateZ(0);
+  visibility: hidden;
 `
--- a/src/helpers/mergeRefs.ts
+++ b/src/helpers/mergeRefs.ts
@@ -1,0 +1,23 @@
+import type { MutableRefObject, Ref, RefCallback } from 'react'
+
+/**
+ * Assigns a value to a ref function or object
+ */
+function assignRef<T = unknown>(ref: Ref<T> | RefCallback<T>, value: T) {
+  if (typeof ref === 'function') {
+    ref(value)
+  } else if (ref) {
+    ;(ref as MutableRefObject<T>).current = value
+  }
+}
+
+/**
+ * Combine multiple React refs into a single ref function. This is used mostly when you
+ * need to allow consumers to forward refs to an internal component that is already
+ * receiving a ref.
+ */
+export default function mergeRefs<T = unknown>(...refs: Array<Ref<T> | RefCallback<T>>): RefCallback<T> {
+  return (node) => {
+    refs.forEach((ref) => assignRef(ref, node))
+  }
+}
--- a/src/storybook/changelog.mdx
+++ b/src/storybook/changelog.mdx
@@ -16,6 +16,10 @@ We will publish release version history and changes here. Where possible, we wil
 
 Beta versions should be relatively stable but subject to occssional breaking changes.
 
+### **5.0.0-beta.14 - 06/08/24**
+
+- **Breaking:** `TextArea` no longer supports manual resizing; instead, it can automatically resize between a min and max row count.
+
 ### **5.0.0-beta.12 - 24/06/24**
 
 - Adds code to import Figma design tokens into project
@@ -511,4 +515,4 @@ We hope the migration process will be relatively smooth in virtually all cases h
 ### **3.0.0-beta.1 - 05/07/21**
 
 - Initial 3.x.x full ground up re-build with zero compat from v2.x.x
-- For migration information see aboveimport { text } from 'stream/consumers'
+- For migration information see above
--- a/src/components/textarea/__tests__/use-resize-textarea-effect.test.tsx
+++ b/src/components/textarea/__tests__/use-resize-textarea-effect.test.tsx
@@ -1,0 +1,45 @@
+import { render, waitFor } from '@testing-library/react'
+import syncTextAreaHeight from '../sync-textarea-height'
+import { useRef } from 'react'
+import useResizeTextAreaEffect from '../use-resize-textarea-effect'
+
+jest.mock('../sync-textarea-height')
+
+beforeEach(() => {
+  jest.mocked(syncTextAreaHeight).mockClear()
+})
+
+test('does NOT sync height when `isEnabled` is false', async () => {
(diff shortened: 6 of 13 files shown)


src/components/textarea/__tests__/__snapshots__/textarea.test.tsx.snap
```diff
@@ -1,5 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
kurtdoherty, line 2: **note:** New snapshot test to assert that we render two text area's whenever the CSS field-sizing property is not supported.

src/components/textarea/__tests__/textarea.test.tsx
kurtdoherty, line 1: **note:** Updating the existing unit tests to mock the result of our new `isCSSFieldSizingSupported` function + adding a new test and enhancing the (still skipped) a11y test.

src/components/textarea/__tests__/useResizeTextAreaEffect.tsx
kurtdoherty, line 1: **note:** New unit tests for the `useResizeTextAreaEffect` hook that has been introduced. These tests are what I'd call a "sociable" test (we're actually rendering the hook in a component) even though I'm replacing the `syncTextAreaHeight` with a test double (so we can assert it is called when we expect it to be).

src/components/textarea/__tests__/useResizeTextAreaOnChange.tsx
kurtdoherty, line 1: **note:** New unit tests for the `useResizeTextAreaOnChange` hook that has been introduced. Like the tests for the other hook, I'd call these "sociable" tests. I still replace `syncTextAreaHeight` with a test double (so we can assert it is called when we expect it to be) and I fire some real change events to confirm what we expect to happen actually happens.

src/components/textarea/styles.ts
```diff
@@ -78,3 +78,13 @@ export const ElTextArea = styled.textarea`
     font-size: var(--font-size-small);
   }
 `
```
kurtdoherty, line 80: **note:** New styled component for our "shadow" text area. Importantly, this component is visually hidden in a number of different ways to ensure users will never see it. When I use it, I also hide it from an a11y perspective via `aria-hidden`.

src/components/textarea/syncTextAreaHeight.ts
kurtdoherty, line 1: **note:** New helper function that grabs the scroll height from the `from` text area (our "shadow" text area) and writes it to the `to` text area's (our visible text area) `blockSize` style property.

src/components/textarea/textarea.tsx
```diff
@@ -47,24 +51,66 @@ export interface TextAreaProps extends RestrictedTextareaHTMLAttributes {
  */
 export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
   ({ className, defaultValue, hasError, maxRows = Infinity, minRows = 2, onChange, rows, value, ...rest }, ref) => {
```
kurtdoherty, line 53: **note:** Here I am wiring up the new hooks with the primary text area and the newly added shadow text area. Important things to note include:

- `shouldResizeViaJS` determines whether the CSS field-sizing property is supported and if so, we will not do any JS resizing. Also, we only resize via JS when `rows` is undefined.
- Both hooks have an `isEnabled` param that effectively "turns off" the resizing logic they otherwise facilitate. We have to do it this way because we can't conditionally use hooks.
- We also use `shouldResizeViaJS` to determine if we render the shadow text area or not.

src/components/textarea/textarea.tsx
```diff
@@ -47,24 +51,66 @@ export interface TextAreaProps extends RestrictedTextareaHTMLAttributes {
...
+          defaultValue={defaultValue}
+          style={
+            {
+              // NOTE: `rows` takes precedence. If it's defined, the text area is effectively fixed-height.
+              // If it's not defined, we allow resizing to occur between the min and max rows.
+              '--textarea-max-rows': rows ?? maxRows,
+              '--textarea-min-rows': rows ?? minRows,
+            } satisfies TextAreaCSSProperties as CSSProperties
+          }
+          onChange={decorateOnChange(onChange)}
```
kurtdoherty, line 89: **note:** here is where we use the new `mergeRefs` function. This allows us to "merge" two refs together to ensure they both receive a handle on the underlying DOM element.

src/helpers/mergeRefs.ts
kurtdoherty, line 1: **note:** Not sure if this is the right place for this new helper. At this stage, TextArea is the only component that needs it, so it could be local to that component, but in my experience with the Console component lib, its likely we'll discover other uses for it in other components.

src/components/textarea/useResizeTextAreaEffect.ts
kurtdoherty, line 1: **note:** One the most important aspects of the design I've implemented here lies with this hook, `useResizeTextAreaEffect`, and it's sibling, `useResizeTextAreaOnChange`.

We have two hooks because we have two different concerns: controlled usage and uncontrolled usage.

- **When the text area is controlled**, the source of truth for its value comes from a parent React component. This means:
  - We can have both the primary and the shadow text areas controlled by the same `value`;
  - We can use a rerender of the component as a signal to update the height of the primary text area;
  - We should not use the change event target's value at all. That is the purview of the parent that responds to the change event and updates (or not) state;
- **When the text area is uncontrolled**, the source of truth for its value comes from the DOM. This means:
  - We have to use change events as our signal for updating the height of our primary text area.
  - We have to manually sync the value from our primary text area to the shadow text area. This is done by `useResizeTextAreaOnChange`.

src/components/textarea/useResizeTextAreaEffect.ts
```diff
@@ -0,0 +1,29 @@
...
+  textAreaRef: RefObject<HTMLTextAreaElement>
+  value?: unknown
+}
+
+export default function useResizeTextAreaEffect({
+  isEnabled,
+  shadowTextAreaRef,
+  textAreaRef,
+  value,
+}: UseResizeTextAreaEffectConfig): void {
```
kurtdoherty, line 18: **note:** I'm using a layout effect instead of a standard effect because we're updating the DOM and we (probably) want the change to be flushed to the DOM before React's render phase completed.

src/components/textarea/useResizeTextAreaOnChange.ts
```diff
@@ -0,0 +1,31 @@
...
+type UseResizeTextAreaOnChangeDecorator = (onChange?: ChangeEventHandler) => ChangeEventHandler
+
+export default function useResizeTextAreaOnChange({
+  isEnabled,
+  shadowTextAreaRef,
+  textAreaRef,
+}: UseResizeTextAreaOnChangeConfig): UseResizeTextAreaOnChangeDecorator {
+  return function decorateTextAreaChangeHandler(onChange?: ChangeEventHandler) {
+    return function syncTextAreaHeightAfterOnChange(event: ChangeEvent) {
+      onChange?.(event)
```
kurtdoherty, line 20: **note:** As the code comment below points out, we allow the consumer-supplied `onChange` handler to call `event.preventDefault()` to stop our resizing behaviour from happening.

src/components/textarea/__tests__/textarea.test.tsx
```diff
@@ -1,19 +1,40 @@
...
+
   const { asFragment } = render(<TextArea maxRows={10} minRows={3} />)
   expect(asFragment()).toMatchSnapshot()
 })
 
 // TODO: This test is currently skipped because our Linaria styled global mock
 // is changing the tag name.
-test.skip('is accessible via the textbox role', () => {
-  render(<TextArea maxRows={10} minRows={3} />)
+test.skip('is always accessible via the textbox role', () => {
```
willmcvay, line 23: Yeah, if I get some time, will work out if that mock is still required. _I think_ not but will validate.

src/components/textarea/isCSSFieldSizingSupported.tsx
```diff
@@ -0,0 +1 @@
+export default () => CSS.supports('field-sizing', 'content')
```
willmcvay, line 1: Would this be nice as a generic "cssSupported" helper? Just a question, no strong feelings!

src/helpers/mergeRefs.ts
willmcvay: It makes sense to live here sure.

willmcvay (changes requested):
So on the implementation, looks good and have tested - no issues. Have just requested changes for a couple of small things;

1. Can you add a changelog entry please - just a brief "what changed" summary.
2. An observation on file name casing - I noticed you used camelCase for your hooks - no issue per-se, it's just we've previously used kebab-case throughout. I don't mind which but I would prefer consistency so if you are proposing a change to the convention, can you add to the wiki please?
3. Similar to above, we have added the .test suffix to spec files - this is purely for visual grepping in the editor and for ease of identifying the source and test files when both are open - just a minor point of dev-ex.

Other than that LGTM 👍🏻

src/components/textarea/__tests__/textarea.test.tsx
kurtdoherty: I've carved out time this sprint to take a look at this too. A quick check suggests they're no longer required, but I need to take a closer look as my gut says that's only part of the problem.

src/components/textarea/isCSSFieldSizingSupported.tsx
kurtdoherty: Personally, I don't think it's worth it. I'm generally in favour of "using the Web" and not abstracting away standard web APIs unless absolutely necessary. In this case, I don't think we'd gain enough to warrant our own abstraction.

kurtdoherty:
**todo:** 

- [ ] Add changelog entry describing TextArea changes
- [ ] Update file names to use kebab case
- [ ] Add `.test` suffix to test files

src/storybook/changelog.mdx
kurtdoherty, line 1: **question:** @willmcvay is this all that needs to happen? I've updated the package manifest version to `5.0.0-beta.14` too; not sure if that is meant to be done manually, or if that's handled by a CI job...

willmcvay: The release pipeline bumps the version and raises a PR - it's just the changelog, cheers!

src/components/textarea/isCSSFieldSizingSupported.tsx
willmcvay: Like I say, just a thought :-)

kurtdoherty:
@willmcvay I've actioned the renaming changes. Opted to keep the kebab casing of file names. Personally, I don't love it, but  it feels like something that should be changed in one hit via a script or similar, not incrementally.

willmcvay (approved):

willmcvay:
@kurtdoherty Thanks, have approved. You comfortable doing a release?

kurtdoherty:
> @kurtdoherty Thanks, have approved. You comfortable doing a release?

@willmcvay I was planning to merge then hand over to Design for their final review before doing a release (in case additional changes are identified). Is that okay or do we need to do a release in order for Design to be able to review?

willmcvay:
> > @kurtdoherty Thanks, have approved. You comfortable doing a release?
> 
> @willmcvay I was planning to merge then hand over to Design for their final review before doing a release (in case additional changes are identified). Is that okay or do we need to do a release in order for Design to be able to review?

@kurtdoherty yes, sorry, that's right!

Issue #163, opened 2024-09-13
v5 new components: Menu (basic) and MenuButton


adrian-reapit:
**Abstract**

To support the AppBar improvements an initial version of the Menu component should be made which can be used in a MenuButton and extended later with all other Menu features. 

Menu should provide the basic design with optional title and list of items

![Image](https://github.com/user-attachments/assets/a948bc7b-91b4-43d4-a46a-0d8a99f7a493)

MenuButton toggles the view of Menu below. 

- Menu alignment can be set to left, middle or right side of the button. 
- The Y offset of the Menu can be changed
- The chevron can flip to match the menu visibility
- Button props such as intent and icon can be used in MenuButton 

![Image](https://github.com/user-attachments/assets/3b7f6925-4a7a-4dbb-b8b9-0391730907b8)

**Specification**

- Is new component or existing? New
- Is the component design signed off by product and design?  Menu is signed off. MenuButton is signed off as part of the Navigation changes but not as a standalone component
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-8537&node-type=FRAME&t=vLPA7faa1KEJRIii-0)
- [Link to AppBar design in Figma](https://github.com/orgs/reapit/projects/16/views/1?pane=issue&itemId=78443969) 

 
**Developer Checklist**

- [x] Styles alignment between Design System and Elements
- [x] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [x] Align with accessibility standards / spec as per above
- [x] If relevant, break down component into Styles Only and React component structures
- [x] Ensure all variants of components are documented as appropriate
- [x] Ensure unit test coverage is adequate for component
- [x] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

kurtdoherty:
**suggestion:** The newly baselined [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) could really help simplify the implementation of a Menu component. It may be a touch early to rely on though, it's pretty fresh (it achieved baseline in April 2024)...

[...]

PR #165, opened 2024-09-19
feat: #163 new menu and menu-button component

nurm717123:
### Context
this pr is for initiation for us on how to make a Reapit element's in using best practice convention, there is a feature that has not been in figma which (hasBorder and intent in menu`MenuButton`) which I think need a discussion whether it will usable or not first.
also I have tried to use a [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) but react and typescript didn't support it yet (at least in current version installed). any review or comments for this pr are really appreciated 🔍 👍.

# Pull request checklist

- Initialize `Menu` and `MenuButton` component

2024-10-0310-22-22-ezgif com-optimize


kurtdoherty:
**question:** for my info, why did we move away from `@testing-library/jest-dom` for some of the tests?

**suggestion:** I think it would be good to discuss some code design principles as a team before we merge this work in. For example, `MenuButton` is overloaded with responsibility at the moment, it's interface has many different, unrelated concerns and it does not match what an engineer will see in Figma (i.e the props a designer uses are not the props an engineer will use).

nurm717123:
Hi @kurtdoherty , thanks for the review, I think I can answer some of your questions

> **question:** For my info, why did we move away from `@testing-library/jest-dom` for some of the tests?

actually, I just avoid the unit test like `toBeInTheDocument .etc` or something which depends on `@testing-library/jest-dom` because I didn't see any usage like that in this codebase

> **question**: I don't understand the requirement for adjusting the offset of the popover via a prop? Where is that coming from in the designs?

actually, the design doesn't have it, it have a default offset but, we think that it also better to allow a user to adjust the offset via props.

then about the anchoring logic, I thought because this component only will used in either top navbar or bottom navbar so the position will be controlled individually by the button which triggering the menu to show. I think it need more discussion for the usage.

and finally, for focus management, thanks for noticing that 👍, I can make an update for that.

adrian-reapit:
The menu popup would sit a little bit lower when used in the navigation than when used as a standalone component. The MenuButton is not officially specified so i have pointed out the features to the design team and we will discuss it properly.

kurtdoherty:
> The menu popup would sit a little bit lower when used in the navigation than when used as a standalone component.

Ah, I hadn't noticed that in the designs, thanks for pointing it out 🙏 I'd be asking Design what their thinking behind that difference is. Is it a top nav vs bottom nav difference, or is it related to the anchor position (top of anchor vs bottom of anchor), or something else entirely.

> The MenuButton is not officially specified so i have pointed out the features to the design team and we will discuss it properly.

Thanks 👍  For clarity on my perspective, I don't think a MenuButton like what's presented here should be in Elements, so I think Design are well-justified in providing no official "spec" for it. Menu's can be attached to many visually-different elements, which means there's a fair bit of risk that the interface of the MenuButton presented here will "balloon" over time (which circles me back to my original comment above about being overloaded with responsibility).

kurtdoherty:
@nurm717123 do we need this PR anymore? It's okay if you do, but if not, let's close it out and point future readers towards the replacement 👍

nurm717123:
closing this pr as it have been continued by another pr #163, there will be more pr to be made please refer to https://github.com/reapit/elements/issues/163#issuecomment-2428129267

Issue #139, opened 2024-07-30
v5 component refactor: ButtonGroup


HollyJoyPhillips:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing
- Is the component being re-named? No - 'Button group' is a subset of 'Buttons' 
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-7408&t=8GcgX59FmafMRAda-4)
- [Link to Elements Audit in Figma](https://www.figma.com/design/J6EneNGfpk2GcUuymiOKW4/Elements-Gaps-%2B-Discrepancies?node-id=180-2462&t=Y1ennycgce64B1Qq-4)
- [Link to Elements ](https://elements.reapit.cloud/?path=/story/button--group)
- [Additional behavioural specification]()
- [Accessibility requirements](https://www.w3.org/WAI/ARIA/apg/patterns/button/) May only relate to 'buttons' rather than groups

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

ksolanki7:
@HollyJoyPhillips I was following the wiki to merge and deploy changes to beta for testing and I have done that and now the ticket is available for testing over beta environment [here](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-buttongroup--docs).
What are the next steps or process to get QA/product to review this ticket?

HollyJoyPhillips:
Hi @ksolanki7, that's great thank you. I'll review first, then I will handover to the designers to review.

[...]

Issue #157, opened 2024-08-21
Avoid capturing RTL internals in DOM snapshots


kurtdoherty:
### Context

Currently, many of the component unit tests in Elements take a snapshot testing approach; more specifically, they take a "DOM snapshot" which capture what we expect to see in the DOM when the component is used.

For example, the Avatar component implements snapshots in the following manner:

```javascript
it('should match a snapshot', () => {
  const wrapper = render(<Avatar>Some Content</Avatar>)
  expect(wrapper).toMatchSnapshot()
})
```

Since `wrapper` is the full result object from React Testing Library's (RTL) `render` function, we are capturing its API instead of just the DOM content:

```
exports[`Avatar component should match a snapshot 1`] = `
{
  "asFragment": [Function],
  "baseElement": <body>
    <div>
      <mock-styled.div
        classname=""
        role="img"
        title="A profile image"
      >
        Some Content
      </mock-styled.div>
    </div>
  </body>,
  "container": <div>
    <mock-styled.div
      classname=""
      role="img"
      title="A profile image"
    >
      Some Content
    </mock-styled.div>
  </div>,
  "debug": [Function],
  "findAllByAltText": [Function],
  "findAllByDisplayValue": [Function],
  "findAllByLabelText": [Function],
  "findAllByPlaceholderText": [Function],
  "findAllByRole": [Function],
  "findAllByTestId": [Function],
  "findAllByText": [Function],
  "findAllByTitle": [Function],
  "findByAltText": [Function],
  "findByDisplayValue": [Function],
  "findByLabelText": [Function],
  "findByPlaceholderText": [Function],
  "findByRole": [Function],
  "findByTestId": [Function],
  "findByText": [Function],
  "findByTitle": [Function],
  "getAllByAltText": [Function],
  "getAllByDisplayValue": [Function],
  "getAllByLabelText": [Function],
  "getAllByPlaceholderText": [Function],
  "getAllByRole": [Function],
  "getAllByTestId": [Function],
  "getAllByText": [Function],
  "getAllByTitle": [Function],
  "getByAltText": [Function],
  "getByDisplayValue": [Function],
  "getByLabelText": [Function],
  "getByPlaceholderText": [Function],
  "getByRole": [Function],
  "getByTestId": [Function],
  "getByText": [Function],
  "getByTitle": [Function],
  "queryAllByAltText": [Function],
  "queryAllByDisplayValue": [Function],
  "queryAllByLabelText": [Function],
[...]

PR #169, opened 2024-10-03
refactor: #139 v5 ButtonGroup component

ksolanki7:
- Isolated the ButtonGroup from being a subset of a button component to establish it as a new component
- Designed to accommodate a maximum of five button components as its children.
- Added validation for
      - Appearance of more than 1 primary or icon-only button
      - Icon-only button to appear rightmost
      - Primary button to appear rightmost
      - Primary button cannot be present in the presence of icon-only button
      - All buttons in buttongroup to have same buttonsize and valid variant


src/components/button-group/button-group.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { ButtonVariant, ButtonGroupProps } from './types'
+import { ElButtonGroup } from './styles'
+import { Button } from '../button/button'
+
+// Max number of button a user can have under Button Group
+const maxNumberOfButtons: number = 5
+
+export const ButtonGroup: FC<ButtonGroupProps> = ({ children, ...rest }) => {
+  // Get first 5 buttons to display
+  const childrenArray = Children.toArray(children).slice(0, maxNumberOfButtons)
```
kurtdoherty, line 11: **note:** @ksolanki7 Firstly, I appreciate how much care you've gone to here to codify all the design guidelines for button groups and can see you've gone to a lot of effort to do so. Please keep that in mind when reading the issue/s I describe below and what my suggestion means.

**issue (blocking):** The approach you've used here where the parent (the button group) is "inspecting" and directly manipulating its `children` is very fragile and generally leads to poor outcomes.

Typically (like 99.99% of the time), we want parent components like ButtonGroup to depend on an abstract type (e.g. `ReactNode`) instead of a concrete type (e.g. `Button`). This is because depending on a concrete type constraints the possible outcomes _a lot_.

Here's some examples of what we _can't_ do with ButtonGroup that almost any FE engineer would expect to be able to do:

```jsx
// Design may want to experiment with a special kind of button
// that doesn't yet exist in the Design System and is therefore
// not supported by Elements' Button component.
<ButtonGroup>
  <SpecialButtonForMyProduct />
  <Button ... />
</ButtonGroup>
```

```jsx
// Some products may have pricing tiers (or other properties)
// that impact what features are available to particular users.
// As such, we may want to wrap a particular Button in a component
// that can determine whether its child should be rendered or not.
<ButtonGroup>
  <IsFeatureAvailable requirements={{ pricingTier: 'premium' }}>
    <Button ... />
  </IsFeatureAvailable>
  <Button ... />
</ButtonGroup>
```

There are many other examples, but hopefully you can see what I'm getting at. Ultimately, as I talked about in our kick-off, many of the design guidelines for the ButtonGroup are what I would class as "authoring" concerns. They are things that the _designer_ must primarily care about when they are authoring designs for a particular product, they are not things we should try to codify.

**suggestion:** I'd suggest we greatly simplify `ButtonGroup` so that it just accepts `children` and renders them inside an appropriately styled `<div>` element (horizontal flex layout with a gap of 8px via the `--spacing-2` design token).

kurtdoherty (commented):
**praise:** @ksolanki7 congratulations on your first PR in Elements! 👏 

**note:** I have not reviewed everything in this PR yet. I think the approach you've taken for ButtonGroup needs resolving before review the rest of the work you've done.

**note:** I use https://conventionalcomments.org/ when providing feedback on work. It should hopefully be pretty self-explanatory, but the linked website provides some good context on why I use them, plus definitions for what some of the "labels" mean 👍 

**aside:** I think we may have finished our kick-off for this work on different pages, because the approach you've taken for ButtonGroup is the opposite of what I thought we'd agreed on when we'd talked 😅

Next time, I think it would be good for one of us to summarise our understanding in writing (via the GitHub issue perhaps?) so that we can confirm we're on the same page before you invest so much time writing code.

src/components/button-group/button-group.tsx
ksolanki7: @kurtdoherty 
**praise:** Thank you so much for your detailed feedback. I appreciate it. I agree, I have over engineered the solution to its minute details and off tracked from the actual understanding of the component requirement.

**note:** I have updated my PR, as discussed in our initial kick-off. Please review and provide feedback.

src/components/button-group/__tests__/button-group.test.tsx
```diff
@@ -0,0 +1,53 @@
+import { render } from '@testing-library/react'
+import { ButtonGroup } from '../button-group'
+import { Button } from '@/components/button/button'
+
```
kurtdoherty, line 4: **todo:** Snapshotting the result of the `render` function means we capture React Testing Library's API. This problem is described in #157. Can you please update these tests to use the `asFragment` helper returned by `render`?

kurtdoherty, line 4: **issue (non-blocking):** When writing test cases for a particular unit (the subject under test), it's helpful to apply the heuristic of _necessary_ and _sufficient_ to help determine if we're specifying the subject's behaviour well.

1. _necessary_ = tests should only test what they need to in order to fully exercise the subject;
2. _sufficient_ = tests should assert that the subject does everything a new implementation would also need to do.

With this heuristic in mind, we can assess the tests below and decide whether they meet the criteria of necessary and sufficient.

1. _Are they necessary?_ The first test is, but the latter two aren't, because they aren't specifying behaviour of the `ButtonGroup`. This is because the `ButtonGroup` does not care _what_ it renders, only that it renders whatever _children_ it has been given in a div _with the correct styles_.
2. _Are they sufficient?_ Kind of. Since we're only capturing a "DOM snapshot", not a visual snapshot, it's hard for us to sufficiently specify the gap between the `ButtonGroup`'s children is `--spacing-2` via an assertion. The best we can do is document that behaviour in the test description and otherwise explain why its not asserted by the test 🤷 

**suggestion:** I'd reduce this to a single test...
```javascript
test('renders its children horizontally with a gap of --spacing-2', () => {
  const { asFragment } = render(
    <ButtonGroup>
      <Button>Button 1</Button>
      <Button>Button 2</Button>
      <Button>Button 3</Button>
    </ButtonGroup>
  )
  // NOTE: We do not currently assert the children are laid out horizontally
  // with the correct gap because that behaviour is defined within CSS.
  expect(asFragment()).toMatchInlineSnapshot()
})
```

src/components/button-group/button-group.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+  },
+}
+
+export const WithPrimaryButton = {
+  render: () => (
+    <ButtonGroup>
+      <Button>Button 1</Button>
+      <Button>Button 2</Button>
+      <Button>Button 3</Button>
+      <Button>Button 4</Button>
```
kurtdoherty, line 38: **quibble:** I'm don't think we need the comment below to be here... until we tackle the Button work, there's no guarantee we'll change `intent` to `variant`.

src/components/button-group/button-group.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC } from 'react'
+import { ButtonGroupProps } from './types'
+import { ElButtonGroup } from './styles'
+
```
kurtdoherty, line 4: **question:** Is there a reason you opted to put `ButtonGroupProps` in `types.ts` instead of adjacent to the component here?

**suggestion:** Unless the type is shared with other units of code in the `/button-group` folder, it's preferable to co-locate it with the component's implementation. Helps increase cohesion in the code.

src/components/button-group/styles.ts
```diff
@@ -0,0 +1,19 @@
+// import { css } from '@linaria/core'
+import { isMobile } from '@/styles/media'
+import { styled } from '@linaria/react'
+
+export const ElButtonGroup = styled.div`
+  display: flex;
```
kurtdoherty, line 6: **question:** Where does the `align-items` CSS come from? Figma?

src/components/button-group/styles.ts
```diff
@@ -0,0 +1,19 @@
...
+  align-items: center;
+  gap: var(--spacing-2);
+  flex-wrap: wrap;
+
```
kurtdoherty, line 10: **question:** Where do these mobile-specific CSS properties come from?

src/components/button/button.stories.tsx
```diff
@@ -42,6 +42,13 @@ export const StandardAttributes = {
   },
 }
 
+/**
+ *
+ * **This option will be removed from `Button` in a future version.**
+ * And it will sever as a new component `ButtonGroup`.
```
kurtdoherty, line 48: **quibble:** I find the first sentence here a little confusing. Doesn't seem clear that we're talking about the `ButtonGroup` component exported from the `/button` folder vs the `ButtonGroup` component exported from `/button-group` 🤔 

**suggestion:** Maybe we can simply remove this story from Button's docs altogether?

src/components/button/button.tsx
```diff
@@ -16,6 +15,7 @@ import {
   elButtonSizeMedium,
   elButtonSizeSmall,
 } from './__styles__'
```
kurtdoherty, line 17: **todo:** I think we need to decide what the migration journey for consumers of the existing `ButtonGroup` will look like. What we want that journey to look like will influence whether we want to be exporting the new ButtonGroup from Button (like you are here) or if we want to continue exporting the now-deprecated ButtonGroup as-is, or if we should just delete the existing ButtonGroup entirely.

kurtdoherty (commented):
Thanks @ksolanki7. I've finished a full, second review 👍 Looking good 👏

src/components/button-group/__tests__/button-group.test.tsx
ksolanki7: I agree with this and will update it.

ksolanki7: Understood, I will update this.

src/components/button-group/button-group.stories.tsx
ksolanki7: Thank you for the clarification. I was under impression that the Button component has been finalized, and that we would be transitioning from 'intent' to 'variant.' I agree that this remains uncertain until we proceed with the Button work.

src/components/button-group/button-group.tsx
ksolanki7: I was following the documentation's guidelines by separating interfaces and types into `types.ts` for consistency across the codebase.

Yes, it is not a shared type, and I agree to move `ButtonGroupProps` to the component to increase cohesion.

src/components/button-group/styles.ts
ksolanki7: I added that to support mobile-first CSS as there wasn't any in figma.

src/components/button-group/styles.ts
ksolanki7: I added that to support mobile-first CSS, as it wasn't provided in the Figma design. My aim was to implement a 100% width button approach.

src/components/button/button.stories.tsx
ksolanki7: I agree, I will update the PR by removing `ButtonGroup` altogether from `/button`

src/components/button/button.tsx
ksolanki7: Yes, I believe this will require a discussion with the team, or at least with you, to reach an agreement on the best approach.

src/components/button-group/__tests__/button-group.test.tsx
ksolanki7: @kurtdoherty Thank you for your feedback.
**issue:** As elements prettier version is `3.2.5`, jest is throwing an error stating `Jest: Inline Snapshots are not supported when using Prettier 3.0.0 or above.`. 

**thought:** Though there are work around like version rollback to < `3.0.0`, I am thinking of using `toMatchSnapshot()` until prettier library is updated with a fix.

**question:** Do you have any other recommendation for this?

kurtdoherty: Ah, that's a pain. All good, `toMatchSnapshot` is fine, it can just be nice to see them inline when they're nice and small.

src/components/button-group/button-group.tsx
kurtdoherty: TIL! Sorry, I had missed that recommendation in the guidelines. Might be something worth raising in the Elements dev chat on Teams; as per my feedback here, I don't think it makes sense to _always_ separate them.

src/components/button-group/styles.ts
kurtdoherty: Okay, I don't think this makes sense to centre the buttons in the button group. That would be something the design spec would be clear about if it were required.

Same deal with the `flex-wrap` property. Nothing in the design spec suggests that buttons should wrap within the button group.

src/components/button-group/styles.ts
kurtdoherty: **issue:** similar to my comments above, I don't think we should be adding styles that aren't documented in Figma.

**suggestion:** Again, if you think this kind of treatment on mobile devices is important, ask Design via Figma.

src/components/button/button.tsx
kurtdoherty: If it were up to me, I'd nuke the old one completely, but it's worth checking with Adrian (perhaps via the Elements dev chat on Teams) to see how he'd prefer the migration journey to look.

src/components/button/button.tsx
```diff
@@ -118,16 +126,16 @@ export const Button: FC<ButtonProps> = ({
   )
 }
 
-export const ButtonGroup: FC<ButtonGroupProps> = ({ children, alignment, ...rest }) => {
+export const DeprecatedButtonGroup: FC<ButtonGroupProps> = ({ children, alignment, ...rest }) => {
```
kurtdoherty, line 129: **todo:** can you please add `/** @deprecated */` JS Doc to this component so that we get strike-through visual behaviour in our IDEs 🙏 

You may also need to add the same JS Doc to the `button/index.ts` file...not 100% sure though.

src/components/button/__styles__/index.ts
```diff
@@ -208,7 +208,7 @@ export const ElButton = styled.button`
   }
 `
 
-export const ElButtonGroup = styled.div`
+export const DeprecatedElButtonGroup = styled.div`
```
kurtdoherty, line 211: **todo:** can you please add `/** @deprecated */` JS Doc to this component so that we get strike-through visual behaviour in our IDEs 🙏

src/components/steps/steps.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import { useState } from 'react'
 import { Steps, StepsVertical } from './index'
-import { Button, ButtonGroup } from '../button'
+import { Button, DeprecatedButtonGroup } from '../button'
 // import { BodyText, Subtitle } from '../typography'
-// import { Button, ButtonGroup } from '../button'
+// import { Button, DeprecatedButtonGroup } from '../button'
```
kurtdoherty, line 5: **question:** can these commented out import statements just be removed?

src/storybook/changelog.mdx
```diff
@@ -16,6 +16,11 @@ We will publish release version history and changes here. Where possible, we wil
 
 Beta versions should be relatively stable but subject to occssional breaking changes.
 
+### **5.0.0-beta.15 - 03/10/24**
+
+- New component
+- `ButtonGroup` - Isolated the ButtonGroup from the button component to establish it as a new component, designed to accommodate a maximum of five button components as its children.
```
kurtdoherty, line 22: **suggestion:** Probably worth mentioning that the old button group component has been deprecated and renamed to `DeprecatedButtonGroup`. Both of these changes constitute a breaking change too.

kurtdoherty (approved):
Nice one @ksolanki7 👏 Looks good. A few more minor comments, but none of them are blocking.

src/storybook/changelog.mdx
ksolanki7: @kurtdoherty Thank you for nitpick, have updated it now.

PR #171, opened 2024-10-11
feat: #163 new menu component

nurm717123:
### Context
this pr is smaller version of #163 , but it only focusing on Menu component

# Pull request checklist

includes: 
- default menu variant
- allow mixed menu group
- reusable popover wrapper

further improvement like:
- Radio menu or other variants
- auto anchoring (currently just top and bottom)

will be separated in different pr

note: other feature like the autofocus still waiting for discussion especially to cover styling (reference: https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj?node-id=2038-7225#982402216)


nurm717123:
https://github.com/user-attachments/assets/0426248e-ea50-410b-82e0-f4ec683d5ee6

src/components/menu/menu.tsx
kurtdoherty, line 1: **issue:** I provided similar feedback for the dialog work (see my [comments](https://github.com/reapit/elements/pull/170/files/c8cd54c48d7f98facf062f800e9398d5c961881d#r1796455334) at on #170).

Specifically, I don't think Elements should have any kind of `MenuComposed` React component. `MenuGroup` and `MenuRadioGroup` also take the same approach as `MenuComposed`, so this feedback applies to them in their current form as well.

**Why do I think this?**

The Menu component specified in Figma is a molecule, but that doesn't mean we need a single React component in Elements to match that molecule, it just means we need the same atoms.
The `MenuComposed` component isn't simplifying a really complex interface, it's just hiding the atoms behind some props. So we gain significantly less flexibility long term as the Design System evolves for the _perception_ of a simpler/more convenient interface.

Unlike the situation with the `DialogComposed` component that I provided feedback for, the `MenuComposed` component does simplify the amount of boilerplate involved, but I think that's only because it's not handling any sort of tigger, which means `MenuComposed` is actually `MenuListComposed`. 

Compare:

```javascript
<Menu>
  <Menu.Trigger>{({ getTriggerProps }) => ...</Menu.Trigger>
  <Menu.Popover>
    <Menu.List>
      <Menu.Item>Item 1</Menu.Item>
      <Menu.Item>Item 2</Menu.Item>
    </Menu.List>
  </Menu.Popover>
</Menu>
```

with:

```javascript
<Menu
  groups={[
    { children: "Item 1" },
    { children: "Item 2" },
  ]}
/>
```

This looks simpler, but that's coming at the cost of some pretty significant trade-offs:
1. The single `groups` prop implies all items must be grouped.
2. The single `groups` prop prevents feature-level code from encapsulating menu groups as React components themselves (the best we can do is share around an object that satisfies the `MenuItemGroupProps` type, which is not the "React way").
3. JSX is typically more readable than arrays of objects. Some of the values passed to the `groups` prop look quite complex compared to how it would look as simple JSX.
4. There's no handling of the trigger; this will require another prop of some kind to be added.
5. There's no handling of the popover; this will require another prop as well.

In contrast, while the first approach (currently) looks like more keystrokes, its simpler to my eye, plus feels easier to read (because it looks more like plain HTML) and is way more flexible for us long term. Again, all of this feedback largely applies to the `MenuRadioGroup` and `MenuGroup` components (they're baking too much into themselves).

**suggestion:** I think all we need here are the styled components, some atoms that connect those styled components with the Menu's context, and the main `Menu` component.

src/components/menu/menu.atoms.tsx
kurtdoherty, line 1: **question:** what value do you see these components providing consumers beyond what the styled components defined in `styles.ts` already provide?

kurtdoherty (commented):
**praise:** Thanks for the work you've done (again) for the Menu component 👏 

**issue:** This PR is way too big for me to review. It's 1200 lines of code. We need to break work into smaller, incremental, more focused pieces of work. As it stands, I don't have time to review all of this in one sitting. There's hours of review work here, from overall design (what I've focused on for `menu.tsx`) to implementation detail, which is compounded by the fact there's lots of different concerns being implemented (Popover vs click away listeners vs radio-style menu item groups vs the basic menu components).

I would like to see this broken up further before I review any more of it. Our PRs need to get under 250 lines of code for our reviews to be effective. If you need help identify ways to break this work up, let me know and I can provide some suggestions 👍

src/components/menu/menu.atoms.tsx
nurm717123: at that time I thought it was the current convention (as this pattern also exist in accordion.atoms), which I think it can simplify React consumer to just use `<Component />` instead of `<ElComponent />`, which I think its little bit unnecessary at that time, I did notice your comment about this in #170 pr but I'm didn't have time to address it, sorry for re-confirming that

src/components/menu/menu.tsx
nurm717123: yeah, I agree I think it more like a `MenuListComposed` or `MenuGroupComposed` because each menu could have multiple group, my aim at that time is because `MenuRadioGroup` and `MenuCheckboxGroup` (this one is excluded in this pr because I'm not sure if its currently needed or not but it might possible in the future) will have the same props, so probably its good to encapsulate them. but I am also agree to let the React consumer to do it their way as its also easy and give more flexibility to them 👍

nurm717123:
thanks Kurt, I will broken this pr further 👍

nurm717123:
Hi @kurtdoherty, I think its the minimum for this pr , the snapshots are quite of code additions, or is there any more than I can cut ? 🤔

src/components/menu/styles.ts
```diff
@@ -0,0 +1,68 @@
+import { styled } from '@linaria/react'
+import { Text2XS } from '../typography'
```
kurtdoherty, line 2: **question:** what's your reasoning behind using this pre-v5 component instead of just plain CSS?

src/components/menu/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const ElMenuPopover = styled.div`
+  position: absolute;
+  white-space: nowrap;
+  z-index: 98;
+`
+
+export const ElMenu = styled.div`
```
kurtdoherty, line 10: **question:** Shouldn't this be `ElMenuList` instead of `ElMenu`?

src/components/menu/styles.ts
```diff
@@ -0,0 +1,68 @@
+import { styled } from '@linaria/react'
+import { Text2XS } from '../typography'
+
+export const ElMenuPopover = styled.div`
+  position: absolute;
+  white-space: nowrap;
+  z-index: 98;
+`
```
kurtdoherty, line 8: **question:** Can you help me understand why these styles are necessary on the popover?

src/components/menu/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+export const ElMenu = styled.div`
+  overflow: hidden;
+  display: flex;
+  min-width: 200px;
+  width: fit-content;
+  padding: var(--spacing-2) var(--spacing-none);
+  border-radius: var(--corner-default);
+  background-color: var(--fill-white);
+  box-shadow: 0px 4px 16px 0px #222b3329;
+`
```
kurtdoherty, line 19: **question:** Does the Design System place the `box-shadow` on the MenuList component? It's not clear to me if this should belong to the Popover or the MenuList 🤔

src/components/menu/menu.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const Trigger = ({ children }) => {
+  const { triggerProps, isOpen } = useMenu()
+
+  return children({ triggerProps, isOpen })
+}
+
+const Popover: FC<PropsWithChildren<{ containerRef: MutableRefObject<HTMLDivElement | null> }>> = ({
+  children,
+  containerRef,
```
kurtdoherty, line 15: **question:** why is this `containerRef` supplied by the consumer? I would have assumed the consumer doesn't need to concern itself with this and this component would simply set it up internally and attach it to the `ElMenuPopover` rendered here 🤔

src/components/menu/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  /* customize html default button */
+  button {
+    background-color: transparent;
+    cursor: pointer;
+    border: none;
+    border-radius: 0;
+  }
+`
+
```
kurtdoherty, line 52: **suggestion:** If `ElMenuItemGroup` and `ElMenuItemGroupTitle` could get moved to a separate piece of work with the group-related atoms, that would be great 🙏

src/hooks/use-click-outside/index.ts
kurtdoherty, line 1: **note:** FYI the newish AbortController API provides a really nice way to handle `addEventListener` logic. Not a big deal here given we're only dealing with a single event listener, but it is quite nice. [Don't sleep on AbortController](https://kettanaito.com/blog/dont-sleep-on-abort-controller) is a really good article that talks through it.

src/components/menu/menu.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { MenuItemGroup } from './menu.atoms'
+import { MenuProvider, useMenu } from './provider'
+import { ElMenu, ElMenuItemContainer, ElMenuPopover } from './styles'
+
+const Trigger = ({ children }) => {
+  const { triggerProps, isOpen } = useMenu()
+
+  return children({ triggerProps, isOpen })
+}
+
```
kurtdoherty, line 12: **suggestion:** I think this `Popover` is doing enough to warrant it's own file (much like we have a separate file for the context/provider.

src/components/menu/provider.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { createContext, Dispatch, FC, HTMLAttributes, SetStateAction, useState } from 'react'
+
+interface MenuContextValue {
+  triggerProps: HTMLAttributes<HTMLButtonElement>
+  popoverProps: HTMLAttributes<HTMLDivElement>
+  isOpen: boolean
+  setIsOpen: Dispatch<SetStateAction<boolean>>
```
kurtdoherty, line 7: **suggestion:** Personally, I think functions like `open` and `close` are preferable to `setIsOpen` because there's a higher cognitive load for the reader when they see `setIsOpen(true)` or `setIsOpen(false)` as compared to reading `open` or `close`

src/components/menu/provider.tsx
```diff
@@ -0,0 +1,41 @@
...
+const MenuProvider: FC = ({ children }) => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  const triggerProps = {
+    'aria-haspopup': true,
+    'aria-expanded': isOpen,
+    onClick: () => {
+      setIsOpen((prev) => !prev)
+    },
+  }
```
kurtdoherty, line 21: **question:** Have you referred to https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/ and https://www.w3.org/WAI/ARIA/apg/patterns/menubar/ for guidance on what keyboard interactions we need to handle? If so, is that work currently in- or out-of-scope for this series of PRs?

src/components/menu/menu.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+          </ElMenuItemContainer>
+          <ElMenuItemContainer>
+            <Button type="button">This is a Button</Button>
+          </ElMenuItemContainer>
+        </ElMenuItemGroup>
+      </ElMenu>
+    )
+  },
+}
+
```
kurtdoherty, line 31: **suggestion:** I think is fine to show just the Menu.List usage in a story, but I think we should also have a complete story with a menu button that opens the menu when clicked etc.

src/components/menu/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+  display: flex;
+  min-width: 200px;
+  width: fit-content;
+  padding: var(--spacing-2) var(--spacing-none);
+  border-radius: var(--corner-default);
+  background-color: var(--fill-white);
+  box-shadow: 0px 4px 16px 0px #222b3329;
+`
+
+export const ElMenuItemContainer = styled.li`
```
kurtdoherty, line 21: **note:** One thing that isn't clear to me from the Design spec is whether menu items that do something asynchronously when clicked need to have a busy state, or if those kinds of async actions from menu items will be communicated in another manner.

kurtdoherty (commented):
**praise:** Thanks for the work here @nurm717123. It's looking pretty solid 👏 Most of my feedback so far has been related to which units own what responsibilities. Once we get that sorted, I'll do a second review that will focus a bit more on the testing side of things 👍 

I think one of the key questions that is unclear for me is how we should handle menu items being `<button>` or `<a>` elements and what the consequences are of _trying_ to support the use of Elements `Button` component or `Link` component?

Beyond that, I'm keen to understand your thoughts around keyboard interaction and animating the opening/closing of the menu.

> Hi @kurtdoherty, I think its the minimum for this pr , the snapshots are quite of code additions, or is there any more than I can cut ? 🤔

That's fine. Snapshots definitely contribute to the line count! Some things that can help in future:
- You can leave the snapshots uncommitted until later in the review process; that way reviewers can initially focus on the test cases and production code then just check the snapshots before giving their final approval.
- I often try to break PRs up into layers of complexity; e.g. this PR could have focused on just the styled components and showing screenshots of how they render; after that, you could then layer in the React-specific code like the context/provider, useMenu hook and so on.

Sometimes it's just hard to keep PRs nice and small. I'll share some PR guidance that the Console FE team try to keep in mind when raising and reviewing PRs; it'll talk about other things you can do as a submitter that make the review process easier/less taxing for reviewers (even when PRs are bigger than either side would prefer :P)

src/components/menu/styles.ts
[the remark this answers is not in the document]
nurm717123: initially its because its a `li` element, which later the consumer use it to wrap their list item e.g: button, link, or .etc. and if we used without any props it will not display anything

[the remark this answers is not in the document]
nurm717123: but now I think `ElMenuItem` is simpler and still intuitive

src/components/menu/styles.ts
nurm717123: sorry, I thought that `Text2Xs, TextXs` are the base style which different with `SubTitle, Tittle, .etc`

src/components/menu/styles.ts
nurm717123: yes, that is initially because despite this is consumed as `Menu.List` its actually the outer container of the menu card that has styling , but I think its better to match the classname with component name now 👍

src/components/menu/styles.ts
nurm717123: >Does the Design System place the box-shadow on the MenuList component?

yes, it has a box-shadow, but for where I should put it, in my understanding, `MenuList` is a styled container of the `MenuGroup`, and the `Popover` is just container which contain the logic.

question: should we combine them instead ?

src/components/menu/styles.ts
nurm717123: yeah, I didn't see any busy state styling in Design spec either so I assume if the list item clicked and its button, then it will close the menu (this logic handled in `Popover`)

src/components/menu/styles.ts
nurm717123: yes, the context is typically this component will open triggered by a button, and they should have a `relative` parent, that is what the `position: absolute and z-index` for the `white-space` is so that the list item text not collapsed
image

src/components/menu/menu.tsx
nurm717123: its because it used in `useClickOutside` hook, the hook basically does two thing:
1. register `handleClick` event listener to the 'document`
2. `handleClick` then will check if clicked `target` is outside the `containerRef` it will run the passed callback ( in this case is "closing menu")

src/components/menu/provider.tsx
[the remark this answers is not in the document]
nurm717123: > suggestion: Regarding your decisions on which units are responsible for what logic...

agree, I think it would be better to rename it to `useMenuContext` and `menu-context.tsx` 👍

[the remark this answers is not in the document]
nurm717123: > suggestion: For the "prop getter pattern", we would typically provide a function (e.g. getTriggerProps) instead of an object so that we can help consumers "merge" props with the same name, instead of leave this up to them.

yes, initially I use this pattern, but I make it as object because I thought it will help reduce the code changes and we can refactor it to be function again later, but I think its not worth time, thanks for explaining it more.

src/components/menu/provider.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const popoverProps = {
+    role: 'menu',
+    hidden: !isOpen,
```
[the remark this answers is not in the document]
nurm717123: > question: does the use of hidden here make it harder for us to animate the opening/closing of the popover?

no, turns out it doesn't required since we custom the `display` prop via css

>suggestion: If it does, I think a data-open attribute would make sense.

ok, I'll try it

src/components/menu/provider.tsx
nurm717123: > question: Have you referred to https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/ and https://www.w3.org/WAI/ARIA/apg/patterns/menubar/ 

not yet, thanks I will look into that

> To be clear, I'm not suggesting keyboard interaction should be implemented in this PR. Just wondering if its been considered.

I thought it would be good to be included as well, I'll try to split it then

src/components/menu/styles.ts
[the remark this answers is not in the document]
nurm717123: >I think this needs some more thought/analysis to work out if there's a simpler way for us to handle this. For example, some questions that come to mind are:
> Do we even need to support the use of a Button component as a child of a MenuItem?

yes, I'm also not sure about that, so I just add a basic style override for `Button`, but I think not supporting it still ok because user still just switch the component and then done.

> Do we need the ElMenuItem to be an <li> element or can it be an <a> or <button> element with a role="menuitem"?

I think `li` will be more flexible because I think the item could also contain `input` probably in the future, if allowed 🤔

src/components/menu/menu.atoms.tsx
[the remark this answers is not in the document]
nurm717123: ok, I'll try it

src/components/menu/styles.ts
kurtdoherty: > in my understanding, MenuList is a styled container of the MenuGroup, and the Popover is just container which contain the logic.

Yeah, I think that's the way I lean too, so let's leave it like that and see what we learn 👍

src/components/menu/menu.tsx
kurtdoherty: > its because it used in `useClickOutside` hook

Yes, but couldn't this component take care of the ref internally?

```javascript
const Popover: FC<...> = ({ children }) => {
  const { setIsOpen, ... } = useMenu()
  const ref = useRef<HTMLDivElement>(null);

  useClickOutside(ref, ...)

  if (isOpen) {
    // ...
    return (
      <ElMenuPopover ref={ref} ...>
        {children}
      </ElMenuPopover>
    )
  }

  return null
}
```

src/components/menu/menu.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <ElMenuPopover {...popoverProps} onClick={onClick}>
+        {children}
+      </ElMenuPopover>
+    )
+  }
+
+  return null
+}
+
```
kurtdoherty, line 38: **note:** I forgot to mention here... if we were to split up the MenuProvider into a simple context provider and a `useMenu` hook, our main `Menu` component here becomes the place where we tie the two together:

```javascript
function Menu({ children }) {
  const menu = useMenu()
  return (
    <Menu.Context.Provider value={menu}>
      {children}
    </Menu.Context.Provider>
  )
}
```

nurm717123:
Hi @kurtdoherty thanks for the detailed feedback 👍 

> I think one of the key questions that is unclear for me is how we should handle menu items being <button> or <a> elements and what the consequences are of trying to support the use of Elements Button component or Link component?

In my opinion, I think we should encourage use to just use the plain html button for this, so we don't have to provide a style override for them

> Beyond that, I'm keen to understand your thoughts around keyboard interaction and animating the opening/closing of the menu.

I think for the animation and keyboard, I'm not sure if this in the scope of this PR/Tickets, or is it ok if I add it on later pr ? cc @adrian-reapit 

feedback summary (will be updated accordingly):
- [ ] exclude snapshots
- [ ] exclude the atoms for later pr
- [ ] refactor provider.tsx, rename provider to 'menu-context` and the hook name
- [ ] use `data-open` attr instead of `hidden`
- [ ] un-use existing `Text2XS .etc`
- [ ] move popover to new file
- [ ] rename `setIsOpen` to `open`
- [ ] add a story to show complete component with button triggerer

src/components/menu/menu.tsx
nurm717123: > Yes, but couldn't this component take care of the ref internally instead of relying on a consumer to set it up?

I think no, its because otherwise the callback ("closeMenu") will trigger when user clicking the button triggerer

nurm717123: I think you're correct, the button triggerer what I mean't here is the "Portfolio: All Portfolios" as the gif shown which is what I expecting, but what I can think of how it does is make the popover as an overlay then we don't have to worry about button triggerer to be clicked too, but now the question when menu is open (and overlay is covering evertyhing) it could also hide `:hover` style of other element behind, is that correct or it was totally different logic ?

kurtdoherty: If I'm understanding you correctly, I don't think the popover element should be covering everything in the browser's viewport (this would prevent the popover from being able to accurately reposition itself when there's not enough space above/below the trigger button). Also, I'm not aware of any design requirement for the popover to have a backdrop like our dialogs and drawers do.

Does that answer your question? If not, we can scheduled a call next week to discuss verbally 👍

nurm717123: yes you right, if console team doesn't use overlay than it should be different logic, i will try that 👍

src/components/menu/styles.ts
nurm717123: I think also because the menu typically used for navbar, it less necessary to provide a busy state 🤔

src/hooks/use-click-outside/__tests__/index.test.tsx
```diff
@@ -0,0 +1,65 @@
+import { useRef } from 'react'
+import { useClickOutside } from '..'
+import { fireEvent, render } from '@testing-library/react'
+
+const TestComponent = ({ onClickOutside, isUsingParent = true }) => {
+  const ref = useRef<HTMLDivElement>(null)
+  useClickOutside(ref, onClickOutside)
+
```
kurtdoherty, line 8: **note:** When writing tests, it's generally good practice to avoid our test code having conditionals in it, as it makes it harder for us to know whether our subject-under-test is behaving correctly, or if our conditional test code is getting in the way.

**suggestion:** I think it would be better to have more than one test component instead, or to do away with the `isUsingParent` prop and just have the same DOM content rendered no matter what (see my comments below for why this may be an option)

src/hooks/use-click-outside/__tests__/index.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <TestComponent onClickOutside={onClickOutside} />
+        <div>Outside Element</div>
+      </>,
+    )
+
+  describe('has parent ', () => {
+    it('should call onClickOutside when clicking outside the element', () => {
+      const { getByText } = renderElement()
+
+      fireEvent.mouseDown(getByText('Outside Element'))
```
kurtdoherty, line 38: **question:** Currently there is no element with this text in it, so I'm not sure how this test is passing 🤔 If it is passing, that would suggest we're getting a false positive and we need to adjust the test so that this scenario correctly fails (otherwise our test isn't designed to "fail well" and its value is undermined)

src/hooks/use-click-outside/__tests__/index.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    it('should not call onClickOutside when clicking inside the element', () => {
+      const { getByText } = renderElement()
+
+      fireEvent.mouseDown(getByText('Inside Element'))
+      expect(onClickOutside).not.toHaveBeenCalled()
+    })
+  })
+
+  describe("doesn't have parent", () => {
```
kurtdoherty, line 50: **question:** Is it actually possible for a React component to never have a parent element? I'm not sure it is, so we might be able to get away with not have these tests of the associated behaviour in `useClickOutside` 🤔

src/hooks/use-click-outside/index.ts
```diff
@@ -0,0 +1,20 @@
+import { MutableRefObject, useEffect } from 'react'
+
+export const useClickOutside = (ref: MutableRefObject<HTMLDivElement | null>, onClickOutside: VoidFunction) => {
```
kurtdoherty, line 3: **note:** Not something you need to change in this PR, but I think this hook might be better named `useClickOutsideListener` 🤔 Currently, `useClickOutside` could be taken to mean it allows a consumer to click outside an element or something... Not a big deal though and I think we're better off moving forward and leaving that rename for another time

src/hooks/use-menu/__tests__/index.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+describe('useMenu', () => {
+  it('getTriggerProps should apply aria and class attributes correctly', () => {
+    const { result } = renderHook(() => useMenu())
+    const triggerProps = result.current.getTriggerProps({
+      className: 'custom-class',
+    })
+
+    expect(triggerProps['aria-haspopup']).toBe(true)
+    expect(triggerProps['aria-expanded']).toBe(false)
```
kurtdoherty, line 12: **todo:** the presence of the `className` attribute isn't being asserted here, but the test description suggests we care about it. We either need to add the assertion or update the test description.

src/hooks/use-menu/__tests__/index.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  })
+
+  it('getPopoverProps should apply data-open attribute correctly', () => {
+    const { result } = renderHook(() => useMenu())
+
+    const popoverProps = result.current.getPopoverProps()
+    expect(popoverProps['data-open']).toBe(false)
+
+    result.current.openMenu()
```
kurtdoherty, line 21: **question:** should we exercise `openMenu` in the `getTriggerProps` as well in order to ensure `aria-expanded` becomes true?

kurtdoherty (approved):
Nice work @nurm717123, I think we've made it! 🥳  I've left a few minor comments/suggestions but I don't think I need to review this again (unless you action my suggestions and want me to take a look at specific things).

Great outcome 👏

src/hooks/use-click-outside/__tests__/index.test.tsx
nurm717123: that text actually under `renderElement` fn, yeah I think it could cause false positive, I will update the test to be more clear 👍

src/hooks/use-click-outside/__tests__/index.test.tsx
nurm717123: that's right, I think its not possible 👍

src/hooks/use-menu/__tests__/index.test.tsx
nurm717123: yes sure, `getTriggerProps` already have test for its `onClick` just need to add one assertion.

src/components/menu/styles.ts
nurm717123: > does the design spec show menu item text wrapping like this?

currently no, I think it because the example only for standard sized text, so long text might be rare

kurtdoherty: Okay, could be worth confirming whether we need text wrapping or not. We can do this as part of the next PR if you'd like, so long as we don't forget to remove this `white-space` styling from the popover once we know how items should/should not handle long label text.

nurm717123: yeah agree, I I'll remove the `white-space` in this pr

nurm717123:
> Nice work @nurm717123, I think we've made it! 🥳 I've left a few minor comments/suggestions but I don't think I need to review this again (unless you action my suggestions and want me to take a look at specific things).
> 
> Great outcome 👏

thanks Kurt for the detailed feedbacks, the resource that you attach really help me 👍 🥳

PR #178, opened 2024-10-23
feat: #163 new menu(item, group, group-title) component 

nurm717123:
### Context
this pr is smaller version of #163 , but it only focusing on Menu sub component which is menu item, menu group and menu group title

# Pull request checklist
- [x] menu item allows (button, anchor, div)
- [x] menu group with optional title

further improvement like:
- Radio menu or other variants
- a11y keyboard interactions
- auto anchoring (currently just top and bottom)

menu item 2


src/components/menu/menu-popover.tsx
```diff
@@ -11,8 +11,8 @@ export const MenuPopover: FC = ({ children }) => {
 
   if (isOpen) {
     const onClick = (e) => {
-      // close menu if button or any other element inside e.g icon is clicked
-      const button = e.target.closest('button')
+      // close menu if button or any other element inside e.g icon button is clicked
+      const button = e.target.closest('button, [role="button"]')
```
nurm717123, line 15: to solve possible issue when consumer use `<div role="button">" for list item

src/components/menu/styles.ts
```diff
@@ -7,7 +7,7 @@ export const ElMenu = styled.div`
 
 export const ElMenuPopover = styled.div`
   position: absolute;
-  z-index: 98;
+  z-index: 100;
```
nurm717123, line 10: I just think that since menu will be used on the app bar, it should shown above the drawer

src/components/menu/menu-popover.tsx
adrian-reapit: i think the menu tickbox version could require the popover to persist so the user can make selections

nurm717123: just found the feature that might allow checkbox or radio to close the menu here https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor

but as long as consumer are using it properly e.g ```<div role="menuitemradio">This is a radio div</div>``` I think it will be okay, I will make an example to include the role in next pr for radio and checkbox 👍

src/components/menu/menu.atoms.tsx
kurtdoherty, line 1: **issue:** I think we're missing a few important a11y props with these atoms. Have a read of https://www.w3.org/WAI/ARIA/apg/patterns/menubar/. Not all of the content will be relevant right now (e.g. submenus aren't in scope yet), but we should try to ensure we're keeping all the guidance there in mind.

src/components/menu/menu-popover.tsx
kurtdoherty: **note:** In reading https://www.w3.org/WAI/ARIA/apg/patterns/menubar/, the guidance for checkbox menu items states, when discussing keyboard interaction for the Space key:

> (Optional): When focus is on a menuitemcheckbox, changes the state without closing the menu.

This suggests that we may want that behaviour _by default_ for checkbox menu items, which increases the likelihood that we'll need to decide how to facilitate that behaviour.

src/components/menu/menu.atoms.tsx
```diff
@@ -0,0 +1,23 @@
...
+  )
+}
+
+export const MenuItemGroup: FC<
+  HTMLAttributes<HTMLUListElement> & {
+    title?: string
+  }
+> = ({ children, title, ...rest }) => {
+  return (
+    <ElMenuItemGroup {...rest}>
```
kurtdoherty, line 18: **todo:** I think groups should have the `role="group"` attribute. I can see reference to this in the ARIA docs for the [menuitemgroup](https://w3c.github.io/aria/#menuitemradio) role.

src/components/menu/styles.ts
```diff
@@ -18,3 +18,55 @@ export const ElMenuList = styled.div`
   background-color: var(--fill-white);
   box-shadow: 0px 4px 16px 0px #222b3329;
 `
```
kurtdoherty, line 20: **issue:** I don't see any handling of the following:

1. Disabled menu items (which according to https://www.w3.org/WAI/ARIA/apg/patterns/menubar/ is handled via `aria-disabled`);
2. Focus state for menu items.

src/components/menu/styles.ts
```diff
@@ -18,3 +18,55 @@ export const ElMenuList = styled.div`
...
+  display: flex;
+  height: 32px;
+  padding: var(--spacing-none, 0px) var(--spacing-4, 16px);
+  align-items: center;
+  align-self: stretch;
+`
+export const ElMenuItemGroup = styled.ul`
+  &,
+  & li > button {
+    width: 100%;
```
kurtdoherty, line 70: **question:** Why is this the `& li > button` selector here and the `width: 100%` not part of the menu item styles? Also, why does it only apply to `li > button` descendants and not the other kinds of descendants our items are currently styling?

**suggestion**: This may be a personal preference, but I find:

```css
width: 100%;
```

much easier to understand than:

```css
& {
  width: 100%;
}
```

kurtdoherty (commented):
Thanks @nurm717123 👍 I've finished my first review. Lots of questions and a few things to discuss. PR size is perfect though; thanks for keeping it small and focused 👌

src/components/menu/menu.atoms.tsx
```diff
@@ -0,0 +1,23 @@
+import { FC, HTMLAttributes } from 'react'
+import { ElMenuItemGroup, ElMenuItemGroupTitle } from './styles'
+
+export const MenuItemGroupTitle: FC<HTMLAttributes<HTMLDivElement>> = ({ children, ...rest }) => {
+  return (
+    <li>
+      <ElMenuItemGroupTitle {...rest}>{children}</ElMenuItemGroupTitle>
+    </li>
```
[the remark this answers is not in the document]
nurm717123: no I think just so that it has the same pattern with other `li` at that time, which is just for semantic purpose and styling applied to its child.
now I think for the title better to use the `li` directly

src/components/menu/styles.ts
```diff
@@ -7,7 +7,7 @@ export const ElMenu = styled.div`
 
 export const ElMenuPopover = styled.div`
   position: absolute;
-  z-index: 98;
+  z-index: 100;
 `
 
 export const ElMenuList = styled.div`
```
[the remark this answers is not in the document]
nurm717123: actually currently `Menu.Item` should under `Menu.Group` at the moment, because `Menu.List` could have multiple group, as the design imply in the menu `mixed` variants

src/components/menu/styles.ts
nurm717123: > question: Why is this the & li > button selector here and the width: 100% not part of the menu item styles?

actually it could be, I just think since only the button need that style it would be less code than to put additional three line in the item `button { width: 100% }` 

> why does it only apply to li > button descendants and not the other kinds of descendants our items are currently styling?

its because only the button which is not full width, it appear to have fit content when I hover the button and the style applied

> suggestion: This may be a personal preference, but I find:

yes, same like  above I just thought that to minimize the code I just chain it with the list > button style, but as the current pr is small I think I can update it, sorry for this😄

src/components/menu/menu-popover.tsx
nurm717123: > todo: @nurm717123 I think Adrian raises a good point. Can you please ask Design if they want that behaviour to be available?

ok

> This suggests that we may want that behaviour by default for checkbox menu items

yes, this is will be default as current behaviour

src/components/menu/styles.ts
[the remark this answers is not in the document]
kurtdoherty: The designs are not necessarily dev-focused and there's no detailed specs for menus, so we need to be careful drawing too many conclusions about how we then implement things. This kind of work typically requires a bit of research, discussion and collaboration with Design.

For example, my reading of the ARIA docs suggests items can be direct descendants of the `role='menu'` element, they do not need to be descendants of a group.

src/components/menu/styles.ts
nurm717123: > issue: I don't see any handling of the following:
Disabled menu items (which according to https://www.w3.org/WAI/ARIA/apg/patterns/menubar/ is handled via aria-disabled)

yes, I also don't see in the design spec have that at the moment, 

> Focus state for menu items

I think its because the planning it to be put it in the app bar, which close the menu on click,  I haven't tought about that, I will add it here using `:hover` style 👍

src/components/menu/menu.atoms.tsx
nurm717123: yes, since the menu could contain multiple group of checkbox, radio, button, .etc
thanks for pointing this out

src/components/menu/styles.ts
```diff
@@ -18,3 +18,55 @@ export const ElMenuList = styled.div`
...
+      background: var(--fill-action-dark);
+    }
+  }
+
+  /* customize html default button */
+  button {
+    background-color: transparent;
+    border: none;
+    border-radius: 0;
+  }
```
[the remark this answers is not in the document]
nurm717123: >  I'm really uncertain about the approach being taken here, specifically around the expectation for consumers to place a button inside a menu item. Here are my concerns/questions

for the context, I think that initially this menu will be used for app bar / navigation purpose

> Will this scale well? I've already identified <div role="button">, but what about other roles/variations on this same challenge?

yes, at the moment the styling only cover those 3 element defined in the style, but I think once other variants joining, which I think will also using div instead of input, its still fine 🤔 

> We want the whole menu item to be the hit area for pointer events, but it's not clear to me if this is currently the case given the elements that will have the click handlers are descendants of the menu item.

the custom handler will still be triggered (as long as the handler doesn't invoke `event.stopPropagation()`

> (Futurism) If we ever need menu items that look like an Element's button, these styles targeting any descendant <button> or <a> elements may get in the way (not 100% sure on this, as I guess it'll depend on rule specificity 🤔).

yes, I think that possible but I'm not sure if that is should be covered in here

src/components/menu/styles.ts
kurtdoherty: > yes, I also don't see in the design spec have that at the moment,

Yes, the design spec is not complete. That doesn't mean we don't need to facilitate standard concerns like disabled states and focus states.

> I think its because the planning it to be put it in the app bar, which close the menu on click, I haven't tought about that, I will add it here using :hover style 👍

Sorry, but I don't understand what the app bar has to do with `:focus` or `:focus-visible` styling of menu items in a menu. Can you clarify please?

src/components/menu/styles.ts
kurtdoherty: Yeah, okay. It's really important we don't decide what unit of code is responsible for what piece of logic or styling based on lines of code. We want components to have the correct responsibilities, and we want code to be readable by others (which may mean more lines of code, not less).

src/components/menu/styles.ts
nurm717123: > Yes, the design spec is not complete. That doesn't mean we don't need to facilitate standard concerns like disabled states and focus states.

👍 

> Sorry, but I don't understand what the app bar has to do with :focus or :focus-visible styling of menu items in a menu. Can you clarify please?

when user click the menu button in the appbar to trigger menu open, then a11y suggest to add a handler when user click the `arrowDown` key, it will focus to the first element in the menu, in that case `:focus` style will be applied, which in our case I will use `:hover` or `Selected` state in design, can I?

src/components/menu/styles.ts
nurm717123: I will update it👍

src/components/menu/styles.ts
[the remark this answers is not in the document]
nurm717123: I will adjust the code to use div, as per discussion below

src/components/menu/styles.ts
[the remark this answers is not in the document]
kurtdoherty: @nurm717123 I've been thinking more about the DOM structure within the menus and here's what I would propose we aim for in this PR:

1. No `<ul>` or `<li>` elements;
2. `Menu.List` renders to a `<div role="menu">`;
3. `Menu.Group` renders to a `<div role="group">`;
4. The group label renders as a `<div>`, maybe with `aria-hidden` (not sure, can always add that in future);
5. `Menu.Item` renders to a `<a role="menuitem">` or `<button role="menuitem">` depending on whether an `href` is provided (we're doing this for the Button work that is currently in PR and I think the same approach could work here); and,
6. We don't support or encourage nested interactive elements like `<button>` or `<a>` etc. Rather, the menu item is the thing that should be interactive.

These decisions would mean the DOM structure for a CSS only consumer's menu list might look something like the following snippets...

Simple:

```html
<div role="menu">
  <a role="menuitem">Link item</button>
  <button role="menuitem">Button item</button>
</div>
```

With a group:

```html
<div role="menu">
  <div role="group">
    <div>Some group</div>
    <a role="menuitem">Link item</button>
    <button role="menuitem">Button item</button>
  </div>
</div>
```

Some other comments on why I'm proposing this:
- From what I can tell from the ARIA docs, and looking at what other design systems have done, the element given the `menuitem` ARIA role is the element that should be interactive (not one of it's children);
- As such, it feels "safer" to me, if we are rendering `<a>` and `<button>` elements _as_ the menu items, not their children;
- This is just my proposal; there could be things I'm missing, so please let me know if you think there's problems with the above 🙏

src/components/menu/menu-popover.tsx
```diff
@@ -10,10 +10,10 @@ export const MenuPopover: FC = ({ children }) => {
...
 
   if (isOpen) {
-    const onClick = (e) => {
-      // close menu if button or any other element inside e.g icon is clicked
-      const button = e.target.closest('button')
-      if (button) closeMenu()
+    const onClick = (event) => {
+      // close menu if clicked target has parent with below selector
+      const menuItem = event.target.closest('[data-close-menu="true"]')
+      if (menuItem) closeMenu()
```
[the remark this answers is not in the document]
nurm717123: > suggestion: In future, when we discover an issue that has no obvious solution, it's really important that we discuss the tradeoffs involved in the different solutions we can think of before we go and submit code for review.

yes, I'll note this, thanks 👍 

>question: Can you please explain some of your reasoning behind the use of data-close-menu in the manner shown here and contrast it with other possible solutions?

certainly, so as we've discussed previously I think we need to give an options for menu item when it clicked whether it will close the menu or not. and this current pr the default behaviour is to close the menu by default when item clicked, and if user want to prevent that user can set ```<Menu.Item closeMenu={false}>...</Menu.Item>```.

the other alternatives I can think of is:
1. to put the `closeMenu` props in the menu group or the parent list instead so the all the grouped items will have the same behaviour, I think its better but less flexible.
2. to automatically decide the behaviour based on element used in the Items (e.g close the menu for buttons, div with role=button .etc). as currently I use div this is not applicable.
3. strictly define, e.g for simple/basic menu it will close but not for checkbox/radio menu. I'm not sure with this because I think consumer could want checbox to close immediately .etc

or is there other approach that you prefer ? 🤔

src/components/menu/styles.ts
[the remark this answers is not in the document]
nurm717123: thanks for the example, as I've already not use `ul` and `li` I think I just need to have a few amendment after this
1. move `role=menu` from popover to the list, I think its better because this way we can move out the popover if we want
2. use `a` or `button` depends on the `href` props, totally agree since it was my 1st proposal to achieve this (not in this repo) but the cons is the use might still put a interactive elements under it like button, if they are not aware of it, that's the reason, but I think that's not our fault.
3. the group with `aria-hidden`, I think we don't need this as we don't render the component at all when it is hidden, unless we use `display: none` approach.

src/components/menu/styles.ts
nurm717123: ok, I have confirmed with design team, and they agree https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu?node-id=83-2711&m=dev#1002341758

src/components/menu/menu-popover.tsx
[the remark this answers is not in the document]
kurtdoherty: Thanks for the explanation 🙏 

> is there other approach that you prefer?

Personally, I think leveraging `event.preventDefault()` is the cheapest, most flexible approach we could take for this right now.

1. It doesn't commit us to any specific component interface (like `closeMenu` currently does);
2. I suspect we can safely rely on `:where([role="menuitem"], [role="menuitemradio"], [role="menuitemcheckbox"]` as the selector we use with `event.target.closest`;
3. It doesn't risk confusing CSS-only consumers if they see `data-close-menu` in the docs and think that it works out-of-the-box for them; and,
4. It feels (subjectively) more natural to me than using a data attribute to affect whether the default click behaviour occurs or not. That is, `event.preventDefault()` seem specifically geared towards this kind of thing (to me at least).

That said, what you've done here is still a simple solution and while it commits us to the `closeMenu` prop, it's only one prop, and it wouldn't be hard to migrate away from in future if we had to. So I think the main concern (which may not be important enough to warrant changing what you've done) is point (3): possible confusion for CSS only consumers 🤷‍♂️ 

What do you think?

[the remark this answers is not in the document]
kurtdoherty: **aside:** Also, when we get access to the `popover` attribute in React 19, we'd be able to support a full HTML-only solution (for button-based menu items, at least) using something like:

```html
<div id="123" role="menu" popover>
  <button role="menuitem" popovertarget="123" popovertargetaction="hide">Item 1</button>
  <button role="menuitem">Item 2</button>
</div>
```

In this case, Item 1 would close the menu popover when clicked, but Item 2 would not.

kurtdoherty (approved):
Thanks for all the work on this one @nurm717123. There's been some really good discussion as we've tried to find out way through the a11y concerns 👍 

I don't believe there's any outstanding blocking issues so I'm approving. This means I'm fine to leave it up to you to decide which way we fall for the menu closing behaviour. At this point, I think it's pretty low consequence given I suspect we'll want to leverage the popover API when we get on to React 19.

src/components/menu/menu-popover.tsx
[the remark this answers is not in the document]
nurm717123: > Personally, I think leveraging event.preventDefault() is the cheapest, most flexible approach we could take for this right now.

I think there is a misunderstanding here, because after I tried `preventDefault()` will have no effect since the onclick handler assigned to the child will run first

image

and that is what I expect because my Idea is actually to expand the `onClick` handler of the `Menu.Item` so we don't have to handle in each of them

> I suspect we can safely rely on :where([role="menuitem"], [role="menuitemradio"], [role="menuitemcheckbox"] as the selector we use with event.target.closest;

as currently `data-close-menu` is always on `Menu.Item` I think its enough already for the selector

> It doesn't risk confusing CSS-only consumers if they see data-close-menu in the docs and think that it works out-of-the-box for them; and,

yes, I think I need to add some notes in the docs

> What do you think?

because the docs for css-only user don't show the `data-close-menu` I think it is safe to use my current approach, I'm not sure if there is better aproach than using popover api like what you've mention before 🤔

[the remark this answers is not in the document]
kurtdoherty: > I tried preventDefault() will have no effect since the onclick handler assigned to the child will run first

I think the misunderstanding may be that it's the child click handler that would call `event.preventDefault()`, and the parent click handler would check if `event.defaultPrevented` is true or not before closing the menu.

All good though. Like I said, you can make the call on this one and, hopefully sometime soon, we'll get all the nice new native functionality from the popover attribute 😄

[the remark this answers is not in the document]
nurm717123: great, thanks for the comment so far it really give me more insights😄 

I will make the amendments as per the discussion previously 👍

PR #185, opened 2024-10-24
fix: #184 index.css correctly includes design tokens

### LINKED ISSUE #184: `index.css` build artefact does not correctly include design tokens CSS
**Describe the bug**
With the implementation of design tokens in Elements, we updated `globals.ts` to `@import` the generated `reapit/tokens.css` file. Unfortunately, we did not notice that this `@import` is not resolved when Elements is built, because it's contained within a Linaria `css` block. As a result, the generated `dist/index.css` for Elements contains `@import url('./../../tokens/reapit/tokens.css')`, yet this referenced file is not part of the package's distribution artefacts, so it cannot be resolved.

The image below shows what happens in the Console Cloud monorepo when we try to build a package that depends on the latest version of Elements.

Screenshot 2024-10-25 at 8 26 07 AM

**Expected behaviour**
The generated `dist/index.css` file should either have the _content_ of the `reapit/tokens.css` file, or that file should be included in the distribution artefacts and the `@import` statement rewritten to correctly resolve it.

kurtdoherty:
Fixes #184 

Changes from using an `@import url(...)` statement within the global `css` block to using a bare ES module import so that the token.css file's content is correclty included in the Rollup build. See #184 for more details.

**Before:**

Screenshot 2024-10-25 at 8 55 14 AM

**After:**

Screenshot 2024-10-25 at 8 45 01 AM

### DIFF
--- a/src/styles/globals.ts
+++ b/src/styles/globals.ts
@@ -1,4 +1,6 @@
 import { css } from '@linaria/core'
+import '../../tokens/reapit/tokens.css'
+
 // export javascript colours for the SVG background in the button
 export const intentPrimary = '#4e56ea'
 export const intentNeutral = '#0080ff'
@@ -15,8 +17,6 @@ export const intentSecondary = '#4e56ea'
 export const elGlobals = css`
   :global() {
     @import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&family=Source+Code+Pro&display=swap');
-    /* Root color pallet */
-    @import url('./../../tokens/reapit/tokens.css');
     /* Reset CSS */
     html,
     body,


src/styles/globals.ts
```diff
@@ -1,4 +1,6 @@
 import { css } from '@linaria/core'
+import '../../tokens/reapit/tokens.css'
```
kurtdoherty, line 2: **note:** by using a bare module import like this, we enable our build tooling to be aware of this CSS module such that it will be processed and included in the final `index.css` artefact generated by a build.

src/styles/globals.ts
```diff
@@ -15,8 +17,6 @@ export const intentSecondary = '#4e56ea'
 export const elGlobals = css`
   :global() {
     @import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&family=Source+Code+Pro&display=swap');
-    /* Root color pallet */
-    @import url('./../../tokens/reapit/tokens.css');
```
kurtdoherty, line 19: **note:** Since we're now including the token.css via our build process, we no longer need to worry about trying to included it here via Linaria's global CSS approach.

kurtdoherty:
FYI @willmcvay, you may be interested in this.

ksolanki7:
**praise:** The PR looks good to me. I have tested it, and it functions as expected.

ksolanki7 (approved):

Issue #159, opened 2024-08-21
Replace Jest with Vite + Vitest


kurtdoherty:
### Context

TL;DR Jest is hard to configure so that the Linaria CSS extraction on components is performed during test runs. Switching to a Vite + Vitest combination is, by comparison, trivial to achieve. This CSS extraction is important when rendering components via React Testing Library (RTL), because without it, the correct styles are not applied to the component under test, undermining many RTL DOM queries.

More details on this issue can be found in #152.

### Considerations

While the setup of Vite and Vitest is trivial, it does have a number of important considerations.

1. @adrian-reapit reported in #152 that they've seen poor performance from Vitest in GitHub Actions. While the Console FE team hasn't had issues in our CI, we're currently using Jenkins, not GitHub Actions, so it's not immediately clear if we'll see performance problems running Vitest in Elements'.
2. Vitest does not provide global functions by default like Jest does. Rather, Vitest prefers explicit imports for utilities like `test`, `describe` and `vi` (the equivalent of `jest`). That said, if we don't like the idea of having to add explicit imports for test functions, we can use Vitest's `globals` option to enable them as globals as per Jest's behaviour.
3. Since Vitest is ESM first, setting up test doubles for a module's default export looks a little different with Vitest's `vi.mock` than it does with Jest's `jest.mock`. This is not a problem, just a matter of education for contributors.
4. There's no easy way (that I'm aware of) to allow an incremental migration of test files from Jest to Vitest, so I believe this will need to be big bang cut-over. The Console FE team has successfully done this across our significant number of packages that back Console Cloud, so I don't believe it will be too difficult.
5. If we use Vite + Vitest for our unit tests, we'll also be able to use the same Vite config to build Elements, instead of our current direct use of Rollup. I would consider this to be out-of-scope for this work and, if we go ahead with this proposal, would prefer to see us replace our Rollup config separately.

### Proposal

The following work would be done as a single, large PR:

1. Add Vite, Vitest and Happy DOM as dev dependencies, and setup the Vite + Vitest config;
2. Rewrite all tests to use `vi` instead of `jest` and import `test`, `describe`, etc from Vitest (unless we would prefer to turn on globals?);
3. Update our `test` package scripts to run Vitest instead of Jest;
4. Remove all Jest-related config.

[...]

Issue #158, opened 2024-08-21
Remove Linaria mocks from Jest global setup


kurtdoherty:
### Context

TL;DR The global mocks we currently setup in `scripts/jest/jest-setup.js` for Linaria's `css` and `styled` utilities interfere with React Testing Library's DOM queries because they result in non-standard tag names being used (e.g. `<mock-styled.textarea>`).

This blocks our ability to write effective unit tests for components, effectively restricting us to the sole use of DOM snapshots.

The mocks for Linaria were required with previous versions of Linaria, but they are no longer necessary. See #152 for more details.

### Work to be done

I see two main options here:

1. Like #157, we remove the mocks and do a big bang update for all the snapshots; or,
2. We defer this work in favour of switching to Vitest (at which point, the Jest mocks for Linaria won't be relevant, and the snapshot will be replaced with Vitest-written versions). This change is proposed in #159.

At this stage, I'm in favour of (2), but that's based on the assumption that we'll actually agree to switch from Jest to Vitest.

PR #190, opened 2024-11-04
feat: #163 new menu (keyboard interaction)

nurm717123:
continuing the sequence of pr for menu that have been discussed https://github.com/reapit/elements/issues/163, this is the 4th pr which only focus on keyboard interaction based on https://www.w3.org/WAI/ARIA/apg/patterns/menubar, here I provide basic keyboard interaction only for menu and menu items related component.

an exception here: since the menu doesn't have sub-menu, the keyboard interaction only focus to the menu-button and menu items

# Pull request checklist
- [x] button `arrowDown` handler when menu open
- [x] menu item `arrowUp` and `arrowDown` handler
- [x] menu item `escape` handler
- [x] menu item `enter` and `space` handler
- [x] menu item `tab` handler (close menu when focus move out of menu)

- new menu keyboard interaction

menu item custom handler

### DIFF
--- a/src/components/menu/__tests__/menu-popover.test.tsx
+++ b/src/components/menu/__tests__/menu-popover.test.tsx
@@ -16,7 +16,7 @@ describe('Menu Popover component', () => {
       </Menu>
     )
   }
-  
+
   it('should render and close Popover based on any element with data-close-menu="true" click and match snapshot', () => {
     const { asFragment, getByText } = render(<MockMenuPopoverComponent />)
 
--- a/src/components/menu/__tests__/menu-popover.test.tsx
+++ b/src/components/menu/__tests__/menu-popover.test.tsx
@@ -16,7 +16,7 @@ describe('Menu Popover component', () => {
       </Menu>
     )
   }
-
+  
   it('should render and close Popover based on any element with data-close-menu="true" click and match snapshot', () => {
     const { asFragment, getByText } = render(<MockMenuPopoverComponent />)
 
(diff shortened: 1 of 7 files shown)


src/components/menu/styles.ts
```diff
@@ -38,6 +38,9 @@ const baseMenuItemStyles = `
     color: var(--text-white);
     background: var(--fill-action-dark);
   }
+  &:focus-visible {
+    outline: none;
+  }
```
nurm717123, line 43: the button have default outline color when focused (using `tab` key for example), but anchor doesn't have, so this is just to make the style the same (use defined `:focus` style)

src/components/menu/menu-popover.tsx
```diff
@@ -1,14 +1,79 @@
...
+    }
+  }
+
 export const MenuPopover: FC = ({ children }) => {
   const { isOpen, closeMenu, getPopoverProps } = useMenuContext()
   const popoverRef = useRef<HTMLDivElement>(null)
 
   useClickOutside(popoverRef, closeMenu)
 
+  useEffect(() => {
```
kurtdoherty, line 55: **issue:** It's really important we make the right unit of code responsible for the right behaviours, and I don't think it's quite right at the moment.

If we breakdown the responsibilities of each component within our Menu molecule, I think it would look something like this:

- `Menu.Item`: responsible for the presentational concerns of an item in the menu;
- `Menu.Group`: responsible for the presentational concerns of labelling a group of items in the menu;
- `Menu.Separator` (not implemented yet): responsible for the presentational concerns of a divider/separator in the menu;
- `Menu.List`: responsible for laying out menu items/groups in a vertical orientation;
- `Menu.Popper`: responsible for disclosing/hiding the menu list;
- `Menu.Trigger`: responsible for opening the menu when clicked;
- `useMenu`: responsible for providing the "glue" (state like `isOpen` and prop getters like `getTriggerProps`) that allow the menu atoms to function as a cohesive whole;
- `Menu`: responsible for integrating `useMenu` with `Menu.Context`.

Given this breakdown, is there a unit that stands out to you as the most appropriate place to implement extra behaviour (via event callbacks like `onKeyDown` and `onFocusOut`) we need for keyboard navigation?

src/components/menu/menu-popover.tsx
```diff
@@ -1,14 +1,79 @@
...
+    if (isOpen) {
+      const menuContainer = popoverRef.current?.parentElement
+      const menuItems = menuContainer!.querySelectorAll('[role="menuitem"]') as NodeListOf<HTMLElement>
+      const menuButton = menuContainer!.querySelector('[role="button"][aria-expanded="true"]') as HTMLButtonElement
+      const menuList = menuContainer?.querySelector('[role="menu"]') as HTMLDivElement
+
+      const controller = new AbortController()
+      const { signal } = controller
+
```
kurtdoherty, line 64: **issue:** While I can see this all works, it doesn't feel very "React-like". I think I would have probably expected to see React event callbacks like `onKeyDown` and `onFocusOut` being used instead of attaching plain DOM event handlers to DOM elements directly.

I'm happy to provide a specific suggestion on this front if you need it, but I'll let you respond to my comment above first 👍

src/components/menu/menu-popover.tsx
```diff
@@ -1,14 +1,79 @@
...
+      menuList.addEventListener('focusout', menuListFocusOutHandler(menuList, closeMenu), { signal })
+      menuButton.addEventListener('keydown', menuButtonKeyDownHandler(menuItems), { signal })
+      menuItems.forEach((menuItem, index) => {
```
kurtdoherty, line 67: **question:** instead of needing to attach specific event handler to each menu item, can the menu list have a single handler that catches keydown events that bubble up from the items?

src/components/menu/menu-popover.tsx
```diff
@@ -1,14 +1,79 @@
...
+        menuItem.addEventListener('keydown', menuItemKeyDownHandler(menuButton, menuItems, index, closeMenu), {
+          signal,
+        })
+      })
+
+      return () => controller.abort()
```
kurtdoherty, line 73: **praise:** I'm loving the use of AbortController for cleanup 👏  It's looks much simpler than having to remove them all individually!

src/components/menu/menu-popover.tsx
```diff
@@ -1,14 +1,79 @@
-import { FC, MutableRefObject, useRef } from 'react'
+import { FC, MutableRefObject, useEffect, useRef } from 'react'
 import { useClickOutside } from '../../hooks/use-click-outside'
 import { useMenuContext } from './menu-context'
 import { ElMenuPopover } from './styles'
 
+// close menu when focus move out of menu
+export const menuListFocusOutHandler = (menuList: HTMLDivElement, closeMenu: VoidFunction) => (event) => {
+  if (!menuList.contains((event as FocusEvent).relatedTarget as Node)) {
```
kurtdoherty, line 8: **question:** typically the use type assertions is viewed as bad practice, unless there's just no way of getting TS to understand the types involved in our code). Can you help me understand why all these type assertions are needed?

src/components/menu/menu-popover.tsx
```diff
@@ -1,14 +1,79 @@
-import { FC, MutableRefObject, useRef } from 'react'
+import { FC, MutableRefObject, useEffect, useRef } from 'react'
```
kurtdoherty, line 1: **note:** It's best practice to ensure type-only imports (like FC and MutableRefObject) use the `type` modifier. For example,

```typescript
import { useEffect, useRef } from 'react'
import type { FC, MutableRefObject } from 'react'
// OR...
import { type FC, type MutableRefObject, useEffect, useRef } from 'react' 
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export for more background.

src/components/menu/__tests__/menu-popover.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { fireEvent, render } from '@testing-library/react'
 import { Menu } from '../menu'
+import { menuButtonKeyDownHandler, menuItemKeyDownHandler, menuListFocusOutHandler } from '../menu-popover'
```
kurtdoherty, line 3: **suggestion:** Personally, I think the desire to test these handlers independently from the menu popper component suggests they should be in their own module.

I also wonder how much value there is in testing these handlers in a solitary (non-integrated) manner 🤔

kurtdoherty (commented):
Thanks @nurm717123, some really great work here 🙌  My main feedback right now is at the code design level: I don't think the menu popper should be responsible for our keyboard navigation logic, but keen to hear your thoughts!

src/components/menu/menu-popover.tsx
nurm717123: > issue: It's really important we make the right unit of code responsible for the right behaviours, and I don't think it's quite right at the moment.

thanks for describing this 👍 , I thought that it would be nice to centralize the keyboard handler logic but now I think It would be better to place it own related component

> Given this breakdown, is there a unit that stands out to you as the most appropriate place to implement the extra behaviour (via event callbacks like onKeyDown and onFocusOut) we need for keyboard navigation?

yes, I think that would be `Menu.Item` and `Menu.List` respectively

src/components/menu/menu-popover.tsx
nurm717123: totally agree, once I moved it I would use the props instead 👍

src/components/menu/menu-popover.tsx
nurm717123: I think that is also possible, I'll try it, that is good Idea

src/components/menu/menu-popover.tsx
nurm717123: yeah, I'm starting to like it too, thanks for the suggestion 👍

src/components/menu/menu-popover.tsx
nurm717123: apoligize, I forgot to delete it after I removed the event type 🙏

src/components/menu/__tests__/menu-popover.test.tsx
nurm717123: actually that pattern is to make the test easier because currently we don't install `@testing-library/jest-dom` (which I think we need it e.g: to test focus called), and also it avoiding snapshot test (eg: testing visiblity of menu). by using handler like that I can mock the handler and test it easily.

kurtdoherty: Ahh, yeah okay. I keep forgetting that restriction still exists 🤦‍♂️ I'll see if I can get buy-in for #158 and #159 from Will and Adrian so that we can start leveraging React Testing Library properly.

src/components/menu/menu-popover.tsx
kurtdoherty: > I think that would be getTriggerProps (for button keyDown), Menu.Item and Menu.List respectively, but I would like to try if Menu.List can handle both

Menu.List could be a reasonable choice and I won't block that approach if you feel strongly about it, but I'd suggest `useMenu` as an even better option.

My reasoning is:

- `Menu.List` is currently a presentational component. If we add keyboard handling to it, we no longer have a purely presentational component. This isn't necessarily bad, but..
- `useMenu` is already responsible for wiring up specific behaviour to specific atoms via its prop getters (like `getTriggerProps` providing an `onClick` to the menu trigger). It makes sense to do the same for the menu list (e.g. providing a `getMenuListProps` prop getter that wires up the necessary `onKeyDown`, `onFocusOut` callbacks). To put it another way, this fits in neatly with its existing responsibility as opposed to adding a new responsibility to `Menu.List`).
- By keeping behaviour inside `useMenu`, it's easy for us to swap out the presentational components with new ones in future without having to reimplement all the keyboard navigation/focus management behaviour.

nurm717123: Ahh yes, I have overlooked the `getPopoverProps` 😞. I totally agree.

src/hooks/use-menu/index.tsx
```diff
@@ -33,6 +46,48 @@ export const useMenu = (): useMenu => {
     return {
       ...props,
       'data-open': isOpen,
```
kurtdoherty, line 48: **note:** No need to do this now, but I want to flag that I think this hook has enough complexity now to warrant splitting out each prop getter to its own module (e.g. a custom hook or factory function) so that useMenu's implementation becomes one of "delegation". To give you and idea of what this _could_ look like (without much thought):

```javascript
function useMenu() {
  const [isOpen, setIsOpen] = useState(false)
  const getTriggerProps = useGetTriggerProps(...)
  const getTriggerProps = useGetPopoverProps(...)
  // and so on...
}
```

src/hooks/use-menu/index.tsx
```diff
@@ -20,6 +20,19 @@ export const useMenu = (): useMenu => {
...
+          const menuItems = e.currentTarget.parentElement!.querySelectorAll(
+            '[role="menuitem"]',
+          ) as NodeListOf<HTMLElement>
+
+          if (menuItems.length) menuItems[0].focus()
+        }
+
+        if (props?.onKeyDown) {
+          props.onKeyDown(e)
+        }
```
kurtdoherty, line 34: **note:** Not a big deal, but we can write `props?.onKeyDown?.(e)` instead of needing the if-statement to assert the function is defined before calling. That is, the optional chaining operator (`?.`) can apply to function calls, not just property access.

src/hooks/use-menu/index.tsx
kurtdoherty, line 1: **note:** I did not realise `useMenu` was off in a completely separate folder to Menu! It's all good to keep it this way for your work here, but if we get a chance to in future, we should move it so it's co-located with the Menu component.

kurtdoherty (approved):
Nice work @nurm717123 👏

src/hooks/use-menu/index.tsx
nurm717123: thanks for the tips 👍

nurm717123:
Thanks Kurt 👍

Issue #164, opened 2024-09-17
v5 new components: NavIconItem


adrian-reapit:
**Abstract**

To support the **AppBar** improvements the **NavIconButton** and **NavIconMenuButton** components should be made which can be used in secondary and footer navigation

![Image](https://github.com/user-attachments/assets/7f4433cf-7de3-432a-80cb-114c2ba401b2)

![Image](https://github.com/user-attachments/assets/231f0d86-4369-4f95-b638-fb1302c6efd0)

![Image](https://github.com/user-attachments/assets/858ff5d1-fc1f-4008-a632-580b3dc1cc78)

**NavIconButton** is a button with a centred icon and optional centred text below. It can be a private component used by **AppBar**

- icon and text highlight can be controlled 
- a notification badge can be shown
- the badge is an icon and so we could allow other icons and intent to be shown 

**NavIconMenuButton** is an extended version of **NavIconButton** which shows the menu above or below

**Specification**

- Is new component or existing? New
- Is the component design signed off by product and design?  Signed off as part of the AppBar changes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to AppBar design in Figma](https://github.com/orgs/reapit/projects/16/views/1?pane=issue&itemId=78443969) 

 
**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into Styles Only and React component structures
- [ ] Ensure all variants of components are documented as appropriate
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

kurtdoherty:
**issue:** @azkanurf-ss I think you may be confusing two different components and treating them as one. The "Nav Icon Item" is exclusively used for the Secondary Nav and only has a visible label when hovered or focused (i.e. the when the tooltip is displayed), while the screenshot you show of an icon + visible label is the "Bottom Bar Item" and is used exclusively for the Bottom Bar.

[...]

PR #191, opened 2024-11-04
feat: #164 - add nav icon item component

ss-dimasm:
# Pull request checklist

Following on the discussion on #164 , we've made an agreement on separating the `NavIconItem` and `BottomBarItem` as a different component

**Context:**
The `NavIconItem` component is a small part of the `AppBar` component. It appears as a secondary navigation item on the AppBar when the screen width exceeds 1440 pixels (see the placement on the image attachment)

This PR initiates the construction of the `NavIconItem` component. It prioritizes the implementation of fundamental states (default, focus, hover, and active) while deferring the addition of supplementary elements such as tooltips and indicators. Further deliberation with the design team is required to finalize these details

**View Preview:**

https://github.com/user-attachments/assets/8d851872-3769-492d-b217-ca153ec2b526

**[Design Specification](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4749&node-type=frame&m=dev):**
{756E86AF-D52F-4B0C-B90F-09233B57CF15}

**[Future placement on AppBar component](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=15-3629&node-type=frame&m=dev):**
image

### DIFF
--- a/src/components/nav-icon-item/nav-icon-item.tsx
+++ b/src/components/nav-icon-item/nav-icon-item.tsx
@@ -1,0 +1,63 @@
+import { AnchorHTMLAttributes, ButtonHTMLAttributes, FC, MouseEventHandler, ReactNode } from 'react'
+import { ElAnchorNavIconItem, ElButtonNavIconItem } from './styles'
+
+interface CommonNavIconItemProps {
+  /**
+   * The JSX element to render as the icon
+   */
+  icon: ReactNode
+
+  /**
+   * Whether the nav item is active
+   * @default false
+   **/
+  isActive?: boolean
+}
+
+interface NavIconItemAsButtonElementProps extends CommonNavIconItemProps, ButtonHTMLAttributes<HTMLButtonElement> {
+  href?: never
+  target?: never
+  ref?: never
+  onClick: MouseEventHandler<HTMLButtonElement>
+}
+
+interface NavIconItemAsAnchorElementProps extends CommonNavIconItemProps, AnchorHTMLAttributes<HTMLAnchorElement> {
+  href?: string
+  target?: string
+  rel?: string
+  onClick?: MouseEventHandler<HTMLAnchorElement>
+}
+
+export type NavIconItemProps = NavIconItemAsButtonElementProps | NavIconItemAsAnchorElementProps
+
+const isButtonElement = (props: NavIconItemProps): props is NavIconItemAsButtonElementProps => {
+  return !props.href
+}
+
+export const NavIconItem: FC<NavIconItemProps> = (props) => {
+  if (!isButtonElement(props)) {
+    const { isActive, onClick, icon, ...rest } = props ?? {}
+
+    return (
+      <ElAnchorNavIconItem
+        {...rest}
+        role="button"
+        data-state={isActive ? 'active' : undefined}
+        onClick={(e) => {
+          e.preventDefault()
+          onClick?.(e)
+        }}
+      >
+        {icon}
+      </ElAnchorNavIconItem>
+    )
+  } else {
+    const { isActive, icon, ...rest } = props ?? {}
+
+    return (
+      <ElButtonNavIconItem {...rest} role="button" data-state={isActive ? 'active' : undefined}>
+        {icon}
+      </ElButtonNavIconItem>
+    )
+  }
+}
--- a/src/components/nav-icon-item/nav-icon-item.tsx
+++ b/src/components/nav-icon-item/nav-icon-item.tsx
@@ -8,20 +8,29 @@ interface CommonNavIconItemProps {
   icon: ReactNode
 
   /**
+   * Defines a string value that labels the current element
+   */
+  'aria-label': string
+
+  /**
    * Whether the nav item is active
    * @default false
    **/
   isActive?: boolean
 }
 
-interface NavIconItemAsButtonElementProps extends CommonNavIconItemProps, ButtonHTMLAttributes<HTMLButtonElement> {
+interface NavIconItemAsButtonElementProps
+  extends CommonNavIconItemProps,
+    Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'aria-label'> {
   href?: never
   target?: never
-  ref?: never
+  rel?: never
   onClick: MouseEventHandler<HTMLButtonElement>
 }
 
-interface NavIconItemAsAnchorElementProps extends CommonNavIconItemProps, AnchorHTMLAttributes<HTMLAnchorElement> {
+interface NavIconItemAsAnchorElementProps
+  extends CommonNavIconItemProps,
+    Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'aria-label'> {
   href?: string
   target?: string
   rel?: string
@@ -35,29 +44,21 @@ const isButtonElement = (props: NavIconItemProps): props is NavIconItemAsButtonE
 }
 
 export const NavIconItem: FC<NavIconItemProps> = (props) => {
-  if (!isButtonElement(props)) {
-    const { isActive, onClick, icon, ...rest } = props ?? {}
+  if (isButtonElement(props)) {
+    const { isActive, icon, ...rest } = props ?? {}
 
     return (
-      <ElAnchorNavIconItem
-        {...rest}
-        role="button"
-        data-state={isActive ? 'active' : undefined}
-        onClick={(e) => {
-          e.preventDefault()
-          onClick?.(e)
-        }}
-      >
+      <ElButtonNavIconItem {...rest} aria-current={isActive ? 'true' : undefined}>
         {icon}
-      </ElAnchorNavIconItem>
+      </ElButtonNavIconItem>
     )
   } else {
     const { isActive, icon, ...rest } = props ?? {}
 
     return (
-      <ElButtonNavIconItem {...rest} role="button" data-state={isActive ? 'active' : undefined}>
+      <ElAnchorNavIconItem {...rest} aria-current={isActive ? 'page' : undefined}>
         {icon}
-      </ElButtonNavIconItem>
+      </ElAnchorNavIconItem>
     )
   }
 }
--- a/src/components/nav-icon-item/nav-icon-item.tsx
+++ b/src/components/nav-icon-item/nav-icon-item.tsx
@@ -23,8 +23,6 @@ interface NavIconItemAsButtonElementProps
   extends CommonNavIconItemProps,
     Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'aria-label'> {
   href?: never
-  target?: never
-  rel?: never
   onClick: MouseEventHandler<HTMLButtonElement>
 }
 
@@ -32,8 +30,6 @@ interface NavIconItemAsAnchorElementProps
   extends CommonNavIconItemProps,
     Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'aria-label'> {
   href?: string
-  target?: string
-  rel?: string
   onClick?: MouseEventHandler<HTMLAnchorElement>
 }
 
@@ -43,6 +39,12 @@ const isButtonElement = (props: NavIconItemProps): props is NavIconItemAsButtonE
   return !props.href
 }
 
+/**
+ * The `NavIconItem` component is a small part of `AppBar` component
+ * It's used to render a nav item with an icon
+ *
+ * To ensure it's aligned with the design, the children component should have a CSS color property with 'inherit' value
+ */
 export const NavIconItem: FC<NavIconItemProps> = (props) => {
   if (isButtonElement(props)) {
     const { isActive, icon, ...rest } = props ?? {}
--- a/src/components/nav-icon-item/styles.ts
+++ b/src/components/nav-icon-item/styles.ts
@@ -2,6 +2,8 @@ import { AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react'
 import { styled } from '@linaria/react'
 import { ElIcon } from '../icon'
 
+// TODO: add red dot integration
+// TODO: add tooltip integration on hover state
 const baseStyles = `
   display: inline-flex;
   padding: var(--space-2, 8px);
--- a/src/components/nav-icon-item/styles.ts
+++ b/src/components/nav-icon-item/styles.ts
@@ -1,0 +1,61 @@
+import { styled } from '@linaria/react'
+import { AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react'
+
+type CommonNavIconItemProps = {
+  /**
+   * NOTE: data attribute to indicate the state of the nav item
+   * by default the styles is already applied with pseudo-class selector
+   * but to make it more flexible, consumer could use this attribute to
+   * change the styles
+   */
+  'data-state'?: 'focus' | 'hover' | 'active'
+}
+
+const baseStyles = `
+  display: inline-flex;
+  padding: var(--spacing-2, 8px);
+  justify-content: center;
+  align-items: center;
+  gap: var(--spacing-none, 0px);
+  border-radius: var(--corner-default, 0.25rem);
+  background: var(--white, #ffffff);
+  border: none;
+  outline: none;
+
+  /* change children color by default */
+  > * {
+    color: var(--neutral-400, #798da1) !important;
+  }
+
+  &:focus, &[data-state='focus'] {
+    border-radius: var(--corner-default, 0.25rem);
+    background: var(--white, #ffffff);
+    box-shadow:
+      0px 0px 0px 1px #fff,
+      0px 0px 0px 4px var(--icon-button_primary-hover, #7e9bfa);
+  }
+
+  &:hover, &[data-state='hover'] {
+    cursor: pointer;
+    border-radius: var(--corner-default, 0.25rem);
+    background: var(--neutral-050, #f2f4f6);
+  }
+
+  &:active, &[data-state='active'] {
+    > * {
+      color: var(--intent-primary, #4e56ea) !important;
+    }
+    border-radius: var(--corner-default, 0.25rem);
+    background: var(--neutral-050, #f2f4f6);
+  }
+`
+
+type ElButtonNavIconItemProps = ButtonHTMLAttributes<HTMLButtonElement> & CommonNavIconItemProps
+export const ElButtonNavIconItem = styled.button<ElButtonNavIconItemProps>`
+  ${baseStyles}
+`
+
+type ElAnchorNavIconItemProps = AnchorHTMLAttributes<HTMLAnchorElement> & CommonNavIconItemProps
+export const ElAnchorNavIconItem = styled.a<ElAnchorNavIconItemProps>`
+  ${baseStyles}
+`
--- a/src/components/nav-icon-item/styles.ts
+++ b/src/components/nav-icon-item/styles.ts
@@ -1,61 +1,52 @@
-import { styled } from '@linaria/react'
 import { AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react'
-
-type CommonNavIconItemProps = {
-  /**
-   * NOTE: data attribute to indicate the state of the nav item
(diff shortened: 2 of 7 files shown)


src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+   * Whether the nav item is active
+   * @default false
+   **/
+  isActive?: boolean
+}
+
+interface NavIconItemAsButtonElementProps extends CommonNavIconItemProps, ButtonHTMLAttributes<HTMLButtonElement> {
+  href?: never
+  target?: never
+  ref?: never
```
kurtdoherty, line 20: **question:** why is `ref` being prevented?

src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Whether the nav item is active
+   * @default false
+   **/
+  isActive?: boolean
+}
+
+interface NavIconItemAsButtonElementProps extends CommonNavIconItemProps, ButtonHTMLAttributes<HTMLButtonElement> {
+  href?: never
+  target?: never
```
kurtdoherty, line 19: **question:** is this to try and get a proper discriminated union happening? Just seems odd that we're explicitly excluding props that, at face value, wouldn't be part of this type anyway (i.e. `ButtonHTMLAttributes<HTMLButtonElement>` doesn't define either of these, right?)

src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export const NavIconItem: FC<NavIconItemProps> = (props) => {
+  if (!isButtonElement(props)) {
+    const { isActive, onClick, icon, ...rest } = props ?? {}
+
+    return (
+      <ElAnchorNavIconItem
+        {...rest}
+        role="button"
```
kurtdoherty, line 44: **question:** Is `role=button` actually correct here? Shouldn't the `<a>` just have its normal implicit role?

src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export type NavIconItemProps = NavIconItemAsButtonElementProps | NavIconItemAsAnchorElementProps
+
+const isButtonElement = (props: NavIconItemProps): props is NavIconItemAsButtonElementProps => {
+  return !props.href
+}
+
+export const NavIconItem: FC<NavIconItemProps> = (props) => {
+  if (!isButtonElement(props)) {
```
kurtdoherty, line 38: **quibble:** for readability, I personally find it best to keep simple if-else blocks like this one written so that the if-condition is _not_ negated. That is, I find

```javascript
if (isButtonElement(props)) {
  // a button element
} else {
  // not a button element
}
```

Alternatively, changing `isButtonElement` to `isAnchorElement` would also make it more readable, because our if-condition does not involve a negation.

src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const { isActive, onClick, icon, ...rest } = props ?? {}
+
+    return (
+      <ElAnchorNavIconItem
+        {...rest}
+        role="button"
+        data-state={isActive ? 'active' : undefined}
```
kurtdoherty, line 45: **suggestion:** I think [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) may be preferable; specifically when `isActive` is true, an anchor element would have `aria-current="page"` or a button element would have `aria-current="true"`.

Would also be reasonable to have a separate data attribute if it somehow makes the CSS styling simpler to achieve, but I suspect we could just style based on the presence of `aria-current` 🤔

src/components/nav-icon-item/styles.ts
```diff
@@ -0,0 +1,61 @@
+import { styled } from '@linaria/react'
+import { AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react'
+
+type CommonNavIconItemProps = {
+  /**
+   * NOTE: data attribute to indicate the state of the nav item
+   * by default the styles is already applied with pseudo-class selector
+   * but to make it more flexible, consumer could use this attribute to
+   * change the styles
```
kurtdoherty, line 9: **question:** Why do we need to have this flexibility for consumers? I'm struggling to see the value here 🤔

src/components/nav-icon-item/styles.ts
```diff
@@ -0,0 +1,61 @@
...
+  background: var(--white, #ffffff);
+  border: none;
+  outline: none;
+
+  /* change children color by default */
+  > * {
+    color: var(--neutral-400, #798da1) !important;
+  }
+
+  &:focus, &[data-state='focus'] {
```
kurtdoherty, line 30: **suggestion:** I _think_ (not 100% sure) we would prefer `:focus-visible`, instead of `:focus`, as the focus outline really only needs to appear when focus is applied via keyboard interaction, not pointers.

src/components/nav-icon-item/styles.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  &:hover, &[data-state='hover'] {
+    cursor: pointer;
+    border-radius: var(--corner-default, 0.25rem);
+    background: var(--neutral-050, #f2f4f6);
+  }
+
+  &:active, &[data-state='active'] {
+    > * {
+      color: var(--intent-primary, #4e56ea) !important;
```
kurtdoherty, line 46: **issue:** `--intent-primary` is not, I believe, a Design System token; rather, I think it's an old token from Elements v4 (unless I'm blind and just can't see this in the DS). The DS tokens are outlined in https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=1800-11798&m=dev. Can you please ensure all CSS variables being used here are listed in the linked Figma file 🙏

src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
+import { AnchorHTMLAttributes, ButtonHTMLAttributes, FC, MouseEventHandler, ReactNode } from 'react'
+import { ElAnchorNavIconItem, ElButtonNavIconItem } from './styles'
+
+interface CommonNavIconItemProps {
```
kurtdoherty, line 4: **issue:** I think `aria-label` should be explicitly defined here as a required prop.

src/components/nav-icon-item/styles.ts
```diff
@@ -0,0 +1,61 @@
...
+  gap: var(--spacing-none, 0px);
+  border-radius: var(--corner-default, 0.25rem);
+  background: var(--white, #ffffff);
+  border: none;
+  outline: none;
+
+  /* change children color by default */
+  > * {
+    color: var(--neutral-400, #798da1) !important;
+  }
```
kurtdoherty, line 28: **question:** Do we strictly need this right now? Ideally our icon components would just inherit the colour of their parent. I'm just wary of use forcing styles on descendants like this...

Same comment applies to the active styles further down.

kurtdoherty (commented):
Thanks @ss-dimasm 👏 A few questions/comments to discuss. PR size was great btw.

src/components/nav-icon-item/nav-icon-item.tsx
kurtdoherty, line 1: **question:** I don't see any tooltip handling here, so just to confirm, is that because we're planning to implement a standalone `Tooltip` component via a separate ticket?

src/components/nav-icon-item/nav-icon-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+  if (!isButtonElement(props)) {
+    const { isActive, onClick, icon, ...rest } = props ?? {}
+
+    return (
+      <ElAnchorNavIconItem
+        {...rest}
+        role="button"
+        data-state={isActive ? 'active' : undefined}
+        onClick={(e) => {
+          e.preventDefault()
```
kurtdoherty, line 47: **issue:** I missed this in my original review, but we can't bake in `e.preventDefault()` here as that will prevent the link from being followed. It's up to the consumer to decide whether that's what they want or not.

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: > think aria-label should be explicitly defined here as a required prop

yeah, I think this could be good for future NavIconItem stuff, especially for the tooltip part beside of a11y

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: > is this to try and get a proper discriminated union happening?

yes, for discriminated union types. Unfortunately, neither `ButtonHTMLAttributes` from React nor `HTMLButtonElement` supports this

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: my bad, I misspelled it, it should be `rel` instead of `ref`

`rel` only valid on `<form>`,  `<link>`, `<area>`, and `<form>`; so I would exclude them for a `button`

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: > for readability, I personally find it best to keep simple if-else blocks like this one written so that the if-condition is not negated

placing the truthy condition based on the function name makes perfect sense

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: > Is role=button actually correct here? Shouldn't the <a> just have its normal implicit role?

after a bit research, I think it's best for us to let anchor nor button have their own implicit role

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: thanks for the suggestion, I think using `aria-current` is also a good option, as the styling is already separated in a styled component

with `aria-current`, it would simplifying the required custom attribute and improving accessibility at once

src/components/nav-icon-item/styles.ts
ss-dimasm: > Why do we need to have this flexibility for consumers? I'm struggling to see the value here

My concern is to keep the state for this component. For example, when a notification `NavIconItem` is clicked, it would open a new node (i.e., a dialog). Then, this data-state attribute could be utilized to make the `NavIconItem` active since the relevant component is in an open state

src/components/nav-icon-item/styles.ts
ss-dimasm: > Do we strictly need this right now? Ideally our icon components would just inherit the colour of their parent. I'm just wary of use forcing styles on descendants like this

I’ve thought about this previously; but when I try to add it to Storybook using `Icon` from Elements, the color won’t change since they have their own isolated color management. So, in the end, I just force-change the color, no matter what the children of `NavIconItem` have, what do you think?

src/components/nav-icon-item/styles.ts
ss-dimasm: > I think (not 100% sure) we would prefer :focus-visible, instead of :focus, as the focus outline really only needs to appear when focus is applied via keyboard interaction, not pointers

My reason for using `:focus` instead of `:focus-visible` is to support the most browser compatibility, especially for Safari users (it’s introduced in 15.4); but I think we could set a fallback for it.

src/components/nav-icon-item/styles.ts
ss-dimasm: > think it's an old token from Elements v4

right, I'll check it again to ensure it's using the latest design system

src/components/nav-icon-item/styles.ts
kurtdoherty: Ah, okay, so it is necessary at the moment. That's a shame.

**thought:** could we import the class name for the existing Icon component and target it here instead of `*`? If not, let's improve the comment you have so that we explain _why_ we overriding the colour of our descendants/children here (the current comment isn't very helpful because it's just describing what the CSS already communicates).

src/components/nav-icon-item/styles.ts
kurtdoherty: Ah, that's interesting. Do we know what devices that relates to? I only ask because `:focus-visible` is part of the Widely Available Baseline, which (personally) I think is reasonable for us to use as our minimum standard. I've not had to deal much with mobile browser compatibility in the past though, so more than happy to take the lead on this stuff from you and @adrian-reapit given it's directly applicable to the Reapit Go product 👍  I think `:focus` is just going to be a bit "in your face" for most desktop users 🤔

src/components/nav-icon-item/styles.ts
kurtdoherty: Yes, it makes sense that we need to put the "active" state in the DOM via an attribute, but I don't understand why we also need to do that for hover or focus?

src/components/nav-icon-item/nav-icon-item.tsx
kurtdoherty: Okay, and does this actually work the way you intend? I'm having a hard time visualising what the impact of not having this defined would be 😵  Maybe you could provide a [TS Playground](https://www.typescriptlang.org/play/) link demonstrating how defining these two as `never` helps with the discriminated union type?

src/components/nav-icon-item/styles.ts
ss-dimasm: I don't think we need that hover and focus states to be placed as part of `data-state` attribute 😶, just mark them with pseudo selector

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: > I don't see any tooltip handling here, so just to confirm, is that because we're planning to implement a standalone Tooltip component via a separate ticket?

that's the plan

**suggestion:** while this component depends on `Tooltip`, I think we could hold the ticket state as In Progress, since it's also requires the "Indicator" (red dot) that need to be clarified from Design Team

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: >  It's up to the consumer to decide whether that's what they want or not

right

src/components/nav-icon-item/styles.ts
ss-dimasm: > could we import the class name for the existing Icon component and target it here instead of *?

I'll try it first, and see how far it goes

src/components/nav-icon-item/nav-icon-item.tsx
ss-dimasm: > Maybe you could provide a [TS Playground](https://www.typescriptlang.org/play/) link demonstrating how defining these two as never helps with the discriminated union type?

I left a simple demonstration in [here](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAQgrsYB7AdgeRQYQDYEsDGA1lALxQAUAlKQHxQDkAzgBZJzYAmUAThMHNxRQABgCMEyFMKiNg3XCgDm9AFChIUAIIp8rbhhwFiZKrQYs2nHnwFDhAQx17ps+UtUqFwCNwBm9-Gh4RFQABW4kMEYoAG8VKChmXl8AfgAuKBQIADcfAG54qFRDIgzgyQM8IgKAXxVPFG8-AOhtXSRucMjouISkiFSM1wVFAoTiqsIMtr1Ko1r69WgASW8AW1JYCTCIqKgAHy0nDq6ogpV8VFkoXEZy1E3yXHWM1Yg1ygzn95vo+5RTtESHQAITfNYAOn6vnOlxQ10UfAAsnx7Bx7MB7I9wa91tRgbFCrhfBRbv8nnjqL0ElA4dcYkUsJMADSJZJQGqbcFjGl0pDYCAQ7BIRTkej-eiUQoJPkCoUi8jQyh5KAAelViSQuW4UGAzGg0NZzygAHdLFw1vZuMR7NFHFA4CgOAMFBAONLaVd+YLhaKJkYqMq1RrWNrdfqRP6iFRhEbgDJWOwuGA3PGAEQWJPWfiCETiEJSGRyEZpwqciDYRjQakyq7xhlRwis6EcrnrHm1+He+Wi+gzDqSj2yn0KpUq9WasN6g3JOOm81QS3WqC2ldCR3O3yurgdFdFtyKIdeuW+8iNwPjkNanzh6DCc+UaTGzNWFNeKAZxNWXg5uyOdrcC4xZKKWCR1DU5wThAAAe9hrGAAoOow9iIhcdZQPmkhvGsZTbACuxAoS4xMkYGSmASTBflwP62HmeFAQe9DMioEH1HS8b-no2HTMcnQEZs1LQhk9DMIgURpOqMFwQhgqXGsTGFI2ZH4nQlELjRuYOLxDEjAprEqIiwAopi6KYuQmGoNhyoGciqKmfY5CcR0Vl5EAA)

src/components/nav-icon-item/styles.ts
ss-dimasm: **note:** I think the wise way is to ensure the child component to have `color:inherit;` style

**thought:** the upcoming commit would import the `ElIcon` class name and have a comment to ensure the child of this component should have `inherit` style

ss-dimasm (commented):
Thanks for the constructive feedback, @kurtdoherty. I've addressed the resolvable points and kept the unresolved ones based on the top conversation

src/components/nav-icon-item/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+  align-items: center;
+  gap: var(--space-none, 0px);
+  border-radius: var(--corner-default, 4px);
+  background: var(--fill-white, #ffffff);
+  border: var(--border-none, 0px);
+  color: var(--icon-app_bar-default, #798da1);
+  outline: none;
+
+  &:focus-visible {
+    border-radius: var(--corner-default, 4px);
```
adrian-reapit, line 18: do we not have a global focus style?

ss-dimasm: > do we need a global focus style?

**thought:** it could be useful, but I'm unsure if it's wise, as nearly all AppBar items share the same global focus style

{B117EE35-67E4-4AD2-9FFD-2AF91D1AA1CA}

kurtdoherty: Yeah, we should really have a design token exported from Figma for the focus style's box shadow, but I believe all the design tokens require a refactor, so I'm not sure it's worth getting Design to specify one now or not 🤷‍♂️ Probably just need to ask them what they think...

src/components/nav-icon-item/nav-icon-item.tsx
kurtdoherty: I don't mind what we do from a ticket perspective. Can either do as you suggest, or just create two new tickets for the red dot and the tooltip integration. Either way, you're right that we need to track that we're only partially done.

src/components/nav-icon-item/nav-icon-item.tsx
kurtdoherty: Thanks for the playground; [here's an updated version](JYWwDg9gTgLgBDAnmApnA3nAggOwMYAW0AEgCoCyAMljDFMAEYCuMKAzgDRwBCLMEOMlRp1GLdlwBiAYS7kITNigCiANxQ4YxAIY4AJgBsUUOAF84AMygQQcAORQU2vDDsAoNwHpPPPgKHUtPTMrGxwACIA8soAynAAcpGkcMD4Bkx6aASOFnC6egjaUADmKDAAdHAAkvB6EOxwSii2) that I think shows the minimum we need (just `href`, not all the ones you have here)

kurtdoherty (approved):
Looking good, thanks @ss-dimasm 👏

Issue #192, opened 2024-11-04
v5 new components: BottomBarItem


ss-dimasm:
**Abstract**

To support the AppBar improvements, the `BottomBarItem` component should be made which can be used as part of `BottomBar` component

BottomBarItem design specification on Figma

![{670233AA-9863-411C-86DF-FE4311380F77}](https://github.com/user-attachments/assets/d6a96f27-230f-4b6b-bac6-8244e9eb1dba)

**Specification**

- Is new component or existing? New
- Is the component design signed off by product and design?  Signed off as part of the AppBar changes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4738&node-type=symbol&m=dev)

**Developer Checklist**

- [x] Styles alignment between Design System and Elements
- [x] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [x] Align with accessibility standards / spec as per above
- [x] If relevant, break down component into `Styles Only` and `React` component structures
- [x] Ensure all variants of components are documented as appropriate 
- [x] Ensure unit test coverage is adequate for component
- [x] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [x] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

ss-dimasm:
Hello @kurtdoherty and @adrian-reapit ,I 'm creating this ticket to focus on the [BottomBarIcon](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4737&node-type=frame&m=dev) component, based on the design. This component is distinct from the one in #164 and has its own complexities, so it's best to treat it as a separate component.

ss-dimasm:
I raised a comment to track the ticket's state, considering the current situation and our dependence on some answers from the Design Team

[ ] Initiate the scaffold of `BottomBarItem` following the design
[ ] Integrate with red dot indicator (awaiting the Design Team's answer [query link](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=16-4746&m=dev#1016315051))

[...]

PR #193, opened 2024-11-04
feat: #192 - initiate bottom bar item 

ss-dimasm:
# Pull request checklist

Part of [`BottomBar` navigation](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=34-6905&node-type=instance&m=dev)

**Context**
`BottomBarItem` is backing the build of the `BottomBar` component. It will be used to display the amount of items within the `BottomBar`

This PR initiates the basic implementation of `BottomBarItem` referring to the design, which only covers two states (default and active)

{7AE3FC42-9D00-4EDF-9E7B-2678F4ECFC87}

**Preview:**

https://github.com/user-attachments/assets/feae2855-17a8-4877-a999-a81b4e3a9b6e

**Design Specification Link:** https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4733&node-type=canvas&m=dev


src/components/bottom-bar-item/bottom-bar-item.tsx
kurtdoherty, line 1: note: I have some similar feedback for this `BottomBarItem` implementation as I did for the `NavIconItem` component in #191. Things like:
- `href` and `target` props being defined as `never` in the button element version;
- The use of `role="button"` on the anchor element instead of letting the anchor have its intrinsic role;
- The if-else block and avoiding negation for the if-condition, as that makes the else represent a double negation, which is harder to read/grok;
- The use of `data-state`; and,
- The use of `e.preventDefault()` in the click handler.

src/components/bottom-bar-item/bottom-bar-item.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    isActive: false,
+  },
+  render: (args) => {
+    return (
```
kurtdoherty, line 30: **question:** Can you help me understand your reasoning behind this component API? Specifically, why do you have `Icon` and `Label` as separate components to the main `BottomBarItem`?

src/components/bottom-bar-item/styles.ts
```diff
@@ -0,0 +1,82 @@
...
+interface ElAnchorBottomBarItemProps extends AnchorHTMLAttributes<HTMLAnchorElement>, CommonBottomBarItemProps {}
+export const ElAnchorBottomBarItemContainer = styled.a<ElAnchorBottomBarItemProps>`
+  ${baseStyles}
+
+  &:active,
+  &[data-state='active'] {
+    ${ElBottomBarItemIcon} {
+      * {
+        color: var(--icon-action, #4e56ea);
+      }
```
kurtdoherty, line 57: **question:** Same question I asked on #191 ... Do we strictly need this right now? Ideally our icon components would just inherit the colour of their parent. I'm just wary of use forcing styles on descendants like this...

kurtdoherty, line 57: **question:** Is there a reason why you use `*` here, but in #191 you used `> *`?

src/components/bottom-bar-item/styles.ts
```diff
@@ -0,0 +1,82 @@
...
+  flex-direction: column;
+  align-items: center;
+  gap: var(--spacing-half, 2px);
+  
+
+`
+
+interface ElAnchorBottomBarItemProps extends AnchorHTMLAttributes<HTMLAnchorElement>, CommonBottomBarItemProps {}
+export const ElAnchorBottomBarItemContainer = styled.a<ElAnchorBottomBarItemProps>`
+  ${baseStyles}
```
kurtdoherty, line 50: **question:** why are the "active" styles not part of the `baseStyles`?

kurtdoherty (commented):
Thanks @ss-dimasm 🙏 A few questions/comments to discuss. I'm particularly interested to understand why you opted for a very verbose API (handling the icon and label as part of `children` vs not).

Overall, I wonder if it's worth having the API for `BottomBarItem` and `NavIconItem` feel more related (at the moment, they feel quite different to me) 🤔

ss-dimasm:
Thanks @kurtdoherty , To avoid confusion, I think it's best to focus on #191 now that we've reached an agreement. I'll make the change and implement the same API from PR later

src/components/bottom-bar-item/styles.ts
ss-dimasm: >  why are the "active" styles not part of the baseStyles?

this is the interesting part. I attempted to incorporate the variable selector into the `baseStyles` using string interpolation from Linaria, but the expected node doesn't seem to have the correct CSS selector

as an example, assuming I've already declared a styled component named `ElBottomBarItemIcon` and need to modify its style based on the parent node's state, the first example works because it's plain text rendered as regular CSS. The generated Linaria class name, `.el-bottom-bar-item-icon`, directly affects the styles.

however, the second example doesn't work. Even after trying to escape the `$` symbol, the `styled` API from Linaria doesn't seem to process the style for that node. It appears that the node isn't being referenced correctly🫨

this works because it's directly translated to CSS
```ts
const baseStyles = `
  background-color: var(--fill-white, white);

  .el-bottom-bar-item-icon {
    color: var(--icon-secondary, #607890);
  }
`
```
here, the intent is to dynamically target the `ElBottomBarItemIcon` component, but the Linaria `styled` API doesn't seem to handle this correctly
```ts
const baseStyles = `
  background-color: var(--fill-white, white);

 ${ElBottomBarItemIcon} { // or \${ElBottomBarItemIcon}
    color: var(--icon-secondary, #607890);
  }
`
```

src/components/bottom-bar-item/styles.ts
ss-dimasm: > Is there a reason why you use * here, but in https://github.com/reapit/elements/pull/191 you used > *?

it's because the `ElIcon` have their own color management; and at that time I am not aware if the semantic color variable between text and icon is different; but with the latest commit it should not use that pattern

ss-dimasm: I assume we've found the answer, the isolated color management from `Icon` elements component

src/components/bottom-bar-item/bottom-bar-item.stories.tsx
ss-dimasm: > Can you help me understand your reasoning behind this component API?

at first, I thought the `BottomBarItem` would also handle the floating menu, as described [here](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=34-6931&node-type=instance&m=dev). However, I was wrong. This functionality should be handled neither by the `BottomBar` nor any upper parent component

src/components/bottom-bar-item/bottom-bar-item.tsx
ss-dimasm: totally agree, I've made the amendments in upcoming commit

ss-dimasm (commented):
Hi @kurtdoherty , this PR is ready for review. I left several comments on your conversation above for further discussion

ss-dimasm:
Following this, I’ve updated the PR description to ensure we’re aligned on excluding the red dot indicator status for `BottomBarItem`.

src/components/bottom-bar-item/bottom-bar-item.tsx
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * The JSX element to render as the icon
+   */
+  icon: ReactNode
+
+  /**
+   * Defines a string value that labels the current element
+   * While it's improve the accessibility, this will also used as a label part
+   */
+  'aria-label': string
```
kurtdoherty, line 19: **issue:** Since we have a visual label, I don't think it's necessary for us to have an `aria-label` as well. The accessible label will implicitly be the text context of the `BottomBarItem`.

**suggestion:** I think we should rename this prop to `label` (or use `children` for the label text) and make it a `ReactNode`. I like that both this prop and `icon` are required though 👍

src/components/bottom-bar-item/bottom-bar-item.tsx
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * To ensure it's aligned with the design, the children component should have a CSS color property with 'inherit' value
+ */
+export const BottomBarItem: FC<BottomBarItemProps> = (props) => {
+  if (isButtonElement(props)) {
+    const { isActive, icon, 'aria-label': ariaLabel, ...rest } = props ?? {}
+
+    return (
+      <ElButtonBottomBarItemContainer {...rest} aria-current={isActive ? 'true' : undefined}>
+        <ElBottomBarItemIcon>{icon}</ElBottomBarItemIcon>
```
kurtdoherty, line 59: **suggestion:** I'm not sure if the icon component we currently have already does this, but I wonder if we should do `<ElBottomBarItemIcon role="presentation">` here so that we explicitly communicate to the a11y tree that whatever content is inside that element should be ignored 🤔

kurtdoherty (approved):
Thanks @ss-dimasm 👏  Nice work here!

src/components/bottom-bar-item/styles.ts
kurtdoherty: **issue:** I think hardcoding the styled element's class name will present a lot of problems long-term:

1. It's not easily maintainable. If we change the name of the styled component, we may easily miss the need to also update the CSS where we have hardcoded its class name;
2. It's not obvious to a reader that the hardcoded class name is related to the styled component (certainly not to someone new to the repo/Linaria);
3. If we change the way Linaria formats the class names for styled components, these styles will break without us knowing.

Instead of needing to change the colour of the subcomponents, why don't we just change the colour of the root element and make sure the subcomponents inherit that colour?

i.e. 
```javascript
const baseStyles = `
  color: ...;

  [aria-current] {
    color: ...;
  }
`

const ElBottomBarItemIcon = styled.div`
  color: inherit;
`

const ElBottomBarItemLabel = styled.div`
  color: inherit;
`
```

ss-dimasm: yes, hardcoding is fragile in the long term. Your suggestion seems understandable; the subcomponents could inherit. I totally agree

src/components/bottom-bar-item/bottom-bar-item.tsx
ss-dimasm: > I think we should rename this prop to label (or use children for the label text) and make it a ReactNode

right `ReactNode` offers more flexibility, as it can accept any valid React component. Passing it as a children prop can improve code structure in certain instances

PR #194, opened 2024-11-05
feat: #163 new app-bar (dropdown button)

nurm717123:
continuing the sequence of pr for menu that have been discussed https://github.com/reapit/elements/issues/163, this is the 5th pr which is the 1st consumer of the menu i.e menu button that will be used in the app bar (top section).

as the app bar is not exist yet, this component will placed under `app-bar/components` folder with `NavMenuButton` name, the naming and the folder is just my idea, any better solution suggested will be well received.

# Pull request checklist
- [x] nav menu button
- [x] menu alignment (`left` | `right`)
- [x] auto anchor (vertical)

menu button


src/components/app-bar/components/nav-menu-button/index.tsx
```diff
@@ -0,0 +1,2 @@
+export * from './nav-menu-button'
```
nurm717123, line 1: should we use `app-bar-menu-button` instead though? since the term `nav` probably refer to older component 🤔

nurm717123:
image

after I go through the figma again the component there is named `NavDropdownButton`

kurtdoherty:
@nurm717123 can you please split this PR into two separate concerns 🙏 This ticket should be focused on the `NavDropdownButton`, not enhancing `Menu` with anchor positioning props; putting them together makes it a lot harder for me to review because those two pieces of work have very different contexts.

nurm717123:
> @nurm717123 can you please split this PR into two separate concerns 🙏 This ticket should be focused on the `NavDropdownButton`, not enhancing `Menu` with anchor positioning props; putting them together makes it a lot harder for me to review because those two pieces of work have very different contexts.

ok, no worries.
I will make another ticket for the anchoring logic 👍

kurtdoherty (approved):
LGTM 👍

src/components/nav-dropdown-button/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+  font-weight: 500;
+  line-height: var(--line-height-sm);
+  letter-spacing: var(--letter-spacing-sm);
+
+  ${ElIcon} {
+    pointer-events: none;
+    margin-left: 0;
+  }
+
+  &:focus {
```
kurtdoherty, line 26: **note:** we need to make sure we're consistent with how we apply focus styles. If we need to have a discussion as a team to decide between `:focus` and `:focus-visible`, let's have that discussion now via Teams/a GitHub discussion before we get too many components implemented in an inconsistent manner.

nurm717123: or should we ask a design team, whether they want the `focus` style to be appear when user click the button 🤔

nurm717123: seems like we can rever to original button component which user `:focus-visible` 

image

Issue #195, opened 2024-11-06
v5 new component: AppBar/SearchButton 


nurm717123:
**Abstract**

_As part of the v5 Elements release, the ticket #153 was created. To break it into smaller part and allow for more focused discussions, this ticket has been created._

***app bar (parent)***
![image](https://github.com/user-attachments/assets/dbc50742-49e5-4c0d-aebf-2305aa72bfc8)

***this ticket***
![image](https://github.com/user-attachments/assets/0752163a-f497-4986-ad5c-deccec7cedc9)

**Specification**

- Is new component or existing? new
- Is the component being re-named? no
- Is the component design signed off by product and design? yes, as part of the new app bar
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4738&node-type=symbol&m=dev)
- [Link to Elements Audit in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4792&m=dev)
- [Accessibility requirements](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)
- [Parent ticket](https://github.com/reapit/elements/issues/153)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

nurm717123:
Hi @kurtdoherty and @adrian-reapit , I created this ticket so the discussion could be more focused and easier, and since the app-bar structure haven't build yet, I think we can discuss on how it will look on this kind of small pr. let me know if this is too small and I will update it 👍

nurm717123:
my Idea of how it will be structured is something like this:

- app-bar
  - components/ (atomic component)
    - search button
    - menu button
[...]

PR #196, opened 2024-11-06
feat: #195 new app-bar (search button)

nurm717123:
this is small part of the app-bar component, which will be used in the top bar navigation

***app bar (parent)***
image

***this ticket***
image

note: _the responsive handling e.g: icon button will display at smaller screen_ will be handled by the parent (consumer)

search button


src/components/app-bar/components/search-button/search-button.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import { Meta, StoryObj } from '@storybook/react'
+import { NavSearchButton } from './search-button'
+
+export default {
+  title: 'Components/App bar/Search button',
+  component: NavSearchButton,
+} as Meta<typeof NavSearchButton>
+
+export const StylesOnlyUsage: StoryObj<typeof NavSearchButton> = {}
+export const ReactUsage: StoryObj<typeof NavSearchButton> = {}
```
nurm717123, line 10: this component is really simple, but since it doesn't have a bigger consumer to give an example I just created this stories

src/components/app-bar/icons/search-icon.svg
nurm717123, line 1: just my initial idea of where it should placed, as I proposed in this ticket https://github.com/reapit/elements/issues/195#issuecomment-2458909326

src/components/app-bar/icons/search-icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+  <path d="M14.4733 13.5266L10.8733 9.93331C11.6055 9.00212 12.0024 7.85126 11.9999 6.66665C11.9999 3.72113 9.6121 1.33331 6.66659 1.33331C3.72107 1.33331 1.33325 3.72113 1.33325 6.66665C1.33325 9.61217 3.72107 12 6.66659 12C7.8512 12.0025 9.00206 11.6056 9.93325 10.8733L13.5266 14.4733C13.6518 14.5995 13.8222 14.6705 13.9999 14.6705C14.1777 14.6705 14.3481 14.5995 14.4733 14.4733C14.5995 14.3481 14.6705 14.1777 14.6705 14C14.6705 13.8222 14.5995 13.6518 14.4733 13.5266ZM2.66659 6.66665C2.66659 4.45751 4.45745 2.66665 6.66659 2.66665C8.87572 2.66665 10.6666 4.45751 10.6666 6.66665C10.6666 8.87579 8.87572 10.6666 6.66659 10.6666C4.45745 10.6666 2.66659 8.87579 2.66659 6.66665Z" fill="#9FAEBC"/>
```
kurtdoherty, line 2: **suggestion:** Not a big deal in this instance, but I think its usually good to change this to be `fill="currentColor"` so that the SVG inherits its parent's colour.

src/components/app-bar/app-bar.atoms.tsx
```diff
@@ -0,0 +1,19 @@
...
+} from './styles'
+
+export interface NavSearchButton extends HTMLAttributes<HTMLButtonElement> {}
+
+export const NavSearchButton: React.FC<NavSearchButton> = (props) => {
+  return (
+    <ElNavSearchButton {...props}>
+      <ElNavSearchButtonIcon icon="search" />
+      <ElNavSearchButtonPlaceholder>Search</ElNavSearchButtonPlaceholder>
+      <ElNavSearchButtonShortcutText>Ctrl + K</ElNavSearchButtonShortcutText>
```
kurtdoherty, line 16: **issue:** I don't think the shortcut should not be hardcoded here. Not all products will necessarily implement the keyboard shortcut, and even if they do, some may not be able to use Ctrl + K specifically.

src/components/app-bar/app-bar.atoms.tsx
```diff
@@ -0,0 +1,19 @@
...
+  ElNavSearchButtonPlaceholder,
+  ElNavSearchButtonShortcutText,
+} from './styles'
+
+export interface NavSearchButton extends HTMLAttributes<HTMLButtonElement> {}
+
+export const NavSearchButton: React.FC<NavSearchButton> = (props) => {
+  return (
+    <ElNavSearchButton {...props}>
+      <ElNavSearchButtonIcon icon="search" />
```
kurtdoherty, line 14: **question:** What is the `icon="search"` prop is doing here?

kurtdoherty, line 14: **suggestion:** I think we should be passing `role="presentation"` to the icon.

src/components/app-bar/app-bar.atoms.tsx
kurtdoherty, line 1: **question:** what's the reasoning behind treating NavSearchButton as an "atom" of the App Bar? i.e. why put it inside `/app-bar` and not `/nav-search-button`?

src/components/app-bar/search-button.mdx
kurtdoherty, line 1: **todo:** this file and the stories file should be named the same as the component they're related to (i.e. `nav-search-button.mdx` and `nav-search-button.stories.tsx`

kurtdoherty (commented):
Thanks @nurm717123 👏 Small amount of feedback. Also, not sure if there's been a miscommunication on the ticket for this work, but I was surprised to see this component implemented within an `/app-bar` folder 🤔

nurm717123:
> @nurm717123 👏 Small amount of feedback. Also, not sure if there's been a miscommunication on the ticket for this work, but I was surprised to see this component implemented within an `/app-bar` folder 🤔

Appreciate it, thanks for the feedback. for the why I put it under /app-bar, I think I've missunderstood it sorry, I thought because it is small component and it will only used in the `app-bar` it would different than what we've discussed. but now I think its better to separate it just like `app-switcher` 👍

src/components/app-bar/app-bar.atoms.tsx
nurm717123: I think I've misunderstood it sorry, I thought because it is small component and it will only used in the app-bar it would different than what we've discussed. but now I think its better to separate it just like app-switcher

src/components/app-bar/app-bar.atoms.tsx
nurm717123: sorry, I forgot to delete it after I tried the existing one

src/components/app-bar/app-bar.atoms.tsx
nurm717123: hmm I haven't seen that in the design, because every example in the design have this shortcut so I thought its really required 🤔 , but I think it make sense to make it optional to be save

src/components/app-bar/app-bar.atoms.tsx
nurm717123: yeah, or I think `aria-hidden="true"` is more appropriate since it already has placeholder

kurtdoherty: Ah, of course, yes I'd forgotten about aria-hidden 🤦‍♂️ 👍

src/components/nav-search-button/icons/search-icon.svg
```diff
@@ -1,3 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="currentColor">
   <path d="M14.4733 13.5266L10.8733 9.93331C11.6055 9.00212 12.0024 7.85126 11.9999 6.66665C11.9999 3.72113 9.6121 1.33331 6.66659 1.33331C3.72107 1.33331 1.33325 3.72113 1.33325 6.66665C1.33325 9.61217 3.72107 12 6.66659 12C7.8512 12.0025 9.00206 11.6056 9.93325 10.8733L13.5266 14.4733C13.6518 14.5995 13.8222 14.6705 13.9999 14.6705C14.1777 14.6705 14.3481 14.5995 14.4733 14.4733C14.5995 14.3481 14.6705 14.1777 14.6705 14C14.6705 13.8222 14.5995 13.6518 14.4733 13.5266ZM2.66659 6.66665C2.66659 4.45751 4.45745 2.66665 6.66659 2.66665C8.87572 2.66665 10.6666 4.45751 10.6666 6.66665C10.6666 8.87579 8.87572 10.6666 6.66659 10.6666C4.45745 10.6666 2.66659 8.87579 2.66659 6.66665Z" fill="#9FAEBC"/>
```
kurtdoherty, line 2: **issue (non-blocking):** I think you've updated the wrong `fill`. It's the `fill` on the `<path>` element that has a hardcoded hex value. We would typically want that to be `currentColor` so that we can set the appropriate colour via a parent element using a design token.

src/components/nav-search-button/nav-search-button.tsx
```diff
@@ -6,14 +6,16 @@ import {
   ElNavSearchButtonShortcutText,
 } from './styles'
 
-export interface NavSearchButton extends HTMLAttributes<HTMLButtonElement> {}
+export interface NavSearchButton extends HTMLAttributes<HTMLButtonElement> {
+  useShortcut?: boolean
```
kurtdoherty, line 10: **issue (non-blocking):** `useShortcut` sounds like it will actually cause the shortcut to function... I think `isShortcutVisible` or `showHotKey` would be better (personally, I lean towards the terminology of "hot key" over "shortcut", but it's not a big deal).

kurtdoherty (approved):
Thanks @nurm717123. Two more minor issues; neither of them are blocking.

src/components/nav-search-button/icons/search-icon.svg
nurm717123: oh yes, I overseen it thanks for noticing that 👍

PR #200, opened 2024-11-08
feat: #195 new app-bar (nav item)

nurm717123:

***app bar (parent)***
image

***this ticket***
image

#### preview storybook example

nav item

### DIFF
--- a/src/components/nav-item/nav-item.tsx
+++ b/src/components/nav-item/nav-item.tsx
@@ -24,7 +24,7 @@ interface NavItemAsAnchorElementProps extends CommonNavItemProps, AnchorHTMLAttr
 export type NavItemProps = NavItemAsButtonElementProps | NavItemAsAnchorElementProps
 
 export const NavItem: FC<NavItemProps> = ({ children, disabled, isActive, ...props }) => {
-  if (isItemAsButtonElement(props)) {
+  if (isButtonElement(props)) {
     return (
       <ElNavItemButton {...props} aria-current={isActive ? 'true' : undefined}>
         <ElNavItemLabelContainer aria-disabled={disabled}>{children}</ElNavItemLabelContainer>
@@ -39,6 +39,6 @@ export const NavItem: FC<NavItemProps> = ({ children, disabled, isActive, ...pro
   )
 }
 
-function isItemAsButtonElement(props: NavItemProps): props is NavItemAsButtonElementProps {
+function isButtonElement(props: NavItemProps): props is NavItemAsButtonElementProps {
   return !props.href
 }
--- a/src/components/nav-item/nav-item.tsx
+++ b/src/components/nav-item/nav-item.tsx
@@ -1,0 +1,44 @@
+import type { AnchorHTMLAttributes, ButtonHTMLAttributes, FC } from 'react'
+import { ElNavItemAnchor, ElNavItemButton, ElNavItemLabelContainer } from './styles'
+
+interface CommonNavItemProps {
+  /**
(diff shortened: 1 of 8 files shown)


src/components/nav/__styles__/index.ts
```diff
@@ -102,7 +102,8 @@ export const ElNavItem = styled.a`
     margin-right: auto;
     width: 100%;
     padding: 0rem 1.25rem;
-    border-bottom: 1px solid var(--neutral-100);
+    /* border-bottom: 1px solid var(--neutral-100); */
+    border-bottom: 1px solid red;
```
kurtdoherty, line 106: **question:** What's this change all about?

src/components/nav-item/styles.ts
```diff
@@ -0,0 +1,51 @@
...
+  align-items: flex-start;
+
+  font-family: var(--font-family, Inter);
+  font-size: var(--font-size-sm, 14px);
+  font-style: normal;
+  font-weight: 500;
+  line-height: var(--line-height-sm, 20px);
+  letter-spacing: var(--letter-spacing-sm, -0.14px);
+`
+
```
kurtdoherty, line 15: **thought:** I didn't realise this with the `NavDropdownButton` work, but that component is really just a `NavItem` with an icon and a menu, so I wonder if we should have `NavDropdownButton` use `NavItem` (or `ElNavItemButton`) under the hood 🤔 

Not something we need to do right now, just flagging that I think we missed this relationship when planning out the sequence of the work (i.e. NavItem should have come first, then NavDropdownButton).

src/components/nav-item/nav-item.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return (
+    <ElNavItemAnchor {...(props as NavItemAsAnchorElementProps)} aria-current={isActive ? 'true' : undefined}>
+      <ElNavItemLabelContainer>{children}</ElNavItemLabelContainer>
+    </ElNavItemAnchor>
+  )
+}
+
+function isItemAsButtonElement(props: NavItemProps): props is NavItemAsButtonElementProps {
```
kurtdoherty, line 42: **quibble:** I find this name a bit awkward. I think we've used `isAnchorElement` or `isButtonElement` in other recent components.

kurtdoherty (approved):
Looks good. No blockers, just a few comments.

src/components/nav-item/styles.ts
nurm717123: ah yes, both of them have different variant (i.e dropdown-button doesn't have `selected` variant) I think that is the reason, but actually it could be combined.

nurm717123: I can make another pr for that later

src/components/nav-item/nav-item.tsx
nurm717123: yes, I think that is simpler 👍

src/components/nav/__styles__/index.ts
nurm717123: Oops, during development I forgot to restore it 😭 , thanks for noticing that 👍

src/components/nav-item/styles.ts
kurtdoherty: I believe the dropdown button would show as active when it's open and/or when one of it's items is active; but yes, we'd need to confirm behaviour with design before doing that work. I don't think it's costly to have this duplication though, so I think DRYing it up would be pretty low priority.

PR #202, opened 2024-11-11
feat: #163 new menu (auto anchor and custom y-position)

nurm717123:
# Context

To enhance the usability of the Menu component for React consumers, an automatic y-offset anchoring functionality needs to be implemented. The x-position will be adjustable through the data-alignment prop of the Menu component.

Additionally, because the Menu in the app-bar appears slightly lower than the original Menu, so a prop is required to control this additional gap.

# Pull request checklist
- [x] auto y-offset anchoring
- [x] manual x-offset anchoring by provide `data-alignment` props to the `Menu`
- [x] provide `additionalGap` props for `Menu.Popover` to adjust the additional y-offset.

_note: html user still can use `data-alignment`, but not for auto anchoring_

fixes: 
 - https://github.com/reapit/elements/issues/163

menu auto anchor


src/components/nav-dropdown-button/nav-dropdown-button.stories.tsx
```diff
@@ -50,7 +50,7 @@ export const ReactUsage: Story = {
   render: (props) => {
     return (
       <FlexContainer isFlexAlignCenter isFlexJustifyCenter>
-        <Menu {...props}>
```
nurm717123, line 53: remove unused props

src/components/menu/__tests__/__snapshots__/menu-popover.test.tsx.snap
```diff
@@ -12,6 +12,7 @@ exports[`Menu Popover component should render and close Popover based on any ele
     </button>
     <mock-styled.div
       data-open="true"
+      style="top: 3px;"
```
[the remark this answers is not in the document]
kurtdoherty: **question:** I can't see where you've gotten this 3px value from in the Figma files?

src/components/menu/menu-popover.tsx
```diff
@@ -1,29 +1,70 @@
...
+  }
+}
+
+export const MenuPopover: FC<{
+  /**
+   * Optional parameter to adjust the vertical position of the popover
+   *
+   * @default 0
+   */
+  additionalGap?: number
```
kurtdoherty, line 32: **suggestion:** I think it's nicer to accept a `string` as that allows consumers to specify any valid CSS length value.

Also, I'd suggest a prop name like `yOffset` where the default is the 3px of whatever is specified by the Design System. This way, engineers only need to match the total offset shown in a design instead of needing to manually subtract the default offset.

src/components/menu/menu-popover.tsx
```diff
@@ -1,29 +1,70 @@
...
+    // close menu if clicked target has parent with below selector
+    const menuItem = event.target.closest('[data-close-menu="true"]')
+    if (menuItem) closeMenu()
   }
 
-  return null
+  return (
+    <ElMenuPopover
+      style={popoverStyle}
+      ref={popoverRef as MutableRefObject<HTMLDivElement>}
```
kurtdoherty, line 61: **question:** I don't understand why you're needing to use a type assertion for the ref here?

src/components/menu/menu-popover.tsx
```diff
@@ -1,29 +1,70 @@
...
 
-  if (isOpen) {
-    const onClick = (event) => {
-      // close menu if clicked target has parent with below selector
-      const menuItem = event.target.closest('[data-close-menu="true"]')
-      if (menuItem) closeMenu()
+  useLayoutEffect(() => {
+    const container = popoverRef.current?.parentElement
+    if (container && isOpen && popoverRef.current) {
+      calculatePopoverPosition(container, popoverRef.current, setPopoverStyle, additionalGap)
```
kurtdoherty, line 46: **question:** Does this work when scrolling the viewport? i.e. if a menu is open and the viewport is scrolled, will this behaviour cause the menu to change position?

**note:** I'm not suggesting we _need_ this behaviour, just wanting to understand what has/has not be explored with the current implementation.

kurtdoherty (commented):
Thanks @nurm717123 👏 Mostly a bunch of questions for clarity.

src/components/menu/menu-popover.tsx
nurm717123: yes, I think I need to update the example, thanks for noticing this

menu auto anchor scroll

src/components/menu/menu-popover.tsx
nurm717123: something wrong with my linter but its good now, turns out it doesn't need to have that assertion👍

src/components/menu/menu-popover.tsx
kurtdoherty: @nurm717123 Sorry, can you scroll without clicking the mouse? Clicking the mouse obviously causes the menu to close, so then reopening would naturally result in the correct positioning. I'm interested in seeing whether the menu's position will update automatically when it's open and the viewport is scrolled. I suspect it won't reposition, which I think is probably fine for now, so long as we make that clear to consumers via the Storybook docs.

In future, I think we'd want to switch to using the new [CSS anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning) features; they just don't have enough support right now to make it worthwhile.

src/components/menu/__tests__/__snapshots__/menu-popover.test.tsx.snap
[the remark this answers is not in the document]
nurm717123: actually I just manually calculating it, as you can see in below screenshot without additional `+3` it look like this
image
image

with `+3`


note: there is no example for the `Menu` usage with `button` in the design as far I've seen (except nav dropdown button which would need more gap) 🤔

src/components/menu/menu-popover.tsx
nurm717123: yes, it need to be reopened

> In future, I think we'd want to switch to using the new [CSS anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning) features; they just don't have enough support right now to make it worthwhile.

I will look into this 👍 

menu auto anchor scroll

src/components/menu/__tests__/__snapshots__/menu-popover.test.tsx.snap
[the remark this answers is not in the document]
kurtdoherty: Okay, I thought `3px` was a strange number. Can you please ask design what the default offset should be? 🙏 It's not our decision to make, unless they explicitly tell us we can make the decision.

For reference, `3px` seems strange to me because it's not one of the spacing design tokens. If design were to specify this offset, I would expect them to use one of their `--space-*` tokens, likely `--space-half` (2px) or `--space-1` (4px), but again, this kind of detail is for Design to specify, not us.

src/components/menu/menu-popover.tsx
kurtdoherty: @nurm717123 sorry, that gif looks like you're still clicking and dragging the scroll bar. I'd like to see how the menu behaves when you scroll without any click and drag (i.e. scrolling via a touchpad or mouse wheel).

nurm717123: I've put the wrong gif previously, I've updated it with newest, but essentially it shows that currenltly it need to be reopened as you said

src/components/menu/__tests__/__snapshots__/menu-popover.test.tsx.snap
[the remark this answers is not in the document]
nurm717123: yes, I guess since the current example only for app-bar (and we will use it for app-bar initially) can we adjust the gap to match the app bar example? , there is no offset prop in the spec but we can calculate it using same app-bar height value which is `size-10` or default to `0` instead, so just need to custom y offset with props?

image

src/components/menu/menu-popover.tsx
kurtdoherty: haha no worries. Thanks for confirming 🙏

src/components/menu/menu-popover.tsx
nurm717123: > suggestion: I think it's nicer to accept a string as that allows consumers to specify any valid CSS length value.

I'm not certain about this, as I calculate the available space using DOM methods like getBoundingClientRect(), which returns values in pixels. I expect this prop to be used rarely, as the default y-offset should generally be correct. However, if the consumer requires additional space, this prop can be utilized

and by the way, since the app-bar and bottom-bar has `position: fixed` container, I think I don't have to support this since it(auto anchor) can be overriden by e.g
```css
 .el-menu-popover {
    top: var(--nav-height) !important;
  }
```

> Also, I'd suggest a prop name like yOffset where the default is the 3px of whatever is specified by the Design System. This way, engineers only need to match the total offset shown in a design instead of needing to manually subtract the default offset.

yeah, I think its more make sense, I will note this 👍

nurm717123:
Thanks @kurtdoherty , In response to your feedback, I have made the following changes: 
- [x] remove default `gap` (as currently no example for the menu usage with trigger, except in appbar) and also app bar doesn't require dynamic positioning
- [x] update the docs
- [x] remove unused type assertion

note: I have tried to make the calculatePopoverPosition function run every time the user scrolls, but it requires the component to know which container to listen to. While I can default it to add the listener to the body, there are cases where the body is not scrollable, but the children are. Therefore, more discussion is needed on this matter.

let me know if there is something missing, thanks.

src/components/menu/menu-popover.tsx
```diff
@@ -7,19 +7,18 @@ export const calculatePopoverPosition = (
   container: HTMLElement,
   popover: HTMLDivElement,
   setPopoverStyle: React.Dispatch<React.SetStateAction<React.CSSProperties>>,
-  additionalGap = 0,
+  yOffset = 0,
```
kurtdoherty, line 10: **suggestion:** Andrei confirmed in Figma that the default can be `--space-1` which is 4px. Be good to get that locked in now. I'd also suggest moving the default value to the component, not this helper.

kurtdoherty (approved):
Just the one suggestion re default offset; otherwise look good 👍

kurtdoherty:
> Thanks @kurtdoherty , In response to your feedback, I have made the following changes:
> 
> * [x]  remove default `gap` (as currently no example for the menu usage with trigger, except in appbar) and also app bar doesn't require dynamic positioning
> * [x]  update the docs
> * [x]  remove unused type assertion
> 
> note: I have tried to make the calculatePopoverPosition function run every time the user scrolls, but it requires the component to know which container to listen to. While I can default it to add the listener to the body, there are cases where the body is not scrollable, but the children are. Therefore, more discussion is needed on this matter.
> 
> let me know if there is something missing, thanks.

Thanks for the updates @nurm717123 🙏 

Now that Andrei has confirmed what the default offset should be for the menu, I think it's worth getting in now. Up to you though.

We definitely don't want `calculatePopoverPosition` running on scroll. Not worth the investment at this stage. As I've mentioned previously, I'd rather leverage the new CSS anchor positioning features once they have wider browser support.

src/components/menu/menu-popover.tsx
nurm717123: Ok 👌

nurm717123: but I think its not applicable for this case to put the default spacing in css, because it will always be overriden by the `calculatePopoverPosition`, even if that function disabled, `top: -4px` will be covering the button. I think I will add this to the function and add a comment for it

kurtdoherty: Yeah, I understand `calculatePopoverPosition` will override any initial offset, my suggestion for moving the 4px default to the component is one of ownership. In my view, it is the component that owns the default, not `calculatePopoverPosition`.

Not sure where you're getting `-4px` from though. Maybe I'm missing something here, but the offset is an absolute value, whether it's positive or negative is up to `calculatePopoverPosition` to determine based on where it decides to place the popover, right?

nurm717123: > Not sure where you're getting -4px from though.
apologize, that should be `4px` (var(--spacing-1) or 4px or 0.25rem)

> Maybe I'm missing something here, but the offset is an absolute value, whether it's positive or negative is up to calculatePopoverPosition to determine based on where it decides to place the popover, right?

yes, its absolute value of the parent (Menu) so if its default 4px it will be covering the trigger (since the trigger is tipically render before menu). the calculate fn is also responsible to position the mene above or below the trigger

nurm717123: Ok, I think I got it now, I think I can move the default to the component props instead of inside helpers, last week I thought it should be on the css which is not possible to combine it with the helper 😆 .

PR #204, opened 2024-11-12
fix: Menu onBlur trigger before trigger click handler

nurm717123:
When a menu item is clicked and configured to keep the menu open, it receives focus. An onBlur handler is set to close the menu when focus moves outside the Menu, causing an issue when focus shifts to the trigger button. The fix is to prevent the onBlur handler from executing if the focus is moving to the trigger button. 

onblur issue
# Pull request checklist

fixes: 
- part of the #163 

fixed result:
onblur issue fix

### DIFF
--- a/src/hooks/use-menu/__tests__/index.test.tsx
+++ b/src/hooks/use-menu/__tests__/index.test.tsx
@@ -217,6 +217,23 @@ describe('useMenu', () => {
         } as any)
         expect(result.current.isOpen).toBe(false)
       })
+
+      it('should not close menu if focus moves outside by clicking trigger button', () => {
+        const { result } = renderHook(() => useMenu())
+        result.current.openMenu()
+        expect(result.current.isOpen).toBe(true)
+
+        result.current.getPopoverProps().onBlur!({
+          currentTarget: {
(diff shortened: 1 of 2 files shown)


src/hooks/use-menu/__tests__/index.test.tsx
```diff
@@ -217,6 +217,23 @@ describe('useMenu', () => {
         } as any)
         expect(result.current.isOpen).toBe(false)
       })
+
+      it('should not close menu if focus moves outside by clicking trigger button', () => {
```
kurtdoherty, line 221: **todo:** can you tweak this test description to say "don't close the menu on blur if focus is moving to the trigger button." It's more about the focus moving to the trigger than it is about the trigger specifically being _clicked_ (i.e focus may move to the tigger via the keyboard, not clicking).

kurtdoherty (approved):
Nice find 👌 Just the one minor change.

src/hooks/use-menu/__tests__/index.test.tsx
nurm717123: ok, that's more accurate 👌

Issue #137, opened 2024-07-30
v5 component refactor: Avatar 


HollyJoyPhillips:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing - called 'Avatar' in Elements. However, should be reviewed/adhere to variants in the DS e.g. circle, rectangle & square 
- Is the component being re-named? No but should review variants and ensure they are aligned
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-7281&t=8GcgX59FmafMRAda-4)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-25698&t=T27RmqyqogSxbWuo-4)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/avatar--docs)
- [Accessibility requirements](https://www.w3.org/WAI/tutorials/images/informative/)

**Developer Checklist**

- [x] Styles alignment between Design System and Elements
- [x] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [x] Align with accessibility standards / spec as per above
- [x] If relevant, break down component into `Styles Only` and `React` component structures
- [x] Ensure all variants of components are documented as appropriate 
- [x] Ensure unit test coverage is adequate for component
- [x] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [x] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

ksolanki7:
After discussing with the Design team, I am moving this ticket back to the "To Do" column, as there are still clarifications needed and updates required for the component. The ticket will not be ready for development until these items are resolved by the Design team. Please refrain from picking up this ticket until it has been cleared by the Design team.

ksolanki7:
[...]

PR #205, opened 2024-11-14
chore: #137 Deprecate previous avatar in preparation for new avatar

nurm717123:
## Context
In preparation for the new v5 avatar release, there will be new Avatar components (`Avatar` and `AvatarRectangle`). Although the previous component was called 'avatar,' the next version will likely use `AvatarRectangle`, depending on whether an image needs to be displayed.

# Pull request checklist
- [x] deprecate previous avatar and consumer

fixes: 
- #137

### DIFF
--- a/src/components/avatar/__styles__/index.ts
+++ b/src/components/avatar/__styles__/index.ts
@@ -1,9 +1,11 @@
 import { css } from '@linaria/core'
 import { styled } from '@linaria/react'
 
+/** @deprecated */
 export const elAvatarSmall = css``
 
-export const ElAvatar = styled.div`
+/** @deprecated will be replaced by new v5 ElAvatarRectangle */
+export const ElDeprecatedAvatar = styled.div`
   display: flex;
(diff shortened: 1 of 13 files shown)


src/components/avatar/__styles__/index.ts
kurtdoherty, line 1: **suggestion:** I would also rename the `avatar` folder to `deprecated-avatar` or similar so that the new work can deal exclusively with new files (will make the PR diff simpler and avoid mixing deprecated code with new code together in the same files).

kurtdoherty (approved):
Thanks @nurm717123 🙏 Just the one suggestion re the folder name. I don't need to see the PR again if you action that, unless you need my input on anything else that "falls out" of that change.

src/components/avatar/__styles__/index.ts
nurm717123: ok, I totally agree with that, I was thinking the same thing 👍

Issue #206, opened 2024-11-14
v5 new component: MobileNavItem 


ss-dimasm:
**Abstract**

To create the [ `MobileNavMenu`](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=18-6292&node-type=instance&m=dev) component (a part of the AppBar), we need to define a scope that focuses on the smaller, individual [`MobileNavItem`](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4988&node-type=frame&m=dev) components that will be contained within it

![{62A59BF2-B739-4EDD-AF96-F8CDE9928C19}](https://github.com/user-attachments/assets/8d25a271-1e45-470d-8b14-e08a393a8f96)

This component is responsible for:
1. Displaying the single basic item variant
2. Displaying the collapsed expandable item variant
3. Displaying the expanded expandable item variant

**Specification**

- Is new component or existing? new
- Is the component design signed off by product and design? Signed off as part of the AppBar changes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4988&node-type=frame&m=dev)
- [Accessibility requirements](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation-hybrid/)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

ss-dimasm:
[...]

PR #209, opened 2024-11-18
feat: #206 initiate mobile navigation item

ss-dimasm:
# Pull request checklist
#206 

## Context
To accommodate the Mobile Navigation Menu, we need to focus on the smaller part first, which would be the `Mobile Navigation Item`

This PR ships the base structure for how the Mobile Navigation Item would look and also creates the basic ARIA accessibility attributes that have been declared on the Disclosure Pattern ([Disclosre Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation))

Talking about the pattern, this PR does not implement keyboard interaction at this moment.  Therefore, I believe it's necessary to have further discussion once the base component is merged or at the phase of building `MobileNavMenu`

Simple Variant Interface
```tsx
<MobileNavItem label="Item 1" href="#item-1" hasBadge />
<MobileNavItem label="Item 1" onClick={noop} hasBadge />
```

Expandable Variant Interface
```tsx
<MobileNavItem label="Item 2" hasBadge isActive>
  <MobileNavItem label="Sub Item 2.1" href="#item-2.1" isActive />
  <MobileNavItem label="Sub Item 2.2" onClick={noop} hasBadge />
  <MobileNavItem label="Sub Item 2.3" href="#item-2.3" />
</MobileNavItem>
```

{D2B55AF1-7789-43ED-97FB-78E791BD8BF9}

https://github.com/user-attachments/assets/daf3cf27-8a78-498c-91cd-5e83c592081c

fixes: #


src/components/mobile-nav-item/mobile-nav-item.tsx
```diff
@@ -0,0 +1,82 @@
...
+              {hasBadge && <ElMobileNavItemBadge />}
+            </ElMobileNavItemContent>
+          </ElMobileNavItemAnchor>
+        </ElMobileNavItemListItem>
+      </ElMobileNavItemUnorderedList>
+    )
+  }
+}
+
+MobileNavItem.SubItem = MobileNavSubItem
```
ss-dimasm, line 80: **thought:** the SubItem naming within a MobileNavItem seems odd

src/components/mobile-nav-item/mobile-nav-item.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const MobileNavItem: MobileNavItemFC = (props) => {
+  const [isExpanded, setIsExpanded] = useState(Boolean(props?.isDefaultExpanded))
+  const panelId = useId()
+
+  if (isExpandableVariant(props)) {
+    const { isActive, label, hasBadge, children } = props ?? {}
+
+    return (
+      <ElMobileNavItemUnorderedList data-is-expanded={isExpanded} role="menubar" aria-label={label}>
```
kurtdoherty, line 38: **issue:** I'm not sure `menubar` is the correct role to use here. If we're trying to follow the [Disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/), it doesn't use this role at all...

src/components/mobile-nav-item/mobile-nav-item.tsx
```diff
@@ -0,0 +1,82 @@
...
+const isExpandableVariant = (props: MobileNavItemProps): props is MobileNavItemExpandableProps => {
+  return 'children' in props
+}
+
+type MobileNavItemFC = FC<MobileNavItemProps> & {
+  SubItem: FC<MobileNavSubItemProps>
+}
+
+const MobileNavItem: MobileNavItemFC = (props) => {
+  const [isExpanded, setIsExpanded] = useState(Boolean(props?.isDefaultExpanded))
```
kurtdoherty, line 31: **question:** do we need a "default expanded" prop? Also, if `isActive` is true, would we not want the nav item to be expanded?

src/components/mobile-nav-item/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  children?: never
+  isDefaultExpanded?: never
+  href: string
+}
+
+export interface MobileNavItemExpandableProps extends CommonMobileNavItemProps {
+  children: ReactNode
+  isDefaultExpanded?: boolean
+  href?: never
+}
```
kurtdoherty, line 19: **question:** Do we really need this separation between "simple" and "expandable" props for the component? Also, are we absolutely sure the parent item can't have a URL associated with it?

src/components/mobile-nav-item/mobile-nav-item.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const panelId = useId()
+
+  if (isExpandableVariant(props)) {
```
kurtdoherty, line 34: **suggestion:** I think there's enough going on here to warrant splitting out each branch of this if-else statement to separate subcomponents.

src/components/mobile-nav-item/mobile-nav-item.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const { isActive, label, hasBadge, children } = props ?? {}
+
+    return (
+      <ElMobileNavItemUnorderedList data-is-expanded={isExpanded} role="menubar" aria-label={label}>
+        <ElMobileNavItemListItem role="none">
+          <ElMobileNavItemExpanderButton
+            type="button"
+            role="menuitem"
```
kurtdoherty, line 42: **note:** similar to my comment above, the Disclosure pattern doesn't show the use of some of the roles you're assigning to some of these components...

src/components/mobile-nav-item/__tests__/__snapshots__/mobile-nav-item.test.tsx.snap
```diff
@@ -0,0 +1,127 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`MobileNavItem should render Expandable MobileNavItem 1`] = `
+<DocumentFragment>
+  <mock-styled.ul
```
kurtdoherty, line 5: **suggestion:** I think we may want the parent element of a MobileNavItem to be an `<li>` element, right?

src/components/mobile-nav-item/types.ts
```diff
@@ -0,0 +1,21 @@
...
+type CommonMobileNavItemProps = {
+  label: string
+  isActive?: boolean
+  hasBadge?: boolean
+}
+
+export interface MobileNavItemSimpleProps extends CommonMobileNavItemProps {
+  children?: never
+  isDefaultExpanded?: never
+  href: string
```
kurtdoherty, line 12: **suggestion:** I think we need to support `onClick` too?

src/components/mobile-nav-item/mobile-nav-item.tsx
kurtdoherty: **question:** do we need to have a "sub item" component or could we just allow consumers to put a MobileNavItem inside a MobileNavItem? It does mean they could have nested expandable items, but it's not clear to me if that's something we should explicitly prevent or not 🤔

src/components/mobile-nav-item/styles.ts
```diff
@@ -0,0 +1,96 @@
...
+  &[data-is-expanded='true'] {
+    background: var(--fill-default-lightest, #f2f4f6);
+
+    ${ElMobileNavItemListItem} > ${ElMobileNavItemExpanderButton} {
+      border-radius: var(--corner-lg, 8px) var(--corner-lg, 8px) var(--corner-none, 0) var(--corner-none, 0);
+    }
+
+    ${ElMobileNavSubItemUnorderedList} > ${ElMobileNavItemListItem}:last-child {
+      border-radius: var(--corner-none, 0) var(--corner-none, 0) var(--corner-lg, 8px) var(--corner-lg, 8px);
+    }
```
kurtdoherty, line 88: **suggestion:** you can avoid all this `border-radius` faff by placing the border-radius on the `<ul>` and setting `overflow: hidden` on the `<ul>` as well. This way, the `<li>` elements are effectively "clipped" by the `<ul>`s shape.

See https://codesandbox.io/p/sandbox/3jkrqc for what I mean. If I'm missing something with the CSS or DOM structure here, that's okay.

kurtdoherty (commented):
Thanks for all the work on this @ss-dimasm!

src/components/mobile-nav-item/mobile-nav-item.tsx
ss-dimasm: > do we need to have a "sub item" component or could we just allow consumers to put a MobileNavItem inside a MobileNavItem

technically "item"and "sub item" is similar in terms of UI, and the only reason why I declare it to be "sub item" is to prevent that scenario of having nested expandable in an expandable

> ...but it's not clear to me if that's something we should explicitly prevent or not

seems it's worthy to ask the Design for allowing the expandable in an expandable for mobile navigation menu, I'll update the conversation update for this

src/components/mobile-nav-item/mobile-nav-item.tsx
ss-dimasm: > I'm not sure menubar is the correct role to use here. If we're trying to follow the [Disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/), it doesn't use this role at all...

Ah yes, I got mixed up while trying to research between `Disclosure` and `Menu` pattern, apologies I'll give the update 🤦

src/components/mobile-nav-item/mobile-nav-item.tsx
ss-dimasm: > do we need a "default expanded" prop?

I think it's worthy to have the "default expanded" behavior, so the UI would automatically sync with the passed state by default (in case, when it's on mounting phase)

>  if isActive is true, would we not want the nav item to be expanded?

we could consider this to only rely on single prop that has the right responsibility, seems just by relying the `isActive` is more than enough instead adding the additional one

src/components/mobile-nav-item/types.ts
ss-dimasm: > Do we really need this separation between "simple" and "expandable" props for the component?

Yes, this should be separated since the "expandable" and "simple" variants are actually different in terms of rendering node

simple is equal with `HTMLAnchorElement`, and expandable is equal with `HTMLButtonElement`; and I have confirm this with Andrei that the parent of expandable is only have the expand and collapse responsobility

src/components/mobile-nav-item/mobile-nav-item.tsx
kurtdoherty: Even if Design don't want to permit nested expandable items from a Design System guidelines perspective, we don't necessarily have to go to effort to prevent that from a code perspective. Engineers will be implementing designs produced by the design team, so the design team just won't produce designs that have nested expandable items. Plus, with or without `MobileNavItem.SubItem`, it will still be possible for someone to do:

```
<MobileNavItem ...>

  <MobileNavItem ...>
    <MobileNavItem.SubItem ... />
    <MobileNavItem.SubItem ... />
  </MobileNavItem ...>

  <MobileNavItem.SubItem ... />
  <MobileNavItem.SubItem ... />
</MobileNavItem ...>
```

From a guidelines perspective, this would be wrong, but its also not a costly mistake. So yeah, I'm not sure there's a whole lot of value have SubItem involved, but it's also not costing us a lot to have it, and it probably does help communicate the correct intention 🤷

src/components/mobile-nav-item/__tests__/__snapshots__/mobile-nav-item.test.tsx.snap
ss-dimasm: >I think we may want the parent element of a MobileNavItem to be an <li> element, right?

thanks for pointing this out, I missed this crucial things really 🙂, because it's a list it should be rendered as `<li>`, and let's the consumer would declare the `<ul>` and `<nav>`

src/components/mobile-nav-item/types.ts
ss-dimasm: >  I think we need to support onClick too?

good catch on this

src/components/mobile-nav-item/styles.ts
ss-dimasm: > See https://codesandbox.io/p/sandbox/3jkrqc for what I mean

I tried to visit the link, but unfortunately I can't see anything on there; I'll try to implement about what you mean in the next commit 😉

src/components/mobile-nav-item/mobile-nav-item.tsx
ss-dimasm: my bad, yes that's a typo

src/components/mobile-nav-item/types.ts
kurtdoherty: Okay great. If we don't want an href to be valid on the the item when it has children, what you've got it what we need 👍

src/components/mobile-nav-item/styles.ts
kurtdoherty: Sorry, it wasn't shared correctly. Should be accessible now.

ss-dimasm: thanks, I got your point for that 👍

ss-dimasm: Hi @kurtdoherty , thanks for the snippet. The purpose of defining or changing the border-radius is to modify the UI when it's expanded (applicable only to the expandable variant)

This would affect three sections:

1. Expandable trigger (top-left and top-right corners would have a corner)
2. Expandable content (no corners)
3. Expandable content's last item (bottom-left and bottom-right corners)

Please let me know if I'm misunderstanding your suggestion. It would be great if you could review the latest changes I've made, thanks

kurtdoherty: Yes, I think you're misunderstanding the CodeSandbox I provided. Here, you're having to adjust the border radii of the elements within the three sections that you describe above.

My CodeSandbox is trying to demonstrate that you can take a simpler approach by defining the border radius of the parent element and hiding any overflow. This allows the child elements to have no border radius, but for the group as a whole to still have the correct radius visually (thanks to the parent).

If you still don't understand, it's not a big.

kurtdoherty (approved):

src/components/mobile-nav-item/styles.ts
ss-dimasm: ah yes, so the border radius is the only one that has the `border-radius` and `overflow` CSS properties, and then the rest of the child nodes would be rectangles. It's more simple agree

But with the current API, I think a refactor is required since the parent and the child nodes are actually the same component (`ElMobileNavItemListItem`). Perhaps using a data attribute could help us identify which is the parent and which is the child

Looking at the current condition, it seems we could leave it as it is for now and see how it works when design verifies the spec

kurtdoherty: > But with the current API, I think a refactor is required since the parent and the child nodes are actually the same component (ElMobileNavItemListItem). Perhaps using a data attribute could help us identify which is the parent and which is the child

Yes, the data attribute is what distinguishes single items from expandable items, just like with the current CSS. All good though. This is all implementation detail that can be improved on in future if we need.

PR #210, opened 2024-11-18
feat: #195 new app-bar (avatar button)

nurm717123:

***app bar (parent)***
image

***this ticket***
image

#### preview storybook example

https://github.com/user-attachments/assets/712dfbeb-31d1-4aae-be79-f68564dae636

### DIFF
--- a/src/components/avatar-button/avatar-button.tsx
+++ b/src/components/avatar-button/avatar-button.tsx
@@ -4,11 +4,16 @@ import { ElAvatarButton } from './styles'
 
 export interface AvatarButtonProps extends HTMLAttributes<HTMLButtonElement> {
   label: string
+  ariaLabel?: string
 }
 
-export const AvatarButton: FC<AvatarButtonProps> = ({ label, ...props }) => {
+export const AvatarButton: FC<AvatarButtonProps> = ({
+  label,
+  'aria-label': ariaLabel = 'Open user navigation menu',
+  ...props
+}) => {
   return (
-    <ElAvatarButton aria-label="User Avatar" {...props}>
+    <ElAvatarButton aria-label={ariaLabel} {...props}>
       <Avatar size="small" shape="circle" colour="purple">
         {label}
       </Avatar>
--- a/src/components/avatar-button/avatar-button.tsx
+++ b/src/components/avatar-button/avatar-button.tsx
@@ -1,0 +1,17 @@
+import type { FC, HTMLAttributes } from 'react'
+import { Avatar } from '../avatar/avatar'
+import { ElAvatarButton } from './styles'
+
+export interface AvatarButtonProps extends HTMLAttributes<HTMLButtonElement> {
+  label: string
+}
+
+export const AvatarButton: FC<AvatarButtonProps> = ({ label, ...props }) => {
+  return (
+    <ElAvatarButton aria-label="User Avatar" {...props}>
+      <Avatar size="small" shape="circle" colour="purple">
+        {label}
+      </Avatar>
+    </ElAvatarButton>
+  )
+}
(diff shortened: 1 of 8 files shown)


src/components/avatar-button/styles.ts
```diff
@@ -0,0 +1,23 @@
...
+export const ElAvatarButton = styled.button`
+  cursor: pointer;
+  background: none;
+  border: none;
+
+  &:focus-visible {
+    outline: none;
+  }
+
+  &:focus-visible ${ElAvatar} {
```
nurm717123, line 13: as per Design spec, there is a description written under anatomy of `button` component that implies to use `focus-visible` https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-24543&m=dev

src/components/avatar-button/styles.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ElAvatar } from '../avatar/styles'
+
+export const ElAvatarButton = styled.button`
+  cursor: pointer;
+  background: none;
+  border: none;
+
+  &:focus-visible {
+    outline: none;
+  }
```
nurm717123, line 11: override default button style

kurtdoherty (approved):
Thanks @nurm717123  🙌  Just a few minor comments.

src/components/avatar-button/styles.ts
nurm717123: > issue: focus should not make the avatar white. Please double check the designs 🙏 It looks like there is no background colour change for focus.

Ah yes, definitely 🤦 

> question: is there a reason for targeting the ElAvatar element here? Why can't we just apply the focus styling to the button itself?

Its possible, but I have to add additional style to the button `border-radius: 100%; padding: 0;`, I just thought to add it to the `el-avatar` is simpler 🤔 , what do you think?

src/components/avatar-button/avatar-button.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Avatar } from '../avatar/avatar'
+import { ElAvatarButton } from './styles'
+
+export interface AvatarButtonProps extends HTMLAttributes<HTMLButtonElement> {
+  label: string
+}
+
+export const AvatarButton: FC<AvatarButtonProps> = ({ label, ...props }) => {
+  return (
+    <ElAvatarButton aria-label="User Avatar" {...props}>
```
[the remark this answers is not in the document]
nurm717123: > Also, this aria label is currently acting as a default value, because if an aria-label is provided in props, it'll override the value you're supply here. We should avoid doing this for defaults and instead leverage defaults (i.e. ({ 'aria-label: ariaLabel = 'blah', label, ...rest }))

ok, sure

src/components/avatar-button/styles.ts
kurtdoherty: > Its possible, but I have to add additional style to the button border-radius: 100%; padding: 0;, I just thought to add it to the el-avatar is simpler 🤔 , I will update that

Ahh, fair point. No leave it as is 👍

nurm717123: I just update it and merge without seeing this thread again 🤦 , let me know if I need to revert it in another pr

kurtdoherty: haha no stress. I think your approach was correct. Avatar has two shapes, circle and square, so we want the box-shadow to match the shape of the Avatar, not the button. I believe what you had originally was the simplest way of achieving that 👍 

Not a big deal though.

kurtdoherty:
@nurm717123 just noticed this duplication of props in the AvatarButton stories. Would be good to fix this up when you tweak the focus styling 🙏 

Screenshot 2024-11-19 at 4 10 41 PM

nurm717123:
> @nurm717123 just noticed this duplication of props in the AvatarButton stories. Would be good to fix this up when you tweak the focus styling 🙏
> 
> Screenshot 2024-11-19 at 4 10 41 PM

thanks, noted.

PR #212, opened 2024-11-19
task: #159 Replace Jest with Vite and Vitest

### LINKED ISSUE #158: Remove Linaria mocks from Jest global setup

### LINKED ISSUE #159: Replace Jest with Vite + Vitest

kurtdoherty:
### Context

TL;DR Jest is hard to configure so that the Linaria CSS extraction on components is performed during test runs. Switching to a Vite + Vitest combination is, by comparison, trivial to achieve. This CSS extraction is important when rendering components via React Testing Library (RTL), because without it, the correct styles are not applied to the component under test, undermining many RTL DOM queries.

More details on this issue can be found in #152.

## This PR

- Fixes #159;
- Fixes #158 (as Jest has been replaced);
- Replaces the use of Rollup with Vite and Jest with Vitest.
  - Vite has a much simpler configuration than Rollup;
  - Vitest uses the Vite config in `vite.config.ts`, so it leverages the same plugins as our production build;
  - Storybook will also use the Vite config in `vite.config.ts` when it builds the stories for our components.
- While Vitest and Jest have similar APIs, they are not identical, so the following changes were also necessary:
  - `jest.fn` -> `vi.fn`, `jest.mock` -> `vi.mock`
  - Vitest is ESM only, which means module mocking works differently. This did not require many changes though.

Other changes include:

- Since Vite and Vitest are ESM first tools, `"type": "module"` has been added to the package's manifest;
- To provide a more modern definition of the package's exports for consumers using modern tools like Vite, an export map has been added that defines the package's exports (ESM bundle, CJS bundle and style.css);
- The TypeScript config was updated to use `module: "Preserve"` (which implies `moduleResolution: "Bundler"`). This is necessary because we're building Elements with a bundler. A new build-specific `tsconfig.build.json` has also been introduced. This allows Vite to focus on emitting production JS files while we use `tsc` to emit types in parallel. It also allows us to reduce the work `tsc` does when emitting types by disabling type checking (we do type checking during the `yarn check` script, so there's no need to do it again in the `yarn build` script;
- The `/tokens` folder was moved inside the `/src` folder;
- Refactor of how coverage badges are generated; and,
- The `@/` alias originally configure in both the TS config and the Rollup config has been replaced with an [import map](https://nodejs.org/docs/latest/api/packages.html#imports) in the package's manifest. This provides us with a single source of truth for any internal module mappings we wish to maintain.

**IMPORTANT:** Due to the nature of these changes, it was not possible to incrementally migrate from Jest to Vitest. As such, this PR takes "big bang" approach. I have called attention to the most relevant changes throughout the PR, so you can likely stay on this overview tab and scan through those comments instead of trying to review all 152 files individually. The vast majority of the changeset will be snapshot updates which can largely be ignored.


src/components/chip/__tests__/__snapshots__/chip.test.tsx.snap
```diff
@@ -1,33 +1,35 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`Chip component should match a snapshot 1`] = `
+exports[`Chip component > should match a snapshot 1`] = `
 {
   "asFragment": [Function],
   "baseElement": <body>
```
kurtdoherty, line 6: **note:** I have not swept through all the existing snapshot tests to update them to use `asFragment` instead of capturing the whole `render` result. I will deal with that via #157 once this PR is merged in.

tsconfig.json
```diff
@@ -2,37 +2,27 @@
...
     "noImplicitAny": false,
     "forceConsistentCasingInFileNames": true,
     "strict": true,
-    "typeRoots": ["./node_modules/@types"],
     "sourceMap": true,
     "inlineSources": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "module": "ESNext",
+    "module": "Preserve",
```
kurtdoherty, line 16: **note:** Because we're using a bundler, we need to use `module: Preserve` to ensure TypeScript deals with module imports in the same way as modern bundlers do (e.g. no need for a file extension)

tsconfig.json
```diff
@@ -2,37 +2,27 @@
...
-    "types": ["jest", "node", "react"],
-    "baseUrl": "./",
-    "paths": {
-      "@/*": ["src/*"]
-    },
-    "lib": ["es2015", "es2017", "dom", "es5", "es6", "esnext"],
-    "outDir": "dist",
+    "declarationMap": true,
+    "types": ["vite/client", "vitest/globals", "vite-plugin-svgr/client"],
+    "lib": ["ES2022", "dom"],
```
kurtdoherty, line 21: **note:** ES2022 is a stable option. The inclusion of `esnext` alongside `es5` etc was not necessary.

tsconfig.json
```diff
@@ -2,37 +2,27 @@
...
-    "module": "ESNext",
+    "module": "Preserve",
     "target": "ES2022",
     "declaration": true,
-    "moduleResolution": "node",
-    "types": ["jest", "node", "react"],
-    "baseUrl": "./",
-    "paths": {
-      "@/*": ["src/*"]
-    },
```
kurtdoherty, line 24: **note:** This `paths` config is replaced with the import map in our package manifest.

tsconfig.json
```diff
@@ -2,37 +2,27 @@
...
-    "lib": ["es2015", "es2017", "dom", "es5", "es6", "esnext"],
-    "outDir": "dist",
+    "declarationMap": true,
+    "types": ["vite/client", "vitest/globals", "vite-plugin-svgr/client"],
+    "lib": ["ES2022", "dom"],
+    "outDir": "dist/types",
+    "rootDir": "src",
```
kurtdoherty, line 23: **note:** `rootDir` is required to help TypeScript understand how to resolve the import maps in our package manifest

kurtdoherty:
@willmcvay @adrian-reapit Apologies for the huge PR, but there's no avoiding all the snapshot updates and Vitest changes. As the PR description points out, I've highlighted all the interesting this to be aware of via the comments above, so there shouldn't be any need for you to view individual files unless there's specific things you need to check. Please don't both reviewing all 152 files individually; most of the changes are snapshot/test updates that are boilerplate.

willmcvay (approved):
It's a pretty big PR so hard to comment too deeply - probably needs merging and bedding in. LGTM 👍🏻

PR #214, opened 2024-11-19
fix: #213 Allow `yarn deploy` scripts to run as CJS

### LINKED ISSUE #213: `yarn deploy` fails because deploy scripts use CJS syntax
**Background context:**

#212 migrated the repo to Vite and Vitest. As part of this work, the Elements package now expects all `.js` and `.ts` modules to use ESM syntax. If CJS syntax is required (e.g. `require`), than `.cjs` and `.cts` must be used.

Unfortunately, #212 failed to update the deployment scripts that are run by `yarn deploy` to conform to this new requirement. As such, the Release Beta pipeline encounters the following error:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/home/runner/work/elements/elements/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

It's worth highlighting that the `scripts/cdk` code uses AWS SDK v2, which is in maintenance mode. Upgrading to AWS SDK v3 is strictly not in scope for this work, but will need to be done at some point in the future.

**Specification or tasks to perform:**

We need to enable the `scripts/cdk` code to run successfully.

kurtdoherty:
### Context

#212 migrated the repo to Vite and Vitest. As part of this work, the Elements package now expects all .js and .ts modules to use ESM syntax. If CJS syntax is required (e.g. require), than .cjs and .cts must be used.

### This PR

- Fixes #213 
- Adds `ts-node` back as a dev dependency (#212 incorrectly removed it)
- Renames all the `.ts` files in `scripts/cdk` to `.cts`
- Replaces `yarn dlx ts-node ...` with `yarn node -r ts-node/register --loader ts-node/esm ...`. Without this change, Node would complain about `.cts` being an unknown syntax. This is because `.cts` is an extension that only applies when in an ESM environment.


kurtdoherty:
**note:** it's not possible for me to test the `yarn deploy` script in full locally as it requires credentials that are only defined within the Release Beta pipeline.The best I could do locally was confirm that we get a credentials-related error, which suggests the changes here will enable the pipeline to successfully deploy.

```
Build error:  Error: connect EHOSTUNREACH 169.254.169.254:80 - Local (192.168.68.51:57022)
    ...
    at processTicksAndRejections (node:internal/process/task_queues:85:11) {
  message: 'Missing credentials in config, if using AWS_CONFIG_FILE, set AWS_SDK_LOAD_CONFIG=1',
  code: 'CredentialsError',
  ...
}
```

Because of this, I'll be merging these changes and, if there's additional errors encountered by the pipeline, I'll have to raise a separate branch/PR. This obviously isn't ideal, and it highlights how valuable it is for pipeline code to be testable locally.

kurtdoherty:
**aside:** In future, it would be good to replace our reliance on `ts-node` with the [--experimental-strip-types](https://nodejs.org/api/typescript.html#type-stripping) option Node 22.6 and to update the `scripts/cdk` code to use ESM instead of CJS

kurtdoherty:
FYI @willmcvay looks like there's some tech debt hiding here with our use of AWS SDK v2. I see this in the CLI output when running `yarn deploy`:

```
NOTE: The AWS SDK for JavaScript (v2) will enter maintenance mode
on September 8, 2024 and reach end-of-support on September 8, 2025.
```

PR #215, opened 2024-11-19
task: #157 Avoid capturing React Testing Library internals in DOM snapshots

### LINKED ISSUE #157: Avoid capturing RTL internals in DOM snapshots

kurtdoherty:
Fixes #157


kurtdoherty:
**note:** I'm merging these changes without review because they are simply snapshot updates.

PR #216, opened 2024-11-19
task: #160 remove cross-env in favour of Yarn's built-in shell

### LINKED ISSUE #160: Remove `cross-env` in favour of Yarn's built-in POSIX-like interpreter
### Context

In some of our package manifest scripts, we're using [cross-env](https://github.com/kentcdodds/cross-env) to provide cross-platform (Windows, Linux, MacOS) support for environment variables. It allows scripts to use the POSIX environment variable syntax to define environment variables for the duration of the command being run. For example, `NODE_ENV=production vite` would not be compatible on Windows without `cross-env` being used: `cross-env NODE_ENV=production vite`.

However, since Elements' is using Yarn Modern, we can simply rely on the [portable shell](https://yarnpkg.com/features/scripting#portable-shell) that it provides for all package manifest scripts. This portable shell comes with a few neat features, one of which is POSIX-like environment variable support. This means `NODE_ENV=production vite` does work out-of-the-box across Windows, Linux and MacOS.

### Work to be done

Remove `cross-env` from Elements and rely on Yarn's portable shell interpreter instead.

kurtdoherty:
Closes #160 

We do not need to use `cross-env` in order to support POSIX-like env var assignment in our package scripts. Yarn runs all package scripts within a POSIX-like shell interpreter, thus providing cross-platform support by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,8 @@
   ],
   "scripts": {
     "build": "yarn build:storybook & yarn build:types & yarn build:dist",
-    "build:storybook": "cross-env NODE_ENV=production && rimraf public/dist && storybook build -o public/dist",
-    "build:dist": "cross-env NODE_ENV=production && vite build",
+    "build:storybook": "rimraf public/dist && NODE_ENV=production storybook build -o public/dist",
+    "build:dist": "vite build",
     "build:sd": "node ./src/tokens/build.mjs && eslint --cache --ext=ts tokens --fix",
     "build:types": "tsc --project tsconfig.build.json",
     "commit": "yarn test run --coverage --silent && yarn coverage:badges && yarn lint --fix && yarn check",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -48,8 +48,8 @@
   ],
   "scripts": {
     "build": "yarn build:storybook & yarn build:types & yarn build:dist",
-    "build:storybook": "cross-env NODE_ENV=production && rimraf public/dist && storybook build -o public/dist",
-    "build:dist": "cross-env NODE_ENV=production && vite build",
+    "build:storybook": "rimraf public/dist && NODE_ENV=production storybook build -o public/dist",
```
kurtdoherty, line 51: **note:** I doubt we need to set `NODE_ENV` to production when running `storybook build`, but I've left it here just in case. I know it's not necessary for `vite build` below, because Vite's build command implicitly builds for production unless explicitly told otherwise.

ksolanki7 (approved):
praise: Seems good to me.

Issue #218, opened 2024-11-21
v5 component refactor: Table


ksolanki7:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing?
- Is the component being re-named?
- Is the component design signed off by product and design?
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=1091-5992&node-type=canvas&m=dev)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/table--docs)
- [Solution Design](https://github.com/reapit/elements/discussions/201)

**PR Checklist**
- [ ] 1) Deprecate existing Table Component
           https://github.com/reapit/elements/pull/219
- [ ] 2) Toolbar
- [ ] 3) New Table Atoms
- [ ] 4) Row Selection
- [ ] 5) Column Sort
- [ ] 6) Accessibility

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

PR #219, opened 2024-11-21
refactor: #218 v5 Deprecate existing Table component

ksolanki7:
- Deprecated current Table component
- Only deprecated exports for which the naming will be required for new Table component


src/components/deprecated-table/__tests__/molecules.test.tsx
```diff
@@ -1,10 +1,10 @@
...
-  TableRowContainer,
-  TableCell,
+  DeprecatedTableHeadersRow,
+  DeprecatedTableHeader,
+  DeprecatedTableRow,
+  DeprecatedTableRowContainer,
+  DeprecatedTableCell,
   TableExpandableRowTriggerCell,
   TableExpandableRow,
   TableCtaTriggerCell,
```
kurtdoherty, line 10: **question:** Not sure, but did we need to rename these as well?

kurtdoherty (approved):
Looks good @ksolanki7. I'm starting to wonder if this is the best approach thought 🤔

I'll put some thought into whether there's an easier way for us to deprecate old components and for consumers to still use them without having to rename all their imports...

In the interim, this is good to move ahead 👍

src/components/deprecated-table/__tests__/molecules.test.tsx
ksolanki7: I do not find this useful for the current DS requirement, so I have only renamed the exports I may have intended to use.

Issue #153, opened 2024-08-15
v5 component refactor: App Bar


HollyJoyPhillips:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing 'Nav' 
- Is the component being re-named? Rename to 'App Bar' 
- Is the component design signed off by product and design? Yes 
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-7209&t=wNqPmfN0KOgru0jN-4)
- [Link to Elements Audit in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/App-bar-and-Side-nav-%5BElements%5D?node-id=18-7578&t=AB2erFDHYuLLxvGj-4)
- [Accessibility requirements](https://m3.material.io/components/top-app-bar/accessibility)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-25707&t=mvIlOCZjDYP90TZT-4)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

jmatiushina:
Pointing out the issue I've had as if it might be relevant to the task and helpful for who ever picks it up. I'm passing in `brandOptions`:
```
brandOptions={{
            callback: openNewPage('agencycloud://process/webpage?url=https://mysite.com'),
            logoUrl: 'https://uk.payprop.com/res/assets/img/pp_logo.svg',
          }}
```
As you can see I'm trying to redirect to the webpage outside of the agency cloud.
[...]

PR #221, opened 2024-11-22
feat: #135 init new top-bar

nurm717123:
### Context
- App bar renamed to Top Bar
- App Switcher and Brand/Logo not included yet

***app bar (parent)***
image

**this ticket**
- same as parent but without App Switcher and Brand/Logo

this pr can act as an initiation on how we going to handle the various responsive UI for each top bar items

https://github.com/user-attachments/assets/af2cf407-0039-4110-adb2-7aff2febd884


kurtdoherty (changes requested):
Thanks for the work here @nurm717123 🙏 

**issue:** Unfortunately, I'm a little confused about the approach you've taken and I think there's been some miscommunication on the ticket. From my perspective, we've had a lot of discussion in the ticket about making `TopBar` an _atomic_ component and _not_ pursuing an interface that uses JavaScript arrays or objects to describe the various nav items and other elements. This is effectively the same conversation we had for the Menu component.

I'm wondering if I did not communicate clearly enough in my comment [here](https://github.com/reapit/elements/issues/153#issuecomment-2484796580) 🤔 Re-reading this, I can see how it may have sounded like I was in favour of the second option you presented. If that's the case, I apologise; I'm absolutely in favour of option 1, I was simply suggesting that we may prefer to avoid children and use separate props for the various ReactNode's instead. Either way, the composition of atoms (option 1) is absolutely the preferred design approach—it satisfies SOLID design principles and is consistent with what we've done for Dialog and Menu.

As it stands, the interface for `TopBar` that you've presented in this PR tightly couples it to a number of other components and, among other things, prevents consumers from customising _when_ certain Nav Items should collapse into the "More" item.

**suggestion:** I think we need to rework this PR to take either Option 1 as you presented it, or Option 1 plus specific props for each "section" instead of `children`.

If you take Option 1 as you presented it, then I think the main navigation, secondary navigation and other components would belong to TopBar. That is, we'd provide `TopBar.MainNavigation`, `TopBar.SecondaryNavigation` and so on. If you take Option 1, but didn't rely on `children`, then these components could either belong to `TopBar` or not; I think it would be fine either way.

The choice between `children` or multiple well-named props seems like a minor one to me. If we go with `children`, it would be possible for consumers to not provide something (like the profile menu) without getting any TS errors. With well-named props, we could enforce the need for consumers to supply everything.

If you're not sure where to go from here, I can provide a more specific solution design, or we can line up a call to discuss it verbally.

nurm717123:
Thanks Kurt 👍. I realize now that I misunderstood your comment [here](https://github.com/reapit/elements/issues/153#issuecomment-2484796580). I initially thought it would be easy to use, but since it isn’t consistent with our other new components and the second approach aligns better with the SOLID principles, it makes more sense. For the next approach, I would prefer to choose the one with children (`TopBar.MainNavigation`, .etc), as it will be consistent with other patterns (such as Menu and Dialog) and will make it easier for users to utilize the `QueryContainer` for handling the main navigation items.

kurtdoherty:
> For the next approach, I would prefer to choose the one with children (TopBar.MainNavigation, .etc)

Okay, it might be worth sketching out the interface and how you think it'll work in the ticket. That way we can confirm we're on the same page before you write too much code again 😅

nurm717123:
yeah, that makes sense 😅 , I think something like this e.g:

```javascript
<TopBar>
  <TopBar.LeftSection>
     <TopBar.LeftSection.MainNavigations> // I think 3 sub is the max depth for me personally
       <NavItem />
       <NavItem />
       <QueryContainer>
         <Menu>
           ..... // menu implementation
         </Menu>
       </QueryContainer>
     </TopBar.LeftSection.MainNavigations>
  </TopBar.LeftSection>

  <TopBar.RightSection>
     <TopBar.RightSection.SearchButton /> 
     <TopBar.RightSection.SecondaryNavigations> // I think 3 sub is the max depth for me personally
       <NavIconItem/>
       <NavIconItem/>
       <NavIconItem/>
     </TopBar.LeftSection.MainNavigations>
     <TopBar.RightSection.AvatarButton /> 
  </TopBar.RightSection>
</TopBar>
```

kurtdoherty:
**question:** What benefit do you see the `LeftSection` and `RightSection` components providing here? Could we just remove those?

**question:** What are `SearchButton` and `AvatarButton`? Are they abstractions that wrap up some other atom or are they "container" style components (like MainNavigation and SecondaryNavigation seem to be) that will accept children?

**note:** Not sure if you saw my [discussion with Andrei in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=15-3629&m=dev#1016273484) re MainNavigation and SecondaryNavigation. In it we agreed that the spacing of the items in each nav can just be handled by ButtonGroup. This means the `MainNavigation` and `SecondaryNavigation` etc components are really just there to handle the media query behaviour defined by the TopBar. Does that make sense?

kurtdoherty:
> I think it can help us to easily style the position of for each (left/right) section with flexbox

**question:** Do we need the extra nesting then (i.e. `TopBar.LeftSection.MainNavigation` vs `TopBar.MainNavigation`)?
**question:** If you used CSS grid instead of flexbox, would that remove the need for them?
**question:** Are you concerned about consumers getting the order of the different components incorrect or are you happy to just leave that as an "authoring" concern?

> I think I will add the condition to handle mobile and wide screen in the SearchButton (as my pr in [here](https://github.com/reapit/elements/pull/221/files#diff-d427490dc7db577a682c303f21340f48debe84c66aa4303670003685b2193078)), so the TopBar consumer will just need to import it from there.

**thought:** This feels like it's putting TopBar's media query knowledge into the SearchButton 🤔 Not 100% sure that's appropriate... If we want to do that, I think that suggests the only access to SearchButton should be via TopBar. My gut says to keep SearchButton and the NavIconItem

> I use ButtonGroup with additional style with classname for this in this pr

**question:** What's the need for the additional styles? Can we not keep these TopBar components agnostic of the content being put inside of them? i.e. the styles you've applied mean we only get the correct behaviour if there's an `ElButtonGroup` descendant; this means we're tightly coupled to the use of that component...generally we don't want to create this kind of tight coupling in our CSS because it's "invisible" to consumers, especially when these TopBar components otherwise accept any `ReactNode`.

nurm717123:
> question: Do we need the extra nesting then (i.e. TopBar.LeftSection.MainNavigation vs TopBar.MainNavigation)?

yes, if we using flexbox (following the design style)

> question: If you used CSS grid instead of flexbox, would that remove the need for them?

yes, we don't need the left/right section wrapper with that

> question: Are you concerned about consumers getting the order of the different components incorrect or are you happy to just leave that as an "authoring" concern?

I think its ok to let the react consumer to structure as their own, as that what non react user also do, but if there is way to enforce the ordering that would be great.

> thought: This feels like it's putting TopBar's media query knowledge into the SearchButton 🤔 Not 100% sure that's appropriate... If we want to do that, I think that suggests the only access to SearchButton should be via TopBar. My gut says to keep SearchButton and the NavIconItem

yeah, I'm not sure if the SearchButton will be used elsewhere, I will separate it then so it more flexible👍

> question: Why do we need to apply these additional styles to the ButtonGroup? Can we not keep these TopBar components agnostic of the content being put inside of them? i.e. the styles you've applied mean we only get the correct behaviour if there's an ElButtonGroup descendant; this means we're tightly coupled to the use of that component...generally we don't want to create this kind of tight coupling in our CSS because it's "invisible" to consumers, especially when these TopBar components otherwise accept any ReactNode.

do you mean we have to provide an atom based on `ButtonGroup` for top-bar, I think that make sense

kurtdoherty:
> > question: Do we need the extra nesting then (i.e. TopBar.LeftSection.MainNavigation vs TopBar.MainNavigation)?
>
> yes, if we using flexbox (following the design style)

I must be missing something here because I don't understand what using a CSS flexbox layout has to do with LeftSection and RightSection? I can't see any reason why we couldn't use flexbox without those components in the mix 🤔 

> I think its ok to let the react consumer to structure as their own, as that what non react user also do, but if there is way to enforce the ordering that would be great.

Okay. I think order is much more important for this TopBar than, say, for the Dialog, but I don't feel strongly about it. I think it's probably okay to leave it as an authoring concern (i.e. it's to devs to order things correctly, like search button before mobile nav menu etc).

> do you mean we have to provide an atom based on ButtonGroup for top-bar, I think that make sense

No, that's not what mean sorry. I'm asking why `ElTopBarRightContent` would apply `display: none` to the `ElButtonGroup` instead of `ElMainNavContainer`. By applying styles to `ElButtonGroup`, you're making `ElTopBarRightContent` tightly coupled to a completely separate atom it should know nothing about. In contrast, `ElMainNavContainer` is not a separate atom, so it's perfectly reasonable for `ElTopBarRightContent` to apply styles to `ElMainNavContainer`. That said, it seems like `ElMainNavContainer` should be responsible for when it displays and when it doesn't...

kurtdoherty:
**note:** Here's a quick HTML + CSS POC of a top bar and other "components" that leverages CSS grid to enforce correct visual order + has the media query behaviour baked in: https://codesandbox.io/p/sandbox/2f58yz. There may be aspects we could improve on here but hopefully it gives you some ideas.

nurm717123:
> I must be missing something here because I don't understand what using a CSS flexbox layout has to do with LeftSection and RightSection? I can't see any reason why we couldn't use flexbox without those components in the mix 🤔

aplogise, actually its still possible without left/right container, but it would need additional style (like flex-grow for the main nav container, margin auto .etc), its just not following the figma structure, I thought that we have to strictly follow figma structure 🤔 

> No, that's not what mean sorry. I'm asking why ElTopBarRightContent would apply display: none to the ElButtonGroup instead of ElMainNavContainer. By applying styles to ElButtonGroup, you're making ElTopBarRightContent tightly coupled to a completely separate atom it should know nothing about. In contrast, ElMainNavContainer is not a separate atom, so it's perfectly reasonable for ElTopBarRightContent to apply styles to ElMainNavContainer. That said, it seems like ElMainNavContainer should be responsible for when it displays and when it doesn't...

Ah yes, I missed that 👍

kurtdoherty:
> I thought that we have to strictly follow figma structure 🤔

Ahh, I see. No, we don't always want to follow what the design team do in Figma. They will do things a certain way in order to make it easy for designers to produce designs for products. These decisions they make will not always align with what we need to do as engineers. Generally, we want:

- Component and prop naming to be as aligned as much as possible;
- Component hierarchy to match what the Design System specifies (e.g. we don't want to couple one component to another unless the Design System also does that).

I'll try to document all this a bit more clearly to help avoid confusion in future 👍

src/components/top-bar/top-bar.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+        </TopBar.SearchContainer>
+
+        <TopBar.SecondaryNavigations>
+          <NavIconItem aria-label="example" icon={<Icon icon="star" />} />
+          <NavIconItem aria-label="example" icon={<Icon icon="star" />} />
+          <NavIconItem aria-label="example" icon={<Icon icon="star" />} />
+        </TopBar.SecondaryNavigations>
+        <TopBar.MenuButtonContainer>
+          <NavIconItem aria-label="mobile secondary nav trigger" icon={<MenuIcon className={elIcon} />} />
+        </TopBar.MenuButtonContainer>
```
nurm717123, line 144: with this pattern user have to implement the desktop and mobile behaviour (search icon button, menu button .etc) by themself

src/components/top-bar/top-bar.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <NavItem href="/">Button 2</NavItem>
+          <NavItem href="/">Button 3</NavItem>
+          <NavItem href="/">Button 4</NavItem>
+          <ContainerQuery conditions={'(width < 1000px)'} containerName="main-nav">
+            <NavItem href="/">Button 5</NavItem>
+            <NavItem href="/">Button 6</NavItem>
+            <NavItem href="/">Button 7</NavItem>
+          </ContainerQuery>
+          <ContainerQuery not conditions={'(width < 1000px)'} containerName="main-nav">
+            <Menu>
```
nurm717123, line 118: here consumer have to implement container query themself

src/components/top-bar/top-bar.stories.tsx
nurm717123: because of this I think this part is more efficient if we use the props approach instead .e.g `<TopBar.SecondaryNav items={items} />` that also can be applied for `MainNav` and `TopBar.SearchButton`.

nurm717123:
Thanks @kurtdoherty , I have addressed the feedback and anticipate there may be more to discuss moving forward (which I hadn’t expected when you initially offered to have a call 😅). If you believe a longer discussion is needed, I would be happy to arrange a follow-up meeting.

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const ElMainNavContainer = styled(ElButtonGroup)`
+  display: flex;
+  padding: var(--spacing-1, 4px) var(--spacing-none, 0px) var(--spacing-1, 4px) var(--spacing-6, 24px);
+  align-items: flex-start;
+  flex-grow: 1;
+`
+
+export const ElTopBarSecondaryNavContainer = styled(ElButtonGroup)`
+  flex-wrap: nowrap;
```
nurm717123, line 24: this secondary nav naming has `TopBar` prefix because there is other component exported with the same name i.e under `layout/styles.tsx`, I think I will need to add prefix to the others too

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+  display: flex;
+  padding: var(--spacing-1, 4px) var(--spacing-none, 0px);
+  padding-left: var(--spacing-2, 8px);
+  align-items: center;
+  gap: 10px;
+`
+
+export const ElTopBarMainNav = styled(ElButtonGroup)`
+  display: flex;
+  padding: var(--spacing-1, 4px) var(--spacing-none, 0px) var(--spacing-1, 4px) var(--spacing-6, 24px);
```
kurtdoherty, line 18: **note:** I realise you're probably just copying and pasting the CSS that Figma provides for these components, but I think this is another area where we need to apply our own experience and judgement to our code. The designers (and Figma) do not care what the code is like to maintain, but we do. As such, I think its important that we, as engineers, don't just copy-paste what other tools suggest; rather we should think critically about what is needed to achieve the required layout.

In this case, I think all we need is:
```css
display: flex;
flex-direction: row;
align-items: center;
padding-left: var(--spacing-6, 24px);
flex-grow: 1;
```

Using `align-items: center` allows us to avoid the complex padding definition and simplify it down to specifying the `padding-left`.

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+  ${ElButtonNavIconItem} {
+    display: none;
+
+    ${isMobile} {
+      display: inline-block;
+    }
+  }
+`
+
+export const ElTopBarMenuButtonContainer = styled.div`
```
kurtdoherty, line 78: **suggestion:** I think `ElTopBarMobileNav` is clearer

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+  ${isTabletOrBelow} {
+    ${elTopBarAvatar} {
+      display: none;
+    }
+  }
+`
+
+export const ElTopBarSearchContainer = styled.div`
+  padding: var(--spacing-1, 4px) var(--spacing-4, 16px) var(--spacing-1, 4px) var(--spacing-none, 0px);
+  ${ElNavSearchButton} {
```
kurtdoherty, line 57: **issue:** Doing this couples us tightly to the `ElNavSearchButton`. We need to avoid doing this wherever possible. We don't need to set the width of `ElNavSearchButton`, we only need to set the width of `ElTopBarSearchContainer` and the component placed inside of it should respect that width.

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  ${isTabletOrBelow} {
+    ${elTopBarAvatar} {
+      display: none;
+    }
+  }
+`
+
+export const ElTopBarSearchContainer = styled.div`
```
kurtdoherty, line 55: **suggestion:** I think `ElTopBarSearch` is simpler and is more aligned with the other styled components here (like `ElTopBarMainNav` and `ElTopBarSecondaryNav`).

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+  padding: var(--spacing-1, 4px) var(--spacing-4, 16px) var(--spacing-1, 4px) var(--spacing-none, 0px);
+  ${ElNavSearchButton} {
+    width: 216px;
+  }
+
+  ${isMobile} {
+    padding-right: var(--spacing-spacing-2, 8px);
+
+    ${ElNavSearchButton} {
```
kurtdoherty, line 64: **issue:** again, we don't need to make the `ElNavSearchButton` `display: none`, rather we can make `ElTopBarSearchContainer` `display: none` and ensure our behaviour (not displaying at a certain screen size) does not depend on the specific component a consumer places inside `ElTopBarSearchContainer`.

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+      display: none;
+    }
+  }
+
+  ${ElButtonNavIconItem} {
```
kurtdoherty, line 69: **issue:** as above

src/components/top-bar/top-bar.tsx
```diff
@@ -0,0 +1,28 @@
...
+  MenuButtonContainer: typeof ElTopBarMenuButtonContainer
+} = ({ children, ...props }) => {
+  return (
+    <nav>
+      <ElTopBar {...props}>{children}</ElTopBar>
+    </nav>
+  )
+}
+
+TopBar.MainNavigations = ElTopBarMainNav
```
kurtdoherty, line 23: **issue:** can we please keep the name of these compound components matching the styled components? 🙏 

If you adopt the name changes I've suggested earlier, this would become:
```
TopBar.MainNav = ElTopBarMainNav
TopBar.SecondaryNav = ElTopBarSecondaryNav
TopBar.Search = ElTopBarSearch
TopBar.MobileNav = ElTopBarMobileNav
```

kurtdoherty (commented):
Thanks for the all the work on this so far @nurm717123 👏 There's a fair bit of feedback, but only some of it blocking.

Aside from my comments in this PR, I'll be interested to see what the use of `children` here (instead of specific, well-named props) is like for consumers. There's certainly some nuance involved in the order of the different components and I wonder if some teams will get that wrong without immediately noticing. That said, if we can make sure we're not tightly coupling TopBar's behaviour to the presence of specific children, we'll be able to iterate easily on the interface in future. Hence why I'm flagging things that are introducing tight coupling 👍

src/components/container-query/container-query.tsx
```diff
@@ -0,0 +1,22 @@
+import { useRef } from 'react'
+
+let baseId = 1
+
+export function ContainerQuery({ children, conditions, not = false, containerName = 'container' }) {
```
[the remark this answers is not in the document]
nurm717123: > todo: Please define a prop interface for this component 🙏 Also, can we call this component CSSContainerQuery, just to help make it clear to readers that this component is related to CSS container queries, not a React Query kind of component.

yes, sure already on my todo but I forgot to add it 😅

> todo: Looking at the [MDN @container docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@container#values) more closely, the syntax for the @container rule is @container <container-condition> { ... } where <container-condition> consists of an optional <container-name> and a <container-query>. As such, I think we should make the following changes:

since I separate this component outside the `topBar` so it can be used anywhere I just thought that by providing `containerName` it would be useful for complex usage, e.g nested container 🤷‍♂️ 

> Rename conditions to condition;
Remove containerName and not. These can just be specified as part of the condition prop;
When demonstrating the use of this in TopBar, don't specify a container name (it works fine in my CodeSandbox without specifying it). This just helps us avoid the awkwardness of TopBar consumers wanting to know the container name use by the Main Nav component.

yes, sure 👍

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+import { ElButtonGroup } from '../button-group'
+import { ElNavSearchButton } from '../nav-search-button/styles'
+import { ElButtonNavIconItem } from '../nav-icon-item'
+
+export const elTopBarAvatar = css`
+  display: flex;
+  padding: var(--spacing-1, 4px) var(--spacing-none, 0px);
+  padding-left: var(--spacing-2, 8px);
+  align-items: center;
+  gap: 10px;
```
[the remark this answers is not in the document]
nurm717123: ah yes that is from Figma, and I missed to delete it 🤦‍♂️

src/components/top-bar/styles.ts
nurm717123: yes, thanks for the clarity👍

src/components/top-bar/styles.ts
nurm717123: because the `search-button` itself has defined padding (400px), but as per previous comment I think we can either delete that defined width or add additional `max-width: 100%` to `search-button` so it can respect the container 👍

src/components/top-bar/styles.ts
nurm717123: actually `ElTopBarSearchContainer` have 2 child, which search button will display in big screen, and search icon for small screen,

```javascript
  <TopBar.Search>
     <NavIconItem icon={<Icon icon="search" />} aria-label="nav-icon-item-example" />
     <NavSearchButton />
  </TopBar.Search>
```

src/components/top-bar/top-bar.tsx
nurm717123: yes, thanks for the example 👍

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,87 @@
...
+  padding: var(--spacing-1, 4px) var(--spacing-none, 0px) var(--spacing-1, 4px) var(--spacing-6, 24px);
+  align-items: flex-start;
+  flex-grow: 1;
+`
+
+export const ElTopBarSecondaryNav = styled(ElButtonGroup)`
+  flex-wrap: nowrap;
+`
+
+export const ElTopBar = styled.div`
```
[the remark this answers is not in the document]
nurm717123: > **suggestion:** One thing I didn't like with my CodeSandbox that showed a CSS grid approach to the TopBar was having all the media query responsiveness spread amongst the various subcomponents. I think I may have recommended doing this in our discussion on the ticket, but having seen it in my example and here, I'm not sure it's the most readable approach to take.
> 
> I've updated my CodeSandbox to co-locate all that logic within the TopBar component and I think it's much easier for a reader to understand what's happening with all the different subcomponents AND much easier to update if needed in future. See https://codesandbox.io/p/sandbox/poc-topbar-with-css-grid-2f58yz

Yes, I think that could work, but how about using this approach: https://codesandbox.io/p/sandbox/app-bar-grid-mxm4l7? Using a grid for a single-line layout feels too much for me it could be more readable though 🤔 .

nurm717123:
> Thanks for the all the work on this so far @nurm717123 👏 There's a fair bit of feedback, but only some of it blocking.
> 

Thanks Kurt, as usual I agree to most of your suggestion, probably just small part that need more discussion 👍 

> Aside from my comments in this PR, I'll be interested to see what the use of `children` here (instead of specific, well-named props) is like for consumers. There's certainly some nuance involved in the order of the different components and I wonder if some teams will get that wrong without immediately noticing. That said, if we can make sure we're not tightly coupling TopBar's behaviour to the presence of specific children, we'll be able to iterate easily on the interface in future. Hence why I'm flagging things that are introducing tight coupling 👍

I'm totally okay with using well-named props. Previously, I thought you preferred a non-props approach because it’s more flexible. However, after making the previous changes, I believe that using props will be much more useful and easier for consumers. For example, we can provide `<TopBar.SearchBar onClick={handler} />`, which would work for `searchBar` and `secondaryNav`, but not for `mainNav`, as it requires the consumer to use `queryContainer`. For secondaryNav, I think it would need additional props because, on mobile screens, it would be responsible for displaying `mobileNavMenu`, which contains all (main-nav, secondary-nav, and user-menu) items. I believe this version would be less tightly coupled than my first approach (react shorthand with single component with combined props).

kurtdoherty:
> Previously, I thought you preferred a non-props approach because it’s more flexible.

To clarify, using props that accept objects or arrays of objects is what leads us into trouble. Using props that accept `ReactNode`'s is perfectly fine. I'm not asking you to change from the current `children` approach to individual, well-named props. I'm just pointing out that we may want to check how engineers go with this interface; if we learn they make mistakes, we can adjust things then.

Here's a [preview of the prop interface design guidelines](https://github.com/user-attachments/files/17942371/DRAFT.Prop.interface.design.pdf) I've been working on the last few days. It was in writing the "On Children" section that I started to notice the pros/cons of using `children` for a component like `TopBar` (previously, I didn't think there was much difference). Have a read of that and hopefully it helps you understand where I'm coming from.

Again, I don't think there's any need to change what we've done here with `children` so please don't go and refactor everything. What you've done here is great (so long as we keep things decoupled where they need to be decoupled).

src/components/top-bar/styles.ts
kurtdoherty: Ah yes, my apologies. I think the evidence for making `NavSearchButton` take care of the button to nav icon item change at the mobile is mounting 🤔

This would simplify the above to:
```
<TopBar.Search>
  <NavSearchButton />
</TopBar.Search>
```

And `TopBar.Search` would only need to be adjusting it's right padding. What do you think?

src/components/top-bar/styles.ts
kurtdoherty: I don't think the search button should have a fixed width. Looking at the  design spec for it, I don't think that 400px is intentional as the use of the component in the top bar maxes out at 200px.

src/components/top-bar/top-bar.stories.tsx
kurtdoherty: > I think this part is more efficient if we use the props approach instead .e.g <TopBar.SecondaryNav items={items} />

As per my response on the PR's overview tab, we do not want to move to an interface based on an array of objects. That walks us into a whole lot of downsides; it's much better for us to stick with composing atoms via JSX instead of abstracting atoms into JavaScript objects.

src/components/top-bar/styles.ts
nurm717123: > And TopBar.Search would only need to be adjusting it's right padding. What do you think?

yes, I think thats simpler, and then `<NavIconItem icon={<Icon icon="search" />} />` will be part of `NavSearchButton`  right?

src/components/top-bar/styles.ts
nurm717123: ok, I will delete it then 👍

src/components/top-bar/styles.ts
kurtdoherty: Yes, which I think may have been something you had originally suggested 😅 If so, apologies for taking a while to see the value haha

nurm717123: No worries, I'm glad we're on the same page now. 😄

src/components/top-bar/styles.ts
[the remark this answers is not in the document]
kurtdoherty: Yeah, I'm not fussed if we use CSS grid or flex. I just think my original POC were I had media queries per "section" wasn't as simple to understand as putting them all together in the one media query 👍

src/components/top-bar/styles.ts
```diff
@@ -1,11 +1,9 @@
...
 import { styled } from '@linaria/react'
-import { isDesktopOrBelow, isMobile, isTabletOrBelow } from '../../styles/media'
+import { isDesktop, isDesktopOrBelow, isTablet, isTabletOrBelow } from '../../styles/media'
 import { ElButtonGroup } from '../button-group'
-import { ElNavSearchButton } from '../nav-search-button/styles'
-import { ElButtonNavIconItem } from '../nav-icon-item'
+import { AvatarButton } from '../avatar-button'
 
-export const elTopBarAvatar = css`
+export const ElTopBarProfile = styled(AvatarButton)`
```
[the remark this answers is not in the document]
nurm717123: Ah yes, I was a little confused because TopBar.Search previously had multiple children. However, after updating it to a single child, I shouldn't have missed it 🤦 .

src/components/avatar-button/styles.ts
```diff
@@ -5,6 +5,7 @@ export const ElAvatarButton = styled.button`
   cursor: pointer;
   background: none;
   border: none;
+  padding: 0;
```
nurm717123, line 8: override default button browser's default `padding` property

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  ${isDesktop} {
+    display: none;
+  }
+`
+
+export const elTopBarMenuPopover = css`
+  // To adjust the menu popover's Y offset so it appears below the top bar instead of directly below the button.
+  // The "important" rule is used to override the Menu's auto-anchor yOffset.
+  top: 44px !important;
```
nurm717123, line 68: Unfortunately, there is no CSS variable that fits this requirement. The closest option is probably `spacing-8`, which is 2.5rem, but 44px is equal to 2.75rem. Since this is likely a specific case and the top bar height is fixed, I believe we can proceed with this approach.

src/components/menu/menu-popover.tsx
```diff
@@ -22,14 +22,16 @@ export const calculatePopoverPosition = (
...
+export const MenuPopover: FC<
+  {
+    /**
+     * Optional parameter to adjust the vertical position of the popover
+     *
+     * @default 4 // confirmed by design team to use var(--spacing-1) initially
+     */
+    yOffset?: number
+  } & HTMLAttributes<HTMLDivElement>
+> = ({ children, yOffset = 4, ...props }) => {
```
kurtdoherty, line 34: **question:** What's the need for the `...props` here?

src/components/menu/menu-popover.tsx
```diff
@@ -55,7 +57,7 @@ export const MenuPopover: FC<{
   }
 
   return (
-    <ElMenuPopover style={popoverStyle} ref={popoverRef} {...getPopoverProps()} onClick={handleClick}>
+    <ElMenuPopover style={popoverStyle} ref={popoverRef} {...getPopoverProps()} onClick={handleClick} {...props}>
```
kurtdoherty, line 60: **issue:** technically speaking, we should really be passing all additional props to `getPopoverProps` so it can correctly handle prop collisions, but I don't believe it supports that yet. For now, we probably just need to ensure `props` are spread on the component first, then our other props last, otherwise a `props.onClick` function will override our `onClick` prop and things will break.

src/components/top-bar/top-bar.atoms.tsx
```diff
@@ -0,0 +1,15 @@
+import type { FC, HTMLAttributes } from 'react'
+import { ElTopBarMainNav } from './styles'
+
+export const TopBarMainNav: FC<HTMLAttributes<HTMLDivElement>> = ({ children, ...props }) => {
+  return (
+    <ElTopBarMainNav
+      style={{
+        containerType: 'inline-size',
```
kurtdoherty, line 8: **question:** is there any reason this is not part of `ElTopBarMainNav`'s styles?

src/components/top-bar/styles.ts
kurtdoherty: **note:** This is fine for now, but I think we could improve on how we facilitate the unique offset in future; probably when we upgrade to React 19 and get access to the HTML popover API we've mentioned previously.

kurtdoherty (approved):
Thanks @nurm717123 👏

src/components/top-bar/top-bar.atoms.tsx
nurm717123: yes, there is a typing issue currently

image

src/components/menu/menu-popover.tsx
nurm717123: to allow the `TopBar` react consumer to use defined yOffset style like this

image
 
but essentially we just need the classname though, should I update it? 🤔

src/components/menu/menu-popover.tsx
nurm717123: ah yes 🤦 , I will update this, thanks.

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding-left: var(--spacing-6, 24px);
+  flex-grow: 1;
+`
+
+export const ElTopBarSecondaryNav = styled(ElButtonGroup)`
+  flex-wrap: nowrap;
+  padding-right: var(--spacing-spacing-2, 8px);
```
ksolanki7, line 20: @nurm717123 
**issue:** Same issue with the naming convention for the variable.

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  ${isTabletOrBelow} {
+    ${ElTopBarProfile} {
+      display: none;
+    }
+  }
+`
+
+export const ElTopBarSearch = styled.div`
+  padding-right: var(--spacing-spacing-2, 8px);
```
ksolanki7, line 48: @nurm717123 Here, too.

**suggestion:** Please review all variables implemented in this ticket and cross-reference with the token file.

src/components/menu/menu-popover.tsx
kurtdoherty: Ahh, of course, sorry. I saw that addition, but didn't connect it to this change here. This is fine 👍

src/components/top-bar/top-bar.atoms.tsx
kurtdoherty: **todo:** Can you add a comment explaining that then 🙏

src/components/top-bar/styles.ts
kurtdoherty: Ahh, good catch @ksolanki7 🙌

src/components/top-bar/top-bar.atoms.tsx
nurm717123: yes, no problem

nurm717123: After I investigate further apparently its just IDE specific issue as I've tried another editor it doesn't show that warning, I've run yarn check its seems ok, I think I will move the style to the `ElTopBarMainNav` as you suggest

PR #223, opened 2024-11-24
fix: #220 ensure svgr plugin provides `icon` option

### LINKED ISSUE #220: v5 Bug - Icon fontSize property stopped working
**Background context:**

After having upgrades with Vite + Vtest over https://github.com/reapit/elements/issues/159, it introduced a bug with the fontSize property of the component Icon stopped working.

**CodeBase**
`<Icon fontSize="1rem" icon="star"/>`

| Before updates | After updates |
| --- | --- |
| image | image |

The issue seems to be with svgr configuration which replaced from `rollup.config.mjs` to `vite.config.ts`, resulting in not generating CSS for SVG from the icon.

kurtdoherty:
Fixes #220 

#212 missed providing the `icon: true` config to the SVGR plugin when `vite-plugin-svgr` replaced `@svgr/rollup`. This PR adds that config back in to allow icons to be sized correctly.

Screenshot 2024-11-25 at 9 30 24 AM

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -19,7 +19,11 @@ export default defineConfig({
   },
   plugins: [
     react(),
-    svgr(),
+    svgr({
+      svgrOptions: {
+        icon: true,
+      },
+    }),
     wyw({
       babelOptions: {
         presets: ['@babel/preset-typescript', '@babel/preset-react'],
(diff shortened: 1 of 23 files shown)


vite.config.ts
```diff
@@ -19,7 +19,11 @@ export default defineConfig({
   },
   plugins: [
     react(),
-    svgr(),
+    svgr({
+      svgrOptions: {
+        icon: true,
+      },
+    }),
```
kurtdoherty, line 26: **note:** this is the actual fix. The [icon](https://react-svgr.com/docs/options/#icon) option is what were relying on prior to the Vite migration, but the Vite migration work missed this detail.

ksolanki7 (approved):
praise: Looks good to me. Thank you Kurt.

Issue #226, opened 2024-11-28
v5 Button Component Refactor - Housekeeping updates


ksolanki7:
**Background context:**

Several housekeeping updates are needed to enhance the storybook representation of the Button component. Below is the background context for these updates.

- The onClick handler for anchor-type buttons appears non-functional and does not trigger as expected.
- The onClick handler does not correctly handle interactions for a disabled button.
- The icon container CSS has conflicts with overrides for different variants, likely due to a placement issue in the styles file.
- The Button component in the playbook lacks robust support for controls, making it challenging to test the component with various properties and accurately validate the UI.

**Specification or tasks to perform:**

- [x] Resolve the issue with the anchor-type button to ensure the onClick handler functions as expected.
- [x] Update onClick handler for disabled button to avoid calling user's onClick event.
- [x] Address the CSS conflicts in the icon container to correctly handle overrides for different variants.
- [x] Enhance Storybook by improving controls for each Button story, where applicable.

[Link to Component](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-button--docs)

kurtdoherty:
> The Button component in the playbook...

**question:** What is the "playbook"? Do you mean the Button's stories?

ksolanki7:
@kurtdoherty I have updated the issue description.

kurtdoherty:
> The Button component in the playbook lacks robust support for controls, making it challenging to test the component with various properties and accurately validate the UI.

**note:** FWIW, this is what unit tests are for, not storybook (though maybe in future we'll have the two aligned) 😃

ksolanki7:
@kurtdoherty Sure, this is more for the Design team to test the component with different props for UI testing. My approach at the moment is not add support to all stories but at least to default, so that Design team can play with the component.

PR #227, opened 2024-11-29
refactor: #226 v5 Button Component housekeeping updates

ksolanki7:
# Pull request checklist

fixes: Housekeeping issues resolved.

- [x] Resolved the issue with the anchor-type button to ensure the onClick handler functions as expected.
- [x] Updated onClick handler for disabled button to avoid calling user's onClick event.
- [x] Addressed the CSS conflicts in the icon container to correctly handle overrides for different variants.
- [x] Enhanced Storybook by improving controls for each Button story, where applicable.

### DIFF
--- a/src/components/button/button.stories.tsx
+++ b/src/components/button/button.stories.tsx
@@ -42,7 +42,6 @@ export default meta
 export const Default = {
   args: {
     children: 'Button',
-    'aria-label': 'Button aria label',
   },
 }
 
--- a/src/components/button/button.stories.tsx
+++ b/src/components/button/button.stories.tsx
@@ -40,7 +40,10 @@ const meta: Meta<typeof Button> = {
 export default meta
 
 export const Default = {
-  render: ({}) => <Button>Button</Button>,
+  args: {
+    children: 'Button',
+    'aria-label': 'Button aria label',
+  },
 }
 
 export const ButtonVariants = {
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -121,8 +121,7 @@ export const Button: FC<ButtonProps> = ({
         data-variant={variant}
         className={combinedClassName}
         onClick={(e) => {
-          e.preventDefault()
-          handleButtonClick(onClick)
+          handleButtonClick(onClick)(e)
         }}
         aria-label={ariaLabel}
         role="button"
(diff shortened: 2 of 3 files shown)


src/components/button/button.stories.tsx
```diff
@@ -40,7 +40,10 @@ const meta: Meta<typeof Button> = {
 export default meta
 
 export const Default = {
-  render: ({}) => <Button>Button</Button>,
+  args: {
+    children: 'Button',
+    'aria-label': 'Button aria label',
```
kurtdoherty, line 45: **quibble:** I think this could potentially confuse engineers and make them think both need to be supplied 🤔 Without a value for `aria-label`, is that prop not appearing in the Story's controls?

src/components/button/button.tsx
```diff
@@ -141,7 +140,13 @@ export const Button: FC<ButtonProps> = ({
       <ElButton
         data-variant={variant}
         className={combinedClassName}
-        onClick={handleButtonClick(onClick)}
+        onClick={(e) => {
+          if (!disabled) {
+            handleButtonClick(onClick)(e)
```
kurtdoherty, line 145: **quibble:** Can we just do `onClick?.(e)` here instead of the indirection involved with `handleButtonClick`?

kurtdoherty (approved):
Looks good 👍 No blockers, just two quibbles/nit-picks. No need to action them if you want to move on.

src/components/button/button.stories.tsx
ksolanki7: Will update this.

src/components/button/button.tsx
ksolanki7: @kurtdoherty I have centralized the `onClick` logic into a separate function to streamline type-checking and reuse it across both anchor and button atoms. Directly using `onClick?.(e)` is not feasible due to ESLint errors, as `onClick` is explicitly passed as a prop and requires proper type definitions. While it's possible to use an inline onClick handler like this:
```
onClick={(e: MouseEvent<HTMLButtonElement>) => {
  if (!disabled) {
    if (onClick) {
      ;(onClick as MouseEventHandler<HTMLButtonElement>)(e)
    }
  } else {
    e.preventDefault() // Explicitly prevent default if disabled
  }
}}
```
extracting the logic ensures better code reuse and consistency. Please let me know your thoughts on this approach.

kurtdoherty: **note:** As we discussed offline, this is happening because we're not correctly narrowing all the props when we call `isButtonAsButtonElement({ href })` in our main if-statement.

We agreed this can wait until another time to improve 👍

Issue #225, opened 2024-11-28
v5 new component: App Logo (Top Bar)


nurm717123:
**Abstract**

_As part of the v5 Elements release, the ticket #153 was created. To break it into smaller part and allow for more focused discussions, this ticket has been created._

***app bar (parent)***
![image](https://github.com/user-attachments/assets/dbc50742-49e5-4c0d-aebf-2305aa72bfc8)

***this ticket***
![image](https://github.com/user-attachments/assets/a211f99e-cd27-40f2-a57a-7f35adc98717)

**Context**
- the App Logo is a component that would be displayed as part of the Top Bar
- App Logo could differ depending on which app will be using (e.g Reapit, Agentbox, or other Reapit product)
- following [this](https://github.com/reapit/elements/issues/153#issuecomment-2487897635) discussion it would need separate discussion from top bar

As discussed in #153, this ticket is for creating reusable Logo component to be used inside `TopBar` component

**Specification**

- Is new component or existing? new
- Is the component being re-named? No
- Is the component design signed off by product and design? yes, as part of App Bar / Top Bar ticket
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=1564-7620&node-type=symbol&m=dev)
- [Link to Elements Audit in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/App-bar-and-Side-nav-%5BElements%5D?node-id=18-7578&t=AB2erFDHYuLLxvGj-4)
- [Accessibility requirements](https://m3.material.io/components/top-app-bar/accessibility)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-25707&t=mvIlOCZjDYP90TZT-4)
- [Parent ticket](https://github.com/reapit/elements/issues/153)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
[...]

PR #229, opened 2024-12-02
feat: #135 init new top-bar logo

nurm717123:

***app bar (parent)***
image

***this ticket***
image

As discussed in #225, `TopBar.Logo` will only be responsible for the layout styling of the logo. React consumers can use their own logo (or exported `ReapitLogo`) component as a child of `TopBar.Logo`. while CSS user can use `.el-top-bar-logo` classname.

https://github.com/user-attachments/assets/ce765b64-482f-41cd-b3bd-8041c9be86ff


src/components/nav-search-button/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './nav-search-button'
+export * from './styles'
```
nurm717123, line 2: missing index.ts in #196 pr 🤦

src/components/top-bar/styles.ts
```diff
@@ -0,0 +1,5 @@
+import { styled } from '@linaria/react'
+
+export const ElTopBarLogo = styled.a`
+  padding: var(--spacing-spacing-2, 8px) var(--spacing-spacing-2, 8px) var(--spacing-spacing-2, 8px) 0px;
```
ksolanki7, line 4: @nurm717123 
**issue:** There is a naming convention difference between DS figma semantic variables v/s token file variables in the codebase. There is no var called `--spacing-spacing-2` in the codebase, it is `--spacing-2` , 

**suggestion:** Please review all variables and cross-reference with the token file.

nurm717123: as dicussed in the group chat, I will use the `--spacing`, thanks for noticing 👍

src/components/top-bar/top-bar.stories.tsx
```diff
@@ -39,30 +43,27 @@ export const StylesOnlyUsage: StoryObj<typeof TopBar> = {
...
               <ElMenuPopover className={elTopBarMenuPopover}>
                 <ElMenuList>
-                  <ElMenuItemButton role="menuitem">Button 5</ElMenuItemButton>
-                  <ElMenuItemButton role="menuitem">Button 6</ElMenuItemButton>
+                  <ElMenuItemButton role="menuitem">Button 3</ElMenuItemButton>
+                  <ElMenuItemButton role="menuitem">Button 4</ElMenuItemButton>
                   <ElMenuItemAnchor href="/#" role="menuitem">
-                    Button 7
+                    Button 5
                   </ElMenuItemAnchor>
```
nurm717123, line 67: need to reduce the amount of button here, so it looks good in the mdx docs, since mdx docs display the preview in smaller container, while stories has full page.

src/components/top-bar/top-bar.stories.tsx
```diff
@@ -122,15 +123,16 @@ export const ReactUsage: StoryObj<typeof TopBar> = {
...
-          <NavItem href="/">Button 4</NavItem>
           <CSSContainerQuery condition={'(width < 1004px)'}>
+            <NavItem href="/">Button 3</NavItem>
+            <NavItem href="/">Button 4</NavItem>
             <NavItem href="/">Button 5</NavItem>
-            <NavItem href="/">Button 6</NavItem>
-            <NavItem href="/">Button 7</NavItem>
           </CSSContainerQuery>
           <CSSContainerQuery condition={'not (width < 1005px)'}>
             <Menu>
```
nurm717123, line 138: same as above, need to reduce the amount of button here, so it looks good in the mdx docs, since mdx docs display the preview in smaller container, while stories has full page.

.yarnrc.yml
```diff
@@ -4,8 +4,8 @@ enableGlobalCache: false
 
 nodeLinker: node-modules
 
-npmAuthToken: "${NPM_TOKEN}"
+npmAuthToken: '${NPM_TOKEN}'
 
-npmRegistryServer: "https://registry.npmjs.org/"
+npmRegistryServer: 'https://registry.npmjs.org/'
```
kurtdoherty, line 9: **question:** have these been applied automatically by one of the package scripts or was this your IDE formatting this file based on different rules than the repo's prettier config?

src/components/top-bar/top-bar.atoms.tsx
```diff
@@ -0,0 +1,4 @@
+import ReapitLogoIcon from './icons/brand-reapit.svg?react'
```
kurtdoherty, line 1: **todo:** As mentioned earlier in this PR, can you please remove the use of `?react` and instead do:

```javascript
import reapitLogoUrl from './icons/brand-reapit.svg'
import type { ImgHTMLAttributes } from 'react'

interface ReapitLogoProps extends Omit<ImgHTMLAttributes<HTMLImageElement>, 'src' | 'srcSet'> {}

export function ReapitLogo(props: ReapitLogoProps) {
  return (
    <img {...props} src={reapitLogoUrl} />
  )
}
```

**note:** The `img` element may need to be a styled component (i.e. `styled.img`) that bakes in some styles to ensure the img fills it's container while maintaining the correct aspect ratio etc, but hopefully the above gives you a better understanding of what we need this `ReapitLogo` atom to do in order to avoid the `?react` query.

src/components/top-bar/top-bar.stories.tsx
```diff
@@ -25,9 +25,13 @@ import {
   ElTopBarSecondaryNav,
   ElTopBarProfile,
   elTopBarMenuPopover,
+  ElTopBarLogo,
+  elReapitLogo,
 } from './styles'
 import { TopBar } from './top-bar'
 import { figmaDesignUrls } from '#src/storybook/figma/index'
+import ReapitLogoIcon from './icons/brand-reapit.svg?react'
```
kurtdoherty, line 33: **todo:** You'll want to avoid the use of `?react` here as well.

kurtdoherty (commented):
Thanks @nurm717123. Only a few things that need adjustment 👍

src/components/top-bar/styles.ts
```diff
@@ -3,6 +3,17 @@ import { isDesktop, isDesktopOrBelow, isTablet, isTabletOrBelow } from '../../st
 import { ElButtonGroup } from '../button-group'
 import { css } from '@linaria/core'
 
+export const elReapitLogo = css`
+  // need to copy width and height here otherwise it will automatically set to 1em
```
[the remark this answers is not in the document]
nurm717123: yep that is works, thanks 👍  

> todo: Can you avoid using ?react when importing the Reapit SVG and remove this class? 🙏 I've left a comment later in this PR describing how to avoid ?react...

Sure

src/components/top-bar/styles.ts
```diff
@@ -23,7 +34,6 @@ export const ElTopBarSecondaryNav = styled(ElButtonGroup)`
 
 export const ElTopBar = styled.div`
   display: flex;
```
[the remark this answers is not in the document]
nurm717123: Yes, sure. As we're using the margin-right: auto approach to allow the `TopBar.Logo` (or `TopBar.MainNav` on wider screens) to span the available space, therefore the justify content doesn't matter anymore.

src/components/top-bar/top-bar.atoms.tsx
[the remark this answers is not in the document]
nurm717123: I'll extract it

.yarnrc.yml
nurm717123: ah yes, I just had an experiment on my IDE config at that time 🤦 , thanks for noticing 👍

src/components/top-bar/top-bar.atoms.tsx
nurm717123: thanks for the example, I haven't thought to wrap it using `img` previously, its 100% works 👍

.yarnrc.yml
nurm717123: but seems like its the correct one since the defined rules is

```
 quotes: ['error', 'single', { avoidEscape: true, allowTemplateLiterals: false }], // .eslintrc.cjs
  // and
 "singleQuote": true, // .prettierrc.cjs
```

probably this file should be ignored 🤔

src/components/reapit-logo/reapit-logo.tsx
```diff
@@ -0,0 +1,8 @@
+import reapitLogoUrl from './icons/brand-reapit.svg'
+import type { ImgHTMLAttributes } from 'react'
+
+interface ReapitLogoProps extends Omit<ImgHTMLAttributes<HTMLImageElement>, 'src' | 'srcSet'> {}
+
+export const ReapitLogo = (props: ReapitLogoProps) => {
+  return <img {...props} src={reapitLogoUrl} />
```
nurm717123, line 7: seemingly looks ok without additional aspect ratio adjustment, so I think I don't need to make it `styled` for now 🤔

src/components/top-bar/styles.ts
```diff
@@ -3,6 +3,11 @@ import { isDesktop, isDesktopOrBelow, isTablet, isTabletOrBelow } from '../../st
 import { ElButtonGroup } from '../button-group'
 import { css } from '@linaria/core'
 
+export const ElTopBarLogo = styled.a`
+  padding-right: var(--spacing-2, 8px);
+  display: inline-flex;
```
nurm717123, line 8: this `inline-flex` to make the logo vertically centered, `flex` also work though

nurm717123:
> Thanks @nurm717123. Only a few things that need adjustment 👍

Thanks, Kurt
I have updated the PR based on your feedback. Please take a look again 👍 .

kurtdoherty (approved):
Nice work @nurm717123 👏

Issue #228, opened 2024-11-29
v5 component refactor: Skeleton


ksolanki7:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? New
- Is the component being re-named? No
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=2659-2964&node-type=canvas&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

PR #230, opened 2024-12-02
feat: #228 New component skeleton

ksolanki7:
# Description

- Added new component Skeleton
- Only implemented blink(pulse) animation. We can amend new animations with props if DS requires changes.
- It can be used within actual components and can be utilise in `<Suspense />`


kurtdoherty (approved):
**praise:** Nice one @ksolanki7 👏

PR #234, opened 2024-12-04
feat: #218 v5 Table Container and Table Toolbar

ksolanki7:
# Description

**Details of Table atoms completed in this PR:**

- [x] Table Container
- [x] Table Toolbar

**Additional Changes:**
- Fix Skeleton style to support flex display


src/components/skeleton/styles.ts
```diff
@@ -1,7 +1,7 @@
 import { styled } from '@linaria/react'
 
 export const ElSkeleton = styled.span`
-  display: inline-block;
+  display: flex;
```
kurtdoherty, line 4: **question:** what was the motivation behind this change?

src/components/table/tableContainer/__tests__/tableContainer.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { MenuItem, MenuItemGroup, MenuList } from '#src/components/menu/menu.atoms'
+
+describe('Table Container', () => {
+  test('should match snapshot', () => {
+    const { asFragment } = render(
+      <TableContainer>
+        <TableToolbar
+          description="125 Properties"
+          actions={
+            <Menu>
```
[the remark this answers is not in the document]
kurtdoherty: For unit tests, I think it's fine to use fake text for any ReactNode props. Composing in actual atoms like `TableToolbar` and `Menu` will result in our subject's snapshot (in this case, `TableContainer`) capturing the implementation detail of those other components. When those other components are then changed, we'll see test failures here because the snapshot content won't match, which is not something we want.

src/components/table/tableContainer/styles.ts
```diff
@@ -0,0 +1,6 @@
+import { styled } from '@linaria/react'
+
+export const ElTableContainer = styled.div`
+  gap: 0px;
+  background: var(--fill-white);
```
kurtdoherty, line 5: **question:** Is this strictly needed? Do we care what the background of this container is?

src/components/table/tableToolbar/styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  padding: var(--spacing-2) 0px var(--spacing-2) 0px;
+  gap: 0px;
+  justify-content: space-between;
+  align-items: center;
+`
+
+export const ElTableToolbarDescription = styled.div`
+  font-family: var(--font-family);
+  font-size: var(--font-size-sm);
+  font-weight: 400;
```
kurtdoherty, line 16: **issue:** I think we have CSS variables for font weights that we should be using. I can certainly see the variables defined in Figma.

kurtdoherty (approved):
Thanks @ksolanki7, looks good 👏  Two nice and simple components 💪 

**issue:** One main problem I've noticed here is the folder and file naming. You seem to have used camel case where the rest of the repo using snake/hyphen case. Can you please update all the files and folders to be consistent with the rest of the repo? 🙏

src/components/table/tableContainer/styles.ts
ksolanki7: **answer:** I copied this across from Figma, I'll double-check with Andrei if it's a requirement. If not, I will remove it.

src/components/table/tableToolbar/styles.ts
ksolanki7: As discussed, the font-weight variables do not work and need some fixes. I will update it once the fix is in place.

src/components/skeleton/styles.ts
ksolanki7: **answer:** I was facing challenges using `display: inline-block` for the parent element, which did not align the child element in the center. So, I updated the `display` property to flex.

FYI: Figma does not have the `display` property, which I used to support the child element alignment in the parent container.

PR #239, opened 2024-12-06
feat: #235 v5 Tooltip component

ksolanki7:
# Description

- Implemented a new tooltip component
- Descope the conditional display for truncated text to only show the tooltip, this will be addressed in a separate ticket https://github.com/reapit/elements/issues/238

**Demo for Tooltip Story**

https://github.com/user-attachments/assets/0feb336b-99e4-413b-98db-273d47da55a4


src/components/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { render } from '@testing-library/react'
+import { ToolTip } from '../index'
+
+describe('ToolTip', () => {
```
kurtdoherty, line 4: **nit-pick:** Tooltip is one word, not two, so I don't think Tip should be capitalised

src/components/tooltip/styles.ts
```diff
@@ -0,0 +1,132 @@
+import { styled } from '@linaria/react'
+
+export const ElToolTipContainer = styled.div`
```
kurtdoherty, line 3: **nit-pick:** Again, I don't think Tip should be capitalised; i.e. `ElTooltipContainer`. Looks like there's a lot of this throughout the PR.

src/components/tooltip/styles.ts
```diff
@@ -0,0 +1,132 @@
...
+  position: relative;
+  display: inline-block;
+  overflow: visible;
+`
+
+export const ElToolTipChild = styled.div`
```
kurtdoherty, line 9: **suggestion:** I think this should be `ElTooltip` since this represents the actual presentational "tooltip" atom.

src/components/tooltip/styles.ts
```diff
@@ -0,0 +1,132 @@
...
+  padding: var(--spacing-2) var(--spacing-3) var(--spacing-2) var(--spacing-3);
+  border-radius: var(--corner-default);
+  font-family: var(--font-family);
+  font-size: var(--font-size-xs);
+  font-weight: 400; // To do : Need to replace with variable
+  line-height: var(--line-height-xs);
+  letter-spacing: var(--letter-spacing-xs);
+  text-align: left;
+  text-underline-position: from-font;
+  text-decoration-skip-ink: none;
```
kurtdoherty, line 22: **question:** where are these underline and decoration properties coming from? Figma?

src/components/tooltip/tooltip.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        }}
+      >
+        <Story />
+      </div>
+    ),
+  ],
+}
+
+export default meta
+
```
kurtdoherty, line 41: **suggestion:** I think it would be valuable to show an additional story or two for the `ElTooltip` atom directly. That way it's easy to see the visual appearance of the atom without first needing to hover or focus some other element.

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { FC, HTMLAttributes, useState, ReactNode } from 'react'
+import { ElToolTipChild, ElToolTipContainer, ElToolTipLabel } from './styles'
+import { useId } from '#src/storybook/random-id'
+
+export interface ToolTipProps extends HTMLAttributes<HTMLDivElement> {
+  children: ReactNode
+  label?: string
+  tip: string
```
kurtdoherty, line 8: **suggestion:** Figma uses `description` as the prop name. I think that's probably more appropriate as I'd think "tip" includes the label and the description.

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [visible, setVisible] = useState(false)
+  const tooltipId = useId()
+
+  return (
+    <ElToolTipContainer
+      onMouseEnter={() => setVisible(true)}
+      onMouseLeave={() => setVisible(false)}
+      onFocus={() => setVisible(true)}
+      onBlur={() => setVisible(false)}
+      aria-describedby={tooltipId}
```
kurtdoherty, line 40: **issue:** I'm not sure placing `aria-describedby` on the parent is correct. In the case where the `children` is a `<button>`, I believe the `aria-describedby` should be on the `<button>` element itself, not it's parent 🤔 

I suspect we need something a little more akin to `Menu.Trigger`

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <ToolTipChild id={tooltipId} position={position} maxWidth={maxWidth}>
+          {label && <ElToolTipLabel>{label}: </ElToolTipLabel>}
+          {tip}
+        </ToolTipChild>
+      )}
+    </ElToolTipContainer>
+  )
+}
+
+export const ToolTipChild: FC<ToolTipChildProps> = ({ children, position, maxWidth, id }) => {
```
kurtdoherty, line 53: **question:** is the main purpose of this component to handle `maxWidth` being passed in via `style`? The other props seem like they could just be explicitly a part of `ElToolTipChild`'s prop interface and thus avoid the need for this extra layer?

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <ElToolTipContainer
+      onMouseEnter={() => setVisible(true)}
+      onMouseLeave={() => setVisible(false)}
+      onFocus={() => setVisible(true)}
+      onBlur={() => setVisible(false)}
+      aria-describedby={tooltipId}
+    >
+      {children}
+      {visible && (
```
kurtdoherty, line 43: **issue:** We should really be animating the tooltip in and out, so I think the visibility is best handled in CSS via a `data-visible` attribute.

In future, when we upgrade to React 19 (hopefully we can do this soon), we'll get access to the new [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API), which will allow us to push the visibility into the DOM (e.g. https://codesandbox.io/p/sandbox/jp8njk)

kurtdoherty (commented):
Thanks @ksolanki7 👏 This is nicely implemented. Only a few comments, the main one being whether the current API allows us to correctly wire up the a11y props.

src/components/tooltip/styles.ts
ksolanki7: Yes, this is coming from Figma. I'll double-check with Andrei to find a solution for non-required CSS to be in a separate section if not explicitly required for elements.

src/components/tooltip/tooltip.tsx
ksolanki7: @kurtdoherty The rationale behind this approach is to address scenarios where the table contains many items (e.g., 100 rows), each requiring a tooltip. Rendering all tooltip data directly in the DOM upfront avoids the need to dynamically generate content on hover dynamically, ensuring better performance and reducing potential rendering delays.
I am happy to discuss this and get it updated to be handled by CSS.

ksolanki7: @kurtdoherty The rationale behind this approach is to address scenarios where the table contains a large number of items (e.g., 100 rows), each requiring a tooltip. Using conditional rendering ensures better performance and saves memory as we wont' load unnecessary tooltip.

I am happy to discuss and get it updated to be handled by CSS

kurtdoherty: > Using conditional rendering ensures better performance and saves memory as we wont' load unnecessary tooltip.

Have you profiled this? I only ask because I wouldn't have expected there to be much difference at the moment 🤔 If the tooltip was measuring its trigger's bounding box or something like that, then I could see there being a noticeable performance impact, but it seems like a pretty cheap element for browsers to render 🤷‍♂️ 

If performance does become an issue in future, there's CSS properties (e.g. [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility)) that could potentially help us either way.

I don't mind having this go through as is. It may actually be possible we can just as easily animate the entry/exit of these DOM elements via [@starting-style](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style) regardless of whether we're handling visibility in React component state or CSS (I'd forgotten about this new @-rule, though it's not supported by Firefox, so maybe animation needs to wait either way 🤷)

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,49 @@
...
+  maxWidth = '400px',
+  position = 'top',
+  ...rest
+}) => {
+  const tooltip = (
+    <ElTooltip
+      role="tooltip"
+      data-position={position}
+      style={{ maxWidth: maxWidth }}
+      aria-live="assertive" // Announce content dynamically
```
adrian-reapit, line 38: Don't think this is needed, it works ok without it on OSX voiceover. and assertive would normally stop any other screenreader announcement which is not what you want.

src/components/tooltip/tooltip.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+    )
+  },
+  parameters: {
+    docs: {
+      source: {
+        code: `
+render: () => {
+ const tooltip = useTooltip()
+ return (
+    <Button {...tooltip.getTriggerProps()}>Hover me</Button>
```
adrian-reapit, line 93: suggestion: a wrapper component to provide spacing would be useful as the tooltip is not visible 
image

src/components/tooltip/tooltip.tsx
ksolanki7: Thank you @adrian-reapit, will update this.

src/components/tooltip/tooltip.stories.tsx
ksolanki7: **note:** Not adding a wrapper component is to support the tooltip for elements with different CSS `display` properties such as inline, block, etc. The wrapper will have a defined CSS display property which may break the UI. That is the reason, I have used `createPortal` to get the tooltip injected into DOM at the end of the body and positioning it in reference to the trigger element.

kurtdoherty: @ksolanki7 I think Adrian is referring to padding out the _story_ so via a "wrapper element" so that the tooltip is visible when positioned along the top edge of the target.

I think the most appropriate way to handle this would be via a story-specific [decorator](https://storybook.js.org/docs/writing-stories/decorators?renderer=react).

src/components/tooltip/use-tooltip.ts
```diff
@@ -0,0 +1,152 @@
...
+        break
+      case 'right-end':
+        top = triggerRect.bottom + scrollTop - tooltipRect.height
+        left = triggerRect.right + scrollLeft + 4
+        break
+      default:
+        break
+    }
+
+    /* The following calculations ensure proper positioning of the tooltip when it overflows the viewport.
```
kurtdoherty, line 87: **question:** are the following lines of code meant to be commented out?

kurtdoherty (approved):

src/components/tooltip/tooltip.stories.tsx
ksolanki7: @kurtdoherty The storybook looks fine so far and does not require additional positioning in the story. See the below gif.
tooltip

src/components/tooltip/use-tooltip.ts
ksolanki7: @kurtdoherty Yes, these lines are intentionally commented out, as they serve as a reference for future development. These lines pertain to an incomplete solution and have been retained to guide any developers working on future enhancements. I have discussed the dynamic positioning of tooltips—particularly how they should behave when overflowing the viewport - with Andrei. We agreed to defer this feature to a future enhancement, as implementing it effectively will require the establishment of specific UI guidelines to determine the appropriate positioning of tooltips in such cases.

Currently, the consumers of this implementation are internal, and the DS team will be responsible for creating designs that define the proper positioning of the tooltip for elements located at the extremes of the viewport, such as navigation items and similar components.

**Note:** Once the guideline is available over DS, I'll create a backlog ticket for it.

src/components/tooltip/tooltip.stories.tsx
kurtdoherty: What about on the canvas view for that story? Adrian's screenshot shows the "Hover me" button sitting pretty close to the Canvas view's toolbar? 🤷

src/components/tooltip/use-tooltip.ts
kurtdoherty: Personally, I don't think we should have commented out code sitting in our production components when we don't have a timeline of _when_ it'll be used. If you want to keep it, I think there's a few other options:

1. Put that code in a ticket that is tracking the need for us to implement that behaviour in future;
2. Put that code in a comment in this PR so that we can refer to it via this PR if we need to in future.

src/components/tooltip/tooltip.stories.tsx
ksolanki7: @kurtdoherty Got it, let me fix that up.

ksolanki7: Fixed
tooltipCanvas

PR #240, opened 2024-12-06
feat: #153 v5 top bar update docs

nurm717123:
In line with our previous team discussions about standardizing the MDX documentation and avoiding the embedding of Figma guideline SVGs, this pull request addresses that issue. Additionally, this pr also contain small fix regarding the default top-bar menu popover y offset, I think  there was a potential misunderstanding regarding the default Y offset of the menu, specifically:

[previous reference](https://www.figma.com/proto/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?page-id=0%3A1&node-id=1-1876&viewport=1174%2C575%2C0.1&t=oncghAlYrq6zf1jL-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=1%3A1826)

image

which should be [this](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=12-2427&node-type=canvas&m=dev) (notice there is wider gap between nav and menu)

image

thankfully to make that gap size we can utilize `spacing-10` to have the same space 👍 .

# Pull request checklist

- [x] update docs
- [x] update default menu popover y offset  

fixes: 
- part of the #153

fixed result:
- menu gap size
  - figma
  image
  - this pr
  image

- docs

https://github.com/user-attachments/assets/41660af0-5e9d-413b-b4e5-00d9258b08c8


nurm717123:
I have re-updated the top bar docs to include stories for each viewport variant, as suggested by the design team, there is 2 option to go through each viewport variant:

1. using available screen selector at toolbar
2. use stories in sidebar

I keep both version, so let me know if we can't keep both

src/components/top-bar/styles.ts
```diff
@@ -34,7 +34,7 @@ export const ElTopBar = styled.div`
   border-bottom: var(--border-default, 1px) solid var(--outline-default, #e5e9ed);
   background: var(--fill-white, #fff);
 
-  ${isDesktopOrBelow} {
+  ${isBelowWideScreen} {
```
nurm717123, line 37: I initially overlooked this because there are no < or > indicators in the design. However, I now believe this is correct since the only viewport with an indicator is mobile (<768px). This implies that tablet starting at 768px should mean 768px++, and similarly, desktop should be 1024px++, and so on.

src/components/top-bar/styles.ts
```diff
@@ -46,7 +46,7 @@ export const ElTopBar = styled.div`
     }
   }
 
-  ${isTabletOrBelow} {
+  ${isBelowDesktop} {
```
nurm717123, line 49: same as above

> 

I initially overlooked this because there are no < or > indicators in the design. However, I now believe this is correct since the only viewport with an indicator is mobile (<768px). This implies that tablet starting at 768px should mean 768px++, and similarly, desktop should be 1024px++, and so on.

src/components/top-bar/top-bar.stories.tsx
```diff
@@ -1,124 +1,131 @@
...
-  ElTopBarSearch,
-  ElTopBarSecondaryNav,
-  elTopBarMenuPopover,
-} from './styles'
+import { elTopBarMenuPopover } from './styles'
 import { TopBar } from './top-bar'
+import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport'
+console.log(INITIAL_VIEWPORTS)
+
+const viewports: typeof INITIAL_VIEWPORTS = {
```
nurm717123, line 19: need to add additionals because storybook's provided viewport doesn't have widescreen .etc

kurtdoherty (approved):

Issue #249, opened 2024-12-10
v5 component refactor: Token Variables Issue


ksolanki7:
**Background context:**

There is a discrepancy between the `--size-x` variables defined in the token file and their corresponding definitions in Figma. In the token file, the `--size-x` variables are specified in rem units, while in Figma, they are defined in px. Although this difference in units is expected, the conversion between them does not align as intended.

**Example:** 

The variable `--size-9` is defined as `3rem` in the token file, which translates to `48px` (based on the standard conversion). However, in Figma, `--size-9` is set to `36px`, resulting in a discrepancy of `12px`. This inconsistency is observed across the entire set of `--size` variables.

**Specification or tasks to perform:**

- [ ] Update token file to match the figma semantic variables for `size`

Below are the variables with its value from Figma:

| Variable | Value |
| --- | --- |
| `--size-none` | `0rem` |
| `--size-px` | `0.0625rem` |
| `--size-half` | `0.125rem` |
| `--size-1` | `0.25rem` |
| `--size-2` | `0.5rem` |
| `--size-3` | `0.75rem` |
| `--size-4` | `1rem` |
| `--size-5` | `1.25rem` |
| `--size-6` | `1.50rem` |
| `--size-7` | `1.75rem` |
| `--size-8` | `2rem` |
| `--size-9` | `2.25rem` |
| `--size-10` | `2.50rem` |
| `--size-11` | `2.75rem` |
| `--size-12` | `3rem` |
| `--size-14` | `3.5rem` |
| `--size-16` | `4rem` |
| `--size-18` | `4.5rem` |
| `--size-20` | `5rem` |
| `--size-24` | `6rem` |
| `--size-28` | `7rem` |
| `--size-32` | `8rem` |
| `--size-36` | `9rem` |
| `--size-40` | `10rem` |

ksolanki7:
@kurtdoherty Have created ticket as discussed.
FYI: @HollyJoyPhillips

PR #250, opened 2024-12-10
fix: #249 using correct size variables matching DS figma

ksolanki7:
# Description

Amendment made to token file for supporting correct `size` variables matching DS figma

**Additional Changes**

- Updated the token build file to use the correct file path
- Updated existing usage of `size` variable with correct values matching DS Figma


src/tokens/build.mjs
```diff
@@ -27,7 +27,7 @@ const getStyleDictionaryConfig = (theme) => {
     ],
     platforms: {
       css: {
-        buildPath: `tokens/${lowerCasedTheme}/`,
+        buildPath: `src/tokens/${lowerCasedTheme}/`,
```
ksolanki7, line 30: @kurtdoherty The Export path also required an update to the export token (CSS/ts) path.

kurtdoherty (approved):

Issue #203, opened 2024-11-12
v5 new component: AvatarButton (App-bar)


nurm717123:
**Abstract**

_As part of the v5 Elements release, the ticket #153 was created. To break it into smaller part and allow for more focused discussions, this ticket has been created._

***app bar (parent)***
![image](https://github.com/user-attachments/assets/dbc50742-49e5-4c0d-aebf-2305aa72bfc8)

***this ticket***
![image](https://github.com/user-attachments/assets/c015db6b-62c2-4e41-b5e6-adddc6767af1)

the component should extends the new Avatar component https://github.com/reapit/elements/issues/137

**Specification**

- Is new component or existing? new
- Is the component being re-named? no
- Is the component design signed off by product and design? yes, as part of the new app bar
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=52-879&node-type=symbol&m=dev)
- [Link to Elements Audit in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4808&node-type=instance&m=dev)
- [Accessibility requirements](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)
- [Parent ticket](https://github.com/reapit/elements/issues/153)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

kurtdoherty:
@nurm717123 can you please outline the proposed interface for this component _before_ you submit any work for review? 🙏 The whole point behind a "design doc" (even a simple "here's the component's prop interface") is to get agreement before we invest in writing code.

nurm717123:
[...]

PR #251, opened 2024-12-13
chore: #137 fix css var name and remove fallback value (Avatar)

nurm717123:
In line with our previous discussion, it is recommended to remove the CSS variable fallback. This will allow us to identify any invalid tokens early on. This pull request also includes fixes to variable names to ensure they are correctly typed.

# Pull request checklist

- [x] remove css fallback
- [x] fix css var name

fixes: #137, #203


ksolanki7 (approved):
praise: Looks good to me.

PR #254, opened 2024-12-16
fix: #163 fix menu "show code" doesn't show expected code

nurm717123:
The Menu mdx docs doesn't show the expected code, its apparently because storybook doesn't support to add controls for react compound component, the fix is to remove the controls and the controls will be represented by available stories instead as the previous controls is just boolean.

previous:

image

fixed:

image

fixes: #163

new docs:

menudocs


ksolanki7 (approved):
**praise:** Looks good.

PR #260, opened 2024-12-17
chore: 248 v5 component refactor button refactor UI

ksolanki7:
# Description

- Replaced vertical padding with height
- Introduced new prop `removePadding` to support the compact view of a Button component with no height, width or padding.
- Updated controls for Icons to be selectable from a list of sample icons
- Made a minor update to the TypeScript definition for the `onClick` prop to type it for Button as Button and Button as Anchor elements explicitly.


src/components/button/button.tsx
```diff
@@ -42,38 +42,51 @@ interface CommonButtonProps {
...
 
-interface ButtonAsButtonElementProps extends CommonButtonProps, ButtonHTMLAttributes<HTMLButtonElement> {
-  href?: never
-  target?: never
-  disabled?: boolean
-  onClick?: MouseEventHandler<HTMLButtonElement>
+// Define a specialized type for tertiary variant for prop removePadding
+type TertiaryButtonProps = {
+  variant: 'tertiary'
+  removePadding?: boolean // Only tertiary buttons can use removePadding
```
ksolanki7, line 48: The reason for naming this prop `removePadding` is confirmed with Andrei after a discussion with him.

src/components/button/button.tsx
```diff
@@ -42,38 +42,51 @@ interface CommonButtonProps {
...
 }
 
-interface ButtonAsButtonElementProps extends CommonButtonProps, ButtonHTMLAttributes<HTMLButtonElement> {
-  href?: never
-  target?: never
-  disabled?: boolean
-  onClick?: MouseEventHandler<HTMLButtonElement>
+// Define a specialized type for tertiary variant for prop removePadding
+type TertiaryButtonProps = {
+  variant: 'tertiary'
```
ksolanki7, line 47: `removePadding` is only available for variant `tertiary`

kurtdoherty: **nit-pick:** per the [(draft) boolean prop interface guidelines](https://consolegroup.atlassian.net/wiki/spaces/DP/pages/3127509024/Prop+Interface+Guidelines#On-Booleans) you have access to, I think `hasNoPadding` would be preferred; `removePadding` reads like an action, not a state.

src/components/button/button.tsx
```diff
@@ -42,38 +34,51 @@ interface CommonButtonProps {
...
+    variant?: ButtonAsAnchorVariant
+    /** Button styled <a> element should always have href */
+    href: string
+    target?: string
+    rel?: string
+    /** Anchor elements cannot be disabled. Use a button element if the component needs to be in a disabled state */
+    disabled?: never
+    onClick?: MouseEventHandler<HTMLAnchorElement>
+  }
+
```
kurtdoherty, line 73: **note:** I don't love how complex these types are becoming. Not really sure what our options are to simplify them, just flagging that this level of complexity can lead to poor DX as type errors become hard to read 🤔 

Intersections are also not the best way of combining types. See the [(draft) prop interface guidelines](https://consolegroup.atlassian.net/wiki/spaces/DP/pages/3127509024/Prop+Interface+Guidelines#Prefer-interface-over-intersections-when-composing-types) for details on why interface/extends is preferred over intersections. No need to rewrite these types yet; rather I think we should do a sweep early next year to tidy up some of our component prop types (there's TS config changes I'd like to rollout, along with ES Lint upgrades etc).

src/components/button/styles.ts
```diff
@@ -10,6 +10,7 @@ export const elButtonLabel = css``
 export const elIcon = css``
 export const elButtonIconOnly = css``
 export const elButtonSpinner = css``
+export const elButtonRemovePadding = css``
```
kurtdoherty, line 13: **suggestion:** I think a data attribute is a nicer, more "componenty" approach than class names. For example, if the React component prop is `hasNoPadding` then the styled component here would accept `data-has-no-padding` and the button styles would use `&[data-has-no-padding="true"] { ... }`

kurtdoherty (approved):
Nice one @ksolanki7 👏 Thanks for doing this work. Just a few minor comments; no blockers.

src/components/button/styles.ts
ksolanki7: Thanks for the reminder, I always forget this one.

Issue #253, opened 2024-12-13
v5 component refactor: existing App Switcher compatibility (Top Bar)


nurm717123:
**Abstract**

_As part of the v5 Elements release, the ticket #153 was created. To break it into smaller part and allow for more focused discussions, this ticket has been created._

***app bar (parent)***
![image](https://github.com/user-attachments/assets/dbc50742-49e5-4c0d-aebf-2305aa72bfc8)

***this ticket***
![image](https://github.com/user-attachments/assets/abea3ff7-d037-4b0f-a890-8729a8de7980)

![image](https://github.com/user-attachments/assets/7c5e8c46-a585-48be-acd2-6f5306aa4dcc)

**Context**
- new app switcher not ready for dev yet
- while #222 will become the ticket for new AppSwitcher once the design is ready, this one is to use existing AppSwitcher in the new TopBar

As discussed in #153, this ticket involves creating a new app switcher based on the existing app switcher (`NavResponsiveAppSwitcher`) for the initial implementation.

**Specification**

- Is new component or existing? existing
- Is the component being re-named? No
- Is the component design signed off by product and design? not yet, but become part of App Bar / Top Bar ticket
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4840&m=dev)
- [Link to Elements Audit in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=16-4840&m=dev)
- [Accessibility requirements](https://m3.material.io/components/top-app-bar/accessibility)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-25707&t=mvIlOCZjDYP90TZT-4)
- [Parent ticket](https://github.com/reapit/elements/issues/153)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
[...]

PR #261, opened 2024-12-17
feat: #253 new app-bar (existing app switcher)

nurm717123:

***app bar (parent)***
image

***this ticket***
image

image

note: this component is just a wrapper to use existing app-switcher in the new `TopBar` 

top-bar appswitcher

### DIFF
--- a/src/components/top-bar/styles.ts
+++ b/src/components/top-bar/styles.ts
@@ -2,6 +2,22 @@ import { styled } from '@linaria/react'
 import { isBelowWideScreen, isTablet, isBelowDesktop, isWideScreen } from '../../styles/media'
 import { ElButtonGroup } from '../button-group'
 import { css } from '@linaria/core'
+import { ElNavResponsiveAppSwitcherIconWrap } from '../nav'
+
+export const ElTopBarAppSwitcher = styled.a`
+  padding-right: var(--spacing-4);
+
+  ${isWideScreen} {
+    padding-right: var(--spacing-5);
+  }
+
+  // TODO: can be removed once new app-switcher ready
+  ${ElNavResponsiveAppSwitcherIconWrap} {
+    margin-right: 0;
+    padding: var(--spacing-half);
+    box-sizing: content-box;
+  }
+`
 
 export const ElTopBarLogo = styled.a`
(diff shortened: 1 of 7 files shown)


src/components/top-bar/styles.ts
```diff
@@ -2,6 +2,22 @@ import { styled } from '@linaria/react'
...
+
+  ${isWideScreen} {
+    padding-right: var(--spacing-5);
+  }
+
+  // TODO: can be removed once new app-switcher ready
+  ${ElNavResponsiveAppSwitcherIconWrap} {
+    margin-right: 0;
+    padding: var(--spacing-half);
+    box-sizing: content-box;
```
nurm717123, line 18: so that the additional padding doesn't reduce the size of the component

kurtdoherty: **question:** @nurm717123 can you post some screenshots that show the problem here?

**suggestion:** I'd prefer we don't add mess to our TopBar just to get a deprecated component looking good. Can we push these styles into `ElNavResponsiveAppSwitcherIconWrap` via an optional prop or something? That way the mess lives with the component that owns it 🤔

kurtdoherty: Thanks for the screenshots 🙏  I think your suggestion to provide a class exported from `NavResponsiveAppSwitcher` that consumers can manually apply is fine, though I don't see how it's any harder to add a prop that just applies the extra class instead of making `className` available and expect consumers to use the new class. Either way though, we're keeping the mess inside the old component, which is great 👍

kurtdoherty (approved):
Thanks @nurm717123 👏 I'm approving on the assumption you'll move those custom styles inside `NavResponsiveAppSwitcher` somehow. Everything else looks fine 👍

nurm717123:
> Thanks @nurm717123 👏 I'm approving on the assumption you'll move those custom styles inside `NavResponsiveAppSwitcher` somehow. Everything else looks fine 👍

Thanks, I will do that 👍

src/components/top-bar/styles.ts
nurm717123: its not that hard actually, I think its easier because that way I don't have to add unit test 😅

Issue #236, opened 2024-12-04
v5 new component: MobileNavMenu


ss-dimasm:
**Abstract**

Now that the `MobileNavItem` component is live in the primary branch, we can focus on developing the `MobileNavMenu` component. This ticket outlines the approach we'll take

![{C3C51760-7848-4A09-803E-ADD029AC92B6}](https://github.com/user-attachments/assets/f37216b6-8d9f-47fc-bdee-74048fa786ee)

**Specification**

- Is new component or existing? new
- Is the component being re-named? no
- Is the component design signed off by product and design? yes
- [Guide to component structure & documentation in storybook](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=18-6292&node-type=instance&m=dev)
- [Accessibility requirements](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1661-8748&node-type=instance&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

ss-dimasm:
Hi @kurtdoherty and @adrian-reapit, I’ll gather some initial information on how this component should be structured. Creating an early POC is necessary, and I’ll return to this conversation once it’s complete, thanks 👍

kurtdoherty:
@ss-dimasm I don't think we're really in a position to start this work yet. The `MobileNavMenu` will need to use new `Drawer` components which aren't implemented yet. Further, I'm probably best positioned to implement these Drawer components as I have a lot of context from recent work done with them in the Console Cloud product.

ss-dimasm:
[...]

Issue #154, opened 2024-08-15
v5 component refactor: Sidebar


HollyJoyPhillips:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing 'Secondary Nav' 
- Is the component being re-named? Renaming to 'Sidebar' 
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1661-6496&t=mvIlOCZjDYP90TZT-4)
- [Link to Elements Audit in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/App-bar-and-Side-nav-%5BElements%5D?node-id=34-6745&t=AB2erFDHYuLLxvGj-4)
- [Accessibility information](https://www.patternfly.org/components/sidebar/accessibility/)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1661-6496&t=1ZcE8H9xiqZy1REJ-4)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

nurm717123:
Hi @kurtdoherty and @adrian-reapit , while the `TopBar` is nearly complete, I would like to initiate the discussion in preparation to complete this ticket;

### Context
- in term of structure Sidebar is almost similar with the Mobile Nav, the difference is Sidebar has icon and able to collapse or expand
[...]

PR #264, opened 2024-12-26
refactor: #205 (mobile-nav-item) listen isExpanded state to outer world argument

ss-dimasm:
# Pull request checklist

part of  #206 

no UI changed, but only how this component should behave when facing with outer world integration

context: https://github.com/reapit/elements/issues/236#issuecomment-2550558575 

The new behavior would resolves upper-level integration issues (like React Router) by avoiding internal state management. The parent component defines the source of truth for expanded/collapsed states. This ensures the component follows the defined guidance. This will allow the upper config to immediately close another item group once another page from another group mobile nav menu independently of click listeners or external events that modify the configuration's active state

- The `isActive` prop is required when using the expandable version
- Add an additional `onClick` prop to the expandable component, which will refer to the button constraint area
- Remove the isolated state previously defined within the expandable component
- Remove the HTML Storybook markup from the mobile nav item expandable story


src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
kurtdoherty, line 1: **question:** I'm not sure this is entirely what we want. Ideally, we'd want the expandable item to expand/collapse when clicked without forcing consumers to manually control that state at all times.

Nur has implemented this behaviour in a POC linked from [this comment](https://github.com/reapit/elements/issues/154#issuecomment-2581611128) related to the SideBar component. You probably don't have to do quite as much as what Nur's doing in that POC (the SideBar has the added complexity of having the whole menu collapse to a smaller size), but it demonstrates how we can avoid forcing consumers to manage open/closed state for these groups up high.

As an aside, another issue with hoisting the state up to the consumer like this is that, any time we expand/collapse a group, the whole menu of items will re-render. This is another motivation for us to try and keep the _manual_ open/closed_ state inside the group.

To summarise, we have two values at play with our groups:
1. `isActive`: supplied by the consumer. When true, this means one of the child items is the current page, so the group should be expanded;
2. `isExpanded`: internal to the expandable group. Toggles between true and false when the group header item is clicked.

One way of expressing this in a React component is:

```javascript
function MyExpandableGroup({ isActive = false }) {
  // We set up our internal expanded/collapsed state and initialise it
  // to match whatever isActive is set to.
  const [isExpanded, setIsExpanded] = useState(isActive)

  const expandOrCollapse = () => {
    // When we're deciding whether to expand or collapse, if isActive
    // is true, we want isExpanded to be true. If isActive is false, we want
    // isExpanded to logically invert.
    setIsExpanded(isExpanded => isActive || !isExpanded)
  }

  // When isActive becomes true, we want isExpanded to become true.
  // When isActive becomes false, we want isExpanded to become false.
  useEffect(function syncIsActiveWithIsExpanded() {
    setIsExpanded(isActive)
  }, [isActive]

  return (
    <ElMobileNavItemListItem {...rest} data-is-expanded={isExpanded} aria-label={label}>
      <ElMobileNavItemExpanderButton
        type="button"
        aria-expanded={isExpanded}
        aria-controls={panelId}
        onClick={expandOrCollapse}
      >
        // ...
      </ElMobileNavItemExpanderButton>
    </ElMobileNavItemListItem>
  )
}
```

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
```diff
@@ -74,33 +73,33 @@ export const MobileNavItemSimple: FC<MobileNavItemSimpleProps> = (props) => {
...
+  const { isActive, label, hasBadge, children, onClick, ...rest } = props ?? {}
 
   return (
-    <ElMobileNavItemListItem {...rest} data-is-expanded={isExpanded} aria-label={label}>
+    <ElMobileNavItemListItem {...rest} data-is-expanded={isActive} aria-label={label}>
       <ElMobileNavItemExpanderButton
         type="button"
-        aria-expanded={isExpanded ?? isActive}
+        aria-expanded={isActive}
         aria-current={isActive ? 'true' : undefined}
```
kurtdoherty, line 90: **suggestion:** I don't think `aria-current` belongs on the "expander button" as it does not actually represent the current page; the current page is represented by on of the child items.

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
kurtdoherty: @nurm717123 it's possible the approach I show above could simplify some of your POC for the sidebar. It's not a big deal if you follow it or not; just letting you know its a possibility.

ss-dimasm: Hey @kurtdoherty thanks for noticing this, it's been a while and I think the problem that we would fix at that time is to make the state aligned when the url is changed (assuming it's not dispatching the mouse event in the mobile nav menu)

however by looking at your suggestion it would also work and be nicer since we have 2 event listeners there, but that would still make the whole expandable item rerender right? I am guessing the second listener will always be reactive based on 'isActive' props, it will keep updating the internal state in item expandable when that prop changed?

another advantage of using this suggestion would be a more decentralized distribution of responsibility based on the declared variables, so it would make more clear for further refactor work from the maintainer side

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
ss-dimasm: this is right, the expander button doesn't need this aria rule

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
```diff
@@ -73,24 +73,28 @@ export const MobileNavItemSimple: FC<MobileNavItemSimpleProps> = (props) => {
...
+  const [isExpanded, setIsExpanded] = useIsMobileNavItemExpanded(Boolean(isActive))
   const panelId = useId()
 
-  const { isActive, label, hasBadge, children, ...rest } = props ?? {}
+  const handleOnExpandButtonClick = useCallback(() => {
+    setIsExpanded((prev) => !prev)
+  }, [])
 
   return (
     <ElMobileNavItemListItem {...rest} data-is-expanded={isExpanded} aria-label={label}>
```
adrian-reapit, line 92: this aria-label is not needed. It repeats the content which will be read by a screen reader twice on the li and a tags

image

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
```diff
@@ -73,24 +73,28 @@ export const MobileNavItemSimple: FC<MobileNavItemSimpleProps> = (props) => {
...
         type="button"
-        aria-expanded={isExpanded ?? isActive}
-        aria-current={isActive ? 'true' : undefined}
+        aria-expanded={isExpanded}
         aria-controls={panelId}
-        onClick={() => setIsExpanded((prev) => !prev)}
+        onClick={handleOnExpandButtonClick}
       >
         <ElMobileNavItemContent>
           {label}
```
adrian-reapit, line 100: The Icon aria label is too long and not helpful. 

image

Can add aria-hidden and then its much easier to understand
image

src/components/mobile-nav-item/mobile-nav-item.stories.tsx
```diff
@@ -31,13 +33,19 @@ export const WithBadge: Story = {
...
+        <Button onClick={() => setIsOpen((prev) => !prev)}>{isOpen ? 'Close' : 'Open'}</Button>
+        <MobileNavItem label="Item 2" hasBadge isActive={isOpen}>
+          <MobileNavItem label="Sub Item 2.1" href="#item-2.1" />
+          <MobileNavItem label="Sub Item 2.2" onClick={action('onClick')} hasBadge />
+          <MobileNavItem label="Sub Item 2.3" href="#item-2.3" />
+        </MobileNavItem>
+      </>
+    )
+  },
 }
```
adrian-reapit, line 50: it would be useful to have a story showing a combination, then its easier to test
** <ul>
        <MobileNavItem label="Click Me" href="#item-1" />
        <MobileNavItem label="Active Link" href="#item-1" isActive />
        <MobileNavItem label="More items">
          <MobileNavItem label="Sub Link 1" href="#item-3.1" />
          <MobileNavItem label="Sub Link 2" href="#item-3.3" />
          <MobileNavItem label="Action" onClick={action('onClick')} />
        </MobileNavItem>
      </ul>**

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
```diff
@@ -73,24 +73,28 @@ export const MobileNavItemSimple: FC<MobileNavItemSimpleProps> = (props) => {
...
 
 export const MobileNavItemExpandable: FC<MobileNavItemExpandableProps> = (props) => {
-  const [isExpanded, setIsExpanded] = useState(Boolean(props?.isActive))
+  const { isActive, label, hasBadge, children, ...rest } = props ?? {}
+
+  const [isExpanded, setIsExpanded] = useIsMobileNavItemExpanded(Boolean(isActive))
   const panelId = useId()
 
-  const { isActive, label, hasBadge, children, ...rest } = props ?? {}
+  const handleOnExpandButtonClick = useCallback(() => {
```
adrian-reapit, line 87: shouldnt need usecallback here, its a basic function

src/components/mobile-nav-item/use-is-mobile-nav-item-expanded.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react'
+
+export const useIsMobileNavItemExpanded = (isActive: boolean) => {
+  const [isExpanded, setIsExpanded] = useState(isActive)
+
+  // NOTE: make the internal state to always tailor to the external isActive state
+  useEffect(
+    function syncIsActiveWithIsExpanded() {
```
adrian-reapit, line 8: doesnt need function name explanation, just simple is ok

  useEffect(() => {
    setIsExpanded(isActive)
  }, [isActive])

src/components/mobile-nav-item/mobile-nav-item.atoms.tsx
ss-dimasm: Hi adrian, good catch on this

the long text aria label is coming from the Icon chevron, but we could move down the aria-label in ul to the button itself; so it's referenced correctly from narator screen reader

adrian-reapit: The label is not needed in aria label, the text will be read out from the button content. Can allow an aria-label prop to be passed through to the button and give an example in stories. Can use it to give more meaning the button's purpose.  **aria-label='Sub menu of ....'**     Could have a default using the button label as the last word but it might not always make sense so let user control it

The icon title is not helpful, most people don't know what a chevron is and aria-expanded is doing the same job for screenreader user as the chevron is doing visually.

ss-dimasm: Hi Adrian, we could achieve that with 2 approaches

1. add aria-label in button and that value would be same like the `label` 
2. add aria-hidden in Icon component

the latest upstream branch already implement the first approach

as an example with the latest changes in this PR

image

Narrator would say "Contact Button Expanded" / "Contact Button Collapsed", "Link Create", "Report Button", "Link Archive" respectively based on the highlighted node

ss-dimasm:
outdated, have raise another PR #387

PR #265, opened 2024-12-26
feat:#243 add features component

ss-dimasm:
# Pull request checklist

part of #243 

## Context
Initialize a new component as part of elements v5 'Features

**Design:**

image

**Preview (with Tooltip support):**

https://github.com/user-attachments/assets/6b559a5e-09b3-4aa8-ace0-e60d2fe5b1c1


ksolanki7:
@ss-dimasm The issue you discussed with me is maybe the race condition on the visibility of the Tooltip v/s the placement of the tooltip on the canvas. The solution to that can be adding data attribute to tooltip by updating the useTooltip hook -> getTooltipProps i.e. `data-visible` to boolean `isVisible` state and then managing the animation using CSS as below.

```
export const ElTooltip = styled.div`
  @keyframes fadeIn {
    from {
      opacity: 0;
    }
    to {
      opacity: 1;
    }
  }

  @keyframes fadeOut {
    from {
      opacity: 1;
    }
    to {
      opacity: 0;
    }
  }

  ... (rest of CSS)

  animation-duration: 0.2s;
  animation-fill-mode: both;
  animation-timing-function: ease-in-out;
  opacity: 0;

  &[data-visible='true'] {
    animation-name: fadeIn;
  }
  &[data-visible='false'] {
    animation-name: fadeOut;
  }
```

ss-dimasm:
Hey @ksolanki7 while I agree it's caused by the race condition that appear between javascript logic and the canvas, and we would fix it with CSS solution

I suppose it would be more handy if it have their own isolated ticket; hence it's not only appearing in the Features and I believe as far as it not changing the API it would not cause a major change, what do you say?

ksolanki7:
@ss-dimasm Yes, definately. This requires changes in the Tooltip itself i.e. at the component level and not at the consumer end.

src/components/features/features.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+    </Features>
+  ),
+}
+
+export const DisplayTooltip = {
+  render: ({}) => {
+    return (
+      <Features>
+        <ElFeaturesItem aria-label="Bathrooms">
+          <ElTooltip
```
kurtdoherty, line 51: **question:** why are you using `ElTooltop` and not `Tooltip`?

ss-dimasm: >  why are you using ElTooltop and not Tooltip?

the main reason is to provide a preview of the Tooltip and Feature integration without need a hover action from end user

kurtdoherty: Ahh okay. It's a bit awkward that we have to do it this way; makes me think we should add a prop to Tooltip that forces it to be visible 🤔 Anyway, all good.

**suggestion:** So we don't confuse devs that look at the story's code, can you add a code comment here highlighting that you're only using `ElTooltip` because we want the tooltip permanently visible in the story? 🙏

kurtdoherty (approved):
Thanks @ss-dimasm. Just the one todo off the back of your answer.

Issue #246, opened 2024-12-09
v5 component refactor: Empty Data


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? New
- Is the component being re-named? N/a
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=1602-7392&node-type=canvas&m=dev)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=3234-46308&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

nurm717123:
Hi @kurtdoherty , cc @ksolanki7 

I've made a pr to initiate the work for this ticket ([here](https://github.com/reapit/elements/pull/267)), I think it just need one more discussion regarding the updated design spec i.e "The height (of the EmptyData) can be either fixed or it can be fluid."

currently it is fixed (using `var(--size-40)`), and to support the fluid version, I think there is 2 approach:
1. provide a `boolean` props to control it (e.g `hasFullHeight`, `hasFluidHeight`, `isFullHeight`, .etc)
2. let user custom it via style or className e.g `<EmptyData className={myCustomStyle}>...</EmptyData>`

I think I prefer the 1st approach, what do you think? 🤔

kurtdoherty:
[...]

PR #267, opened 2024-12-27
feat: #246 new v5 EmptyData

nurm717123:
This pr is intended to initiate a creation of a small component called `EmptyData`

image

there is 3 version, but as the component is built with a composite pattern just like `Menu`, `Avatar` .etc, the consumer simply just need to use the composition that they need.

# Pull request checklist

- [x] Default version
- [x] Description only
- [x] Action only

Result:

https://github.com/user-attachments/assets/ada0e19c-ca8c-47df-83a0-77df3adf9fd5


src/components/empty-data/empty-data.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * There are three variants available for the EmptyData component:
+ * Default (full version), Description Only, and Action Only.
+ * See the examples below for more details of the usage.
+ */
+export const Default: Story = {
+  render() {
+    return (
+      <EmptyData>
+        <EmptyData.Description>No things found</EmptyData.Description>
```
adrian-reapit, line 23: Can we use 'items' as a friendlier word than 'things' ;)

src/components/empty-data/styles.ts
```diff
@@ -0,0 +1,49 @@
+import { styled } from '@linaria/react'
+
+export const ElEmptyDataDescription = styled.div`
+  color: var(text-primary);
```
adrian-reapit, line 4: this shows an error in inspector

image

src/components/empty-data/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+  border: none;
+  background: none;
+  cursor: pointer;
+`
+
+export const ElEmptyData = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
```
adrian-reapit, line 35: I think it needs text-align center as well, the centre alignment breaks on narrow widths which could occur if this was used inside a nested component on mobile

image

src/components/empty-data/styles.ts
```diff
@@ -0,0 +1,49 @@
+import { styled } from '@linaria/react'
+
+export const ElEmptyDataDescription = styled.div`
```
adrian-reapit, line 3: This is a heading, can it be ElEmptyDataHeading and extend a heading instead of div, maybe h4 as its so small

src/components/empty-data/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+  color: var(text-primary);
+  font-size: var(--font-size-base);
+  font-weight: var(--font-weight-regular);
+  line-height: var(--line-height-base);
+  letter-spacing: var(--letter-spacing-base);
+`
+
+export const ElEmptyDataSecondaryDescription = styled.div`
```
adrian-reapit, line 11: Can make this ElEmptyDataContent and it can extend a p tag. Can use more semantic markup than just divs

image

src/components/empty-data/styles.ts
nurm717123: my bad, I missed it 🤦

src/components/empty-data/empty-data.stories.tsx
nurm717123: Ah, I just follow the figma's example, I thought its ok to use that 😅, I'll update it anyway 👍

src/components/empty-data/styles.ts
nurm717123: Ah I see, haven't thought about that, I got it 👍

src/components/empty-data/styles.ts
nurm717123: yes, I think this is the part that need to be discussed, because if we rename it `ElEmptyDataHeading` we also need to rename `ElEmptyDataSecondaryDescription`

image

should it be `ElEmptyDataDescriptionHeading` and `ElEmptyDataDescriptionSubHeading` ?

src/components/empty-data/styles.ts
nurm717123: I've asked to Andrei [here](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=1663-23346&m=dev#1095630233)

kurtdoherty (approved):
Nice one @nurm717123 👏 Looks good to me, just the one thought/suggestion around Adrian's feedback + a note that I've asked Andrei to clarify whether the Action is really a component of it's own (like a generic `Link` component or something).

I don't think you need to wait for clarity on my question to Andrei before merging though.

src/components/empty-data/styles.ts
kurtdoherty: > In a notification type component...
This isn't a notification component though, right? It's just some static text on the page...
If we want the visual emphasis to be communicated via the accessibility tree, I think `<strong>` would be closer to the mark, but even then, the semantics of `<strong>` don't really seem to relate to this "no results" text to me 🤷 

> The faint second line fails on contrast ratio against the background colour
That needs to be fed back to design; they're the ones in control of the colour palette.

nurm717123: after I research a bit, actually there is some example i.e in github or jira that use `heading` for that

image
image

but also there is some that use a `div` e.g gmail and outlook.

I didn't find the solid reasoning if heading is really required here yet, so I think I will make it as `p` first and then we can change it later if needed 🙏

PR #269, opened 2024-12-27
chore: #241 deprecate current Tag component

nurm717123:
## Context

As there is new v5 tag #241  that will replace the previous (v4) tag component.

this pr is just the work to deprecate the previous version.

# Pull request checklist

**Detail as per issue below (required):**

fixes: #

Issue #241, opened 2024-12-09
v5 component refactor: Tag/Tag Group


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing
- Is the component being re-named? No
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=1282-7386&m=dev)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/tag--docs)
- [Accessibility requirements](https://carbondesignsystem.com/components/tag/accessibility/#accessibility-testing-statusobject-objectobject-object)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

nurm717123:
Hi @kurtdoherty , cc @adrian-reapit @ksolanki7 

following the discussion [here](https://github.com/reapit/elements/pull/270#issuecomment-2597555809), I think its need a follow up discussion about how we going to approach rendering single tag and groups of tag.

### Context
- Tag can either rendered as `standalone` or in group
- Tag component need to know how it will be used in order to set appropriate role
- current implementation is using `isStandalone` (default=false) to approach it

### Question
- What approach is best to ensure that developers can intuitively understand how to use the component?

in this case I think there is 3 way, I can think of

1. use `isStandalone` prop (current approach)
```javascript
<Tag isStandalone>bar</Tag> {/* for single tag */}
<TagGroup>
  <Tag>bar 1</Tag>
  <Tag>bar 2</Tag>
  <Tag>bar 2</Tag>
</TagGroup>
```
[...]

PR #270, opened 2024-12-27
feat: #241 init v5 new Tag component

nurm717123:
## Context
There is new v5 tag #241 that will replace the previous (v4) tag component.

_note: #269 need to be merged first in order deprecate older component_

this pr is to create the new component to replace the previous one

# Pull request checklist

[x] new tag component

result preview:

https://github.com/user-attachments/assets/cad5ef0d-c4d9-4463-8223-683cb20d543f

### DIFF
--- a/src/components/tag/__tests__/__snapshots__/tag.test.tsx.snap
+++ b/src/components/tag/__tests__/__snapshots__/tag.test.tsx.snap
@@ -2,14 +2,14 @@
 
 exports[`Tag > should render properly and match snapshot 1`] = `
 <DocumentFragment>
-  <div
+  <ul
     class="mocked-styled-1 el-tag-group"
   >
-    <div
+    <li
       class="mocked-styled-0 el-tag"
     >
       Tag
-    </div>
-  </div>
+    </li>
+  </ul>
 </DocumentFragment>
 `;
--- a/src/components/tag/__tests__/__snapshots__/tag.test.tsx.snap
+++ b/src/components/tag/__tests__/__snapshots__/tag.test.tsx.snap
@@ -1,0 +1,15 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Tag > should render properly and match snapshot 1`] = `
+<DocumentFragment>
+  <div
+    class="mocked-styled-1 el-tag-group"
+  >
+    <div
+      class="mocked-styled-0 el-tag"
+    >
+      Tag
+    </div>
+  </div>
+</DocumentFragment>
+`;
--- a/src/components/tag/__tests__/tag.test.tsx
+++ b/src/components/tag/__tests__/tag.test.tsx
@@ -1,0 +1,13 @@
+import { render } from '@testing-library/react'
+import { Tag, TagGroup } from '../tag'
+
+describe('Tag', () => {
+  it('should render properly and match snapshot', () => {
+    const { asFragment } = render(
+      <TagGroup>
+        <Tag>Tag</Tag>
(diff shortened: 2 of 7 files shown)


ksolanki7 (approved):
I believe the changes are from the previous PR for Tag component and doesn't require more review feedback on it.
Looks good to me.

nurm717123:
> I believe the changes are from the previous PR for Tag component and doesn't require more review feedback on it. Looks good to me.

yup, just one addition `isStandalone` props for `Tag`, for the usage when consumer use it without `TagGroup` so it render as `span` instead of `li`.just let me know if It need more correction and I will update it, thanks.

kurtdoherty:
@nurm717123 is it really necessary to use `<li>` items when in the tag group?

nurm717123:
> @nurm717123 is it really necessary to use `<li>` items when in the tag group?

yes, as I use `ul` and `li` pattern for `TagGroup`, because it usually will be used to display list of status, roles, .etc right?, what do you think 🤔 ?

kurtdoherty:
@nurm717123 are we basing the use of ul and li for the tag group on any a11y research or it just a gut feel?

I'm not sure many devs will remember to supply an `isStandalone` prop to a Tag when they use it outside a tag group. I think we should either be handling that automatically for devs, or perhaps better, allowing devs to decide what DOM element the tag group and tag render as.

nurm717123:
> @nurm717123 are we basing the use of ul and li for the tag group on any a11y research or it just a gut feel? (Nothing wrong with gut feels, just seeking clarity).

yeah, I done a research and I came up with 2 choice
1. use `ul` and `li`
2. use `div` with role (e.g list and list-item)

and my consideration is, even if we use `div`, I think the component still need to know whether it is used as standalone or group, to provide correct role.

> I'm not sure many devs will remember to supply an `isStandalone` prop to a Tag when they use it outside a tag group. I think we should either be handling that automatically for devs, or perhaps better, allowing devs to decide what DOM element the tag group and tag render as.

yeah, I agree.

kurtdoherty:
Okay, thanks for clarifying 👍 

I'm okay with either option you mention, I just think we should either handle that decision making automatically for devs OR force them to think about the elements (instead of a more abstract concept like "standalone").

nurm717123:
> Okay, thanks for clarifying 👍
> 
> I'm okay with either option you mention, I just think we should either handle the "inside tag group" vs "solo tag" decision making automatically for devs OR force them to think about the DOM _elements_ themselves (e.g. via an `as="span"` or `as="li"` required prop) instead of a more abstract concept like "standalone".
> 
> Edit: to be clear though, this is _not_ blocking feedback. It's easy enough to deprecate `isStandalone` in future if we do the extra work. Happy for you to move ahead as is if you want to keep moving.

Thanks, I think I will merge it first, and I will create another ticket to discuss how to approach it 👍

nurm717123:
> Okay, thanks for clarifying 👍
> 
> I'm okay with either option you mention, I just think we should either handle the "inside tag group" vs "solo tag" decision making automatically for devs OR force them to think about the DOM _elements_ themselves (e.g. via an `as="span"` or `as="li"` required prop) instead of a more abstract concept like "standalone".
> 
> Edit: to be clear though, this is _not_ blocking feedback. It's easy enough to deprecate `isStandalone` in future if we do the extra work. Happy for you to move ahead as is if you want to keep moving.

instead of creating new ticket, I think I will work on it as part of current ticket and add a new comment to propose the approach, please have a look [here](https://github.com/reapit/elements/issues/241#issuecomment-2601266229)

Issue #242, opened 2024-12-09
v5 component refactor: Badge


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing
- Is the component being re-named? No
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=1245-7191&node-type=section&m=dev)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/badge--docs)
- [Accessibility requirements](https://m3.material.io/components/badges/accessibility)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1441-5560&m=dev)

**PR Checklist**

 - [x] Deprecate existing Badge component
https://github.com/reapit/elements/pull/271
 - [ ] Implement a new Badge component as per DS
 - [ ] Retire existing Badge component (Open for discussion if this PR is required or not)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

DanishAli-Reapit:
Hello Team,

I have reviewed this ticket's requirements and summarized my understanding below. Please review and let me know if I've missed anything or if adjustments are needed to my approach.

**Updates to the Button Component:**

Since the Badge component already exists, the necessary updates will be applied directly. The required changes are as follows:

**1. UI Updates:**

- Adjust the padding, font size, and other styles according to the design system (DS).

**2. New Variants for the Badge Component:**

**Badge Varients:** 

- Default
- Reversed

[...]

PR #273, opened 2025-01-09
feat: #242 v5 badge component

DanishAli-Reapit:
# Description

- Created a new v5 Badge component to reflect DS Figma
- Added accessibility support to the badge

https://github.com/user-attachments/assets/12dcdd82-f6be-4d8a-8680-981527c32b76


src/components/badge/badge.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC, HTMLAttributes, ReactNode } from 'react'
+import { ElBadge, ElBadgeLabel } from './styles'
+
+type BadgeVariant = 'neutral' | 'success' | 'pending' | 'warning' | 'danger' | 'inactive' | 'accent_1' | 'accent_2'
+
+interface BadgeProps extends HTMLAttributes<HTMLDivElement> {
+  children?: ReactNode
+  reversed?: boolean
```
ksolanki7, line 8: **suggestion:** I would prefer to use boolean props with an [auxiliary verb](https://en.wikipedia.org/wiki/Auxiliary_verb) i.e. `isReversed`. `reversed` indicates a condition (the background color is reversed or not), making it a state representing a true/false value. It could be more intuitive if named something like `isReversed` or `reverse` to clearly reflect that it's a boolean value denoting the current state of the background.

ksolanki7 (commented):
**praise:** Congratulations on your 1st element implementation PR. There are some couple of suggestions I have left in the PR feedback.

src/components/badge/styles.ts
```diff
@@ -0,0 +1,173 @@
+import { css } from '@linaria/core'
+import { styled } from '@linaria/react'
+
+export const elBadgeLabel = css``
```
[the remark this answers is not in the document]
DanishAli-Reapit: Its use in the badge.tsx line number 36 `{children && <ElBadgeLabel>{children}</ElBadgeLabel>}`

src/components/badge/badge.tsx
DanishAli-Reapit: Thanks good suggestion I will update my PR.

src/components/badge/styles.ts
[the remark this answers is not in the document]
ksolanki7: **issue (non-blocking):** ElBadgeLabel is already instantiated for the styled div [here](https://github.com/reapit/elements/pull/273/files#:~:text=export%20const-,ElBadgeLabel,-%3D%20styled.). To utilize it effectively, you only need to define this variable for instantiation, as the className is specifically used for targeted styling. For example: 

```
export ElBadge = styled.div`
    .${elBadgeLabel} {
        ....
    }
...
`
```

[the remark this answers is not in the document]
DanishAli-Reapit: I will remove the const declare in line 4 as it is not used in the file.

src/components/badge/badge.tsx
```diff
@@ -0,0 +1,40 @@
...
+type BadgeVariant = 'neutral' | 'success' | 'pending' | 'warning' | 'danger' | 'inactive' | 'accent_1' | 'accent_2'
+
+interface BadgeProps extends HTMLAttributes<HTMLDivElement> {
+  children?: ReactNode
+  isReversed?: boolean
+  variant?: BadgeVariant
+  iconLeft?: ReactNode
+  iconRight?: ReactNode
+  'aria-label'?: string
+  className?: string
```
kurtdoherty, line 13: **quibble:** `aria-label` and `className` shouldn't need to be included explicitly; they'll be part of the `HTMLAttributes<HTMLDivElement>` interface that we're extending here.

src/components/badge/badge.tsx
```diff
@@ -0,0 +1,40 @@
...
+  iconLeft,
+  iconRight,
+  'aria-label': ariaLabel,
+  className,
+  ...rest
+}) => {
+  return (
+    <ElBadge
+      data-variant={variant}
+      data-isReversed={isReversed}
```
kurtdoherty, line 29: **issue:** `data-*` attributes should use snake/hyphen case. That is, this should be `data-is-reversed`. Doing this ensures JavaScript access of the data attribute has the camel casing we'd expect (`isReversed`). You can read more about this in the [MDN docs on using data attributes](https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Solve_HTML_problems/Use_data_attributes#javascript_access)

src/components/badge/styles.ts
```diff
@@ -0,0 +1,172 @@
+import { css } from '@linaria/core'
+import { styled } from '@linaria/react'
+
+export const elIcon = css``
```
kurtdoherty, line 4: **question:** I don't see this used in the actual Badge component as a wrapper around the `iconLeft` or `iconRight`. So is this class actually doing what the CSS below suggests it is doing?

kurtdoherty (commented):
**praise:** As @ksolanki7 said, congrats on your first contribution to Elements 👏  Lot of good work here; particularly nice to see some consideration going to the storybook control values 👍

src/components/badge/badge.tsx
DanishAli-Reapit: Thanks @kurtdoherty for the feedback. I will update my PR accordingly.

src/components/badge/badge.tsx
DanishAli-Reapit: Thanks @kurtdoherty for the feedback. I will remove them from my PR.

src/components/badge/styles.ts
DanishAli-Reapit: Thanks @kurtdoherty for the feedback. Yes, this class was not required. I have updated my PR. Please review.

src/components/badge/styles.ts
```diff
@@ -0,0 +1,169 @@
...
+  display: inline-flex;
+  height: var(--size-5);
+  padding: var(--spacing-half) var(--spacing-1);
+  align-item: center;
+  gap: var(--spacing-half);
+  border-radius: var(--corner-default);
+  color: var(--text-info);
+  background: var(--fill-info-lightest);
+
+  .el-icon {
```
kurtdoherty, line 13: **issue:** I think we've misunderstood each other. Previously, you had a local `elIcon` CSS class defined that you were using here. That was fine, it just wasn't being used explicitly in our Badge.

Now, you've hardcoded what I assume is the `Icon` component's class within our Badge component's CSS. This creates an implicit (hidden), tight coupling between our Badge component and our Icon component. We don't want to do this, because it means the behaviour owned by our Badge (changing the icon colour) will only work if our Icon component is used to handle the display of the icon and that Icon component does not change the name of it's primary class.

Until we implement a new Icon component, or update how the existing one is implemented so that it respects the colour of its ancestor element, we want to do two things:

1. We want to implement our Badge in a way that is agnostic of what component is used to render the icon; and,
2. Add a well-commented, explicit dependency on the `ElIcon` styled component so that:
  a. We force `ElIcon` to have `color: inherit`; and,
  b. It's easy for us to delete that "patch" when we refactor/replace Icon.

I believe this is what you'll see in some of the other components we've recently worked on that also need to influence the colour of icons provided via `ReactNode` props (IIRC, `MobileNavItem` may be taking this approach)

DanishAli-Reapit: Thanks @kurtdoherty for the review. I was not thinking that way. I have now updated my PR and forced `ElIcon` to have a `color: inherit`

src/components/badge/styles.ts
```diff
@@ -24,135 +25,75 @@ export const ElBadge = styled.div`
   &[data-is-reversed='true'] {
     color: var(--text-white);
     background: var(--fill-info-dark);
-
```
kurtdoherty: **question:** since the icon's are using different CSS variables than the badge text colour, don't we still need to wrap the icon in an element controlled by our Badge component and have styles for that element similar to what you had here?

DanishAli-Reapit: I know the Icon has different CSS variables but the value is the same and by using the icon color `inherit` it will use the same color as the label text color. If we want to use the icon CSS variables then I can't declare the icon color as `inherit` I have to use the icon color like `var(--icon-white)`

kurtdoherty: Hmm, when we're using CSS variables, I think we should be focused on using the correct, _semantic_ CSS variable in the correct context. Allowing icons to be coloured based on a text-related CSS variable means we're misusing that CSS variable, even if it happens to have the same value as it's icon-related counterpart. Ultimately, I don't actually understand _why_ Design have separate text and icon colour design tokens, but it's what we currently have to work with.

In a previous comment about the relationship between Badge and Icon, I communicated the following:

> Until we implement a new Icon component, or update how the existing one is implemented so that it respects the colour of its ancestor element, we want to do two things:
> 
> 1. We want to implement our Badge in a way that is agnostic of what component is used to render the icon. What we want is to wrap whatever icon element is provided with our own element that applies the styles we want (like the color); and,
> 2. Add a well-commented, explicit dependency on the ElIcon styled component so that:
>   a. We force ElIcon to have color: inherit; and,
>   b. It's easy for us to delete that "patch" when we refactor/replace Icon.

I'll try to expand on what I meant by (1) and (2)...

By (1), I mean our structure probably needs to be:
```
<ElBadge> // <--- sets "color: var(--text-white);"
    <ElBadgeIcon>...</ElBadgeIcon> // <---- sets "color: var(--icon-white);"
    <ElBadgeLabel>Label</ElBadgeLabel>
  </ElBadge>
```

By (2), I mean, our CSS for `ElBadgeIcon` (or `ElBadge`) monkey-patches/overrides descendant `ElIcon` elements to ensure they inherit their ancestor's colour, which will be what ElBadgeIcon has set.

More than happy for alternative suggestions, but this is the only way I can see at the moment for us to have a good change of correctly colouring whatever icon is supplied to the Badge component, while also ensuring use of the current Elements Icon component behaves how we need it to when inside a Badge.

DanishAli-Reapit: Thanks @kurtdoherty for the feedback. I have updated my with your suggestions also applied the same on the split-button component.

kurtdoherty (approved):
Thanks for your patience with this one @DanishAli-Reapit 🙏 I think we can merge it in now.

Issue #263, opened 2024-12-18
v5 component refactor: Token Variables update - Font Weight


ksolanki7:
**Background context:**

The current token variables for font-weight use labels like Regular, Medium, and Semibold for font-weight and is not effective i.e. as follows
```
  --font-weight-regular: Regular;
  --font-weight-medium: Medium;
  --font-weight-semibold: Semi Bold;
```

DS figma has been updated to use the numeric values and so needs to get the token file revised to reflect the same approach.

**Task**

Update the token files to match the correct variable values as defined below.![image](https://github.com/user-attachments/assets/29d633bf-efab-4c0f-85d6-e86dde3c11f2)

Here is the link for the Figma - [Figma](https://teams.microsoft.com/l/message/19:4a9faeca-7633-496e-a084-e10b4ccc4d1d_6ac3b40c-1a44-4012-849b-210ea8882c70@unq.gbl.spaces/1734397634014?context=%7B%22contextType%22%3A%22chat%22%7D)

PR #275, opened 2025-01-09
chore: #263 v5 Fix token file for font-weight variable values

ksolanki7:
# Description

**fixes:**
- Updated token files to read correct font-weight values same as figma
- Updated respective files using hard-coded font-weight values.


kurtdoherty (approved):
Thanks for actioning this @ksolanki7!

PR #278, opened 2025-01-14
feat: 272 v5 label component

rpt-rfoxy:
# Pull request checklist

fixes: #
#### **2. New Parameters**

# Label Solution Design

The Label Text will have the following main prop interface

| Prop | Value | Default | Usage |
| --- | --- | --- |--- |
| variant?: union  | "soft" , "strong", undefined |soft | label variant depend on text color |
| size?: union | "small" , "medium", undefined | medium | label size using --font-size-xs  and --font-size-sm |
| isRequired?: boolean | true, false, undefined | undefined | to set `*` mark in the end of element child |

---

source: 
src/components/label-text

image


src/components/label/label.stories.tsx
```diff
@@ -1,10 +1,28 @@
...
+        example label strong medium
+      </Label>
+      <br />
+      <Label variant="soft" size="small">
+        example label soft small
+      </Label>
+      <br />
+      <Label variant="soft" isRequired>
+        example label soft required
+      </Label>
```
adrian-reapit, line 25: Suggestion: It would be nice to see the Label next to an input so the usage is more clear and can be tested here
image

rpt-rfoxy:
hi @adrian-reapit , thanks for the feedback, i make new branch for label and rename it as LabelText to make it easiest for review, here the PR: https://github.com/reapit/elements/pull/298

PR #285, opened 2025-01-22
feat: #241 v5 new Tag component props refactor

nurm717123:
following the discussion [here](https://github.com/reapit/elements/issues/241#issuecomment-2603753748), we decided to use `as="span"` patten instead of `isStandalone` props pattern

additionally as the term/props changed, I update the exported styles component name
- `ElStandaloneTag` -> `ElTag`
- `ElTag` -> `ElTagGroupItem`

note: react user can still use them as `<Tag />`

# Pull request checklist
- [x] refactor props
- [x] rename styles var name
- [x] update stories and test

fixes: #241

### DIFF
--- a/src/components/tag/tag.tsx
+++ b/src/components/tag/tag.tsx
@@ -1,11 +1,11 @@
-import { ElStandaloneTag, ElTag, ElTagGroup } from './styles'
+import { ElTag, ElTagGroupItem, ElTagGroup } from './styles'
 
 export interface TagProps extends React.HTMLAttributes<HTMLLIElement | HTMLSpanElement> {
-  isStandalone?: boolean
+  as?: 'span' | 'li'
 }
 
-export const Tag: React.FC<TagProps> = ({ children, isStandalone = false, ...props }) => {
-  return isStandalone ? <ElStandaloneTag {...props}>{children}</ElStandaloneTag> : <ElTag {...props}>{children}</ElTag>
+export const Tag: React.FC<TagProps> = ({ children, as = 'li', ...props }) => {
+  return as == 'span' ? <ElTag {...props}>{children}</ElTag> : <ElTagGroupItem {...props}>{children}</ElTagGroupItem>
 }
 
 export const TagGroup = ElTagGroup
--- a/src/components/tag/tag.tsx
+++ b/src/components/tag/tag.tsx
@@ -1,9 +1,16 @@
 import { cx } from '@linaria/core'
 import { elTag, ElTagGroup } from './styles'
+import type { HTMLAttributes } from 'react'
 
-export interface TagProps extends React.HTMLAttributes<HTMLLIElement | HTMLSpanElement> {
-  as?: 'span' | 'li'
+interface TagAsSpanProps extends HTMLAttributes<HTMLSpanElement> {
+  as: 'span'
 }
+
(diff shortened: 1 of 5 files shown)


src/components/tag/tag.tsx
```diff
@@ -1,11 +1,11 @@
...
 
 export interface TagProps extends React.HTMLAttributes<HTMLLIElement | HTMLSpanElement> {
-  isStandalone?: boolean
+  as?: 'span' | 'li'
 }
 
-export const Tag: React.FC<TagProps> = ({ children, isStandalone = false, ...props }) => {
-  return isStandalone ? <ElStandaloneTag {...props}>{children}</ElStandaloneTag> : <ElTag {...props}>{children}</ElTag>
+export const Tag: React.FC<TagProps> = ({ children, as = 'li', ...props }) => {
+  return as == 'span' ? <ElTag {...props}>{children}</ElTag> : <ElTagGroupItem {...props}>{children}</ElTagGroupItem>
```
kurtdoherty, line 8: **issue:** While this is nice and simple, when the standalone version gets removed in future, I think we'd want to have two classes at play: `el-tag-group` and `el-tag`. I we did that, consumers currently relying on the `el-tag-group-item` would have to replace all their usage of that class, which isn't great.

Given the approach in `styles.ts` where the template literal is shared between two styled components, `ElTag` and `ElTagGroupItem`, I see no reason for why we need to have those components strictly coupled to `<span>` and `<li>` elements respectively. Instead, I think it makes more sense to do:

```javascript
// styles.ts

const elTag = css`
  // styles for the tag atom
`

// tag.tsx
function Tag({ as: Element = 'li', ...rest }: TagProps) {
  return <Element {...rest} />
}
```

This approach allows the React `Tag` component to render as whatever `as` element is specified (in this case, we're only allowing two options) and we still have a well-named class name, `el-tag` that CSS only consumers can leverage.

nurm717123: Ah yes, I thought we going to `retire the Tag component altogether` means all of it, but turns out just the standalone one , definitely agree then, that's better approach, thanks 👍

src/components/tag/tag.tsx
```diff
@@ -1,11 +1,16 @@
-import { ElStandaloneTag, ElTag, ElTagGroup } from './styles'
+import { cx } from '@linaria/core'
+import { elTag, ElTagGroup } from './styles'
 
 export interface TagProps extends React.HTMLAttributes<HTMLLIElement | HTMLSpanElement> {
-  isStandalone?: boolean
+  as?: 'span' | 'li'
 }
```
kurtdoherty, line 6: **note:** I think we'd probably want a discriminated union here...though I'm not 100% sure if there's any practical difference in the interfaces of the `HTMLSpanElement` or `HTMLLIElement`. If we were typing the ref for this component, we'd certainly want to know what DOM element we're working with though.

```typescript
interface TagAsSpanProps extends HTMLAttributes<HTMLSpanElement> {
  as: 'span'
}

interface TagAsListItemProps extends HTMLAttributes<HTMLSpanElement> {
  as?: 'li'
}

type TagProps = TagAsSpanProps | TagAsListItemProps
```

Probably not a big deal right now, so I think it's okay if you decide to leave it as is.

kurtdoherty (approved):

PR #290, opened 2025-01-29
fix: 289 added badge label container

DanishAli-Reapit:
# Description

- Added Badge Label Container


src/components/badge/badge.tsx
```diff
@@ -31,7 +31,11 @@ export const Badge: FC<BadgeProps> = ({
       {...(rest as HTMLAttributes<HTMLDivElement>)}
     >
       {iconLeft && <ElBadgeIcon>{iconLeft}</ElBadgeIcon>}
-      {children && <ElBadgeLabel>{children}</ElBadgeLabel>}
+      {children && (
+        <ElBadgeLabelContainer>
```
ksolanki7, line 35: @DanishAli-Reapit 

**question:** Is there any specific reason to add an extra HTML element, i.e., `div` to wrap around the span?

DanishAli-Reapit: Yes, we can. I have updated my PR.

Issue #291, opened 2025-01-29
v5 component refactor: Tooltip - Bug


ksolanki7:
**Describe the bug**
Hovering over the trigger flickers the tooltip for a milli-second as an extreme left position on Canvas before it places it in the correct location over the Canvas.

**To Reproduce**
Steps to reproduce the behavior eg:
1. Go to Tooltip -> https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-tooltip--basic-usage
2. Hover multiple times over the trigger i.e. Button, do it fast
3. You will notice the displacement of tthe ooltip initial and it positions it correctly after a milli-second

**Expected behaviour**
Tooltip should have animation of appearing next to the trigger as defined by the placement i.e. `position` prop or default to top

**Screenshots**

![Image](https://github.com/user-attachments/assets/c1d8c6ce-056c-4d95-a660-c2b2542cacbd)

ksolanki7:
@fazil-99  @andrei-dragut Ready for QA

andrei-dragut:
@ksolanki7 I think we're good. Can't replicate the bug anymore

PR #293, opened 2025-01-29
fix: #291 Fixed issue with the placement of the tooltip

ksolanki7:
# Pull request checklist

fixes: Added animation as the issue was raised due to race condition of rendering the tooltip to DOM v/s the calculation done to get the position of the trigger to place the tooltip.

### DIFF
--- a/src/components/tooltip/use-tooltip.ts
+++ b/src/components/tooltip/use-tooltip.ts
@@ -130,6 +130,7 @@ export const useTooltip = () => {
     id: tooltipId,
     role: 'tooltip',
     'aria-hidden': !isVisible,
+    'data-visible': isVisible,
     isVisible,
   })
 
--- a/src/components/tooltip/use-tooltip.ts
+++ b/src/components/tooltip/use-tooltip.ts
@@ -130,7 +130,7 @@ export const useTooltip = () => {
     id: tooltipId,
     role: 'tooltip',
     'aria-hidden': !isVisible,
-    'data-visible': isVisible,
+    'data-is-visible': isVisible,
     isVisible,
   })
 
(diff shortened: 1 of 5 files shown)


src/components/tooltip/use-tooltip.ts
```diff
@@ -130,6 +130,7 @@ export const useTooltip = () => {
     id: tooltipId,
     role: 'tooltip',
     'aria-hidden': !isVisible,
+    'data-visible': isVisible,
```
kurtdoherty, line 133: **nit-pick:** same thing I said on the other PR 😅 Sorry, I didn't notice the other one was closed until I'd submitted the review...

TL;DR `data-visible` -> `data-is-visible` because [reasons](https://consolegroup.atlassian.net/wiki/spaces/DP/pages/3127509024/Prop+Interface+Guidelines#On-Data-Attributes).

kurtdoherty (approved):
As I said on the other PR for these changes, I'm not sure what the underlying issue is or how adding animation solves it, but the changes here look fine on their own merit.

src/components/tooltip/use-tooltip.ts
ksolanki7: @kurtdoherty When the tooltip is rendered for the first-time, it doesn't have the correct dimensions because it hasn't been fully rendered and measured yet. The browser may use the default positioning of `position: absolute; inset: 0px auto auto 0px;`, which results in the tooltip being rendered at the end of the `body` of the page momentarily before the position calculation happens. Delaying the animation fade-in effect and opacity resolves the initial renderer issue of the tooltip.

kurtdoherty: Ahh, okay. Thanks for the more detailed explanation 👍 

Two thoughts:
- I think it's fairly common to have tooltips delay a little;
- In future, we may be able to leverage the [@starting-style](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style) rule to work around this kind of issue with animations and needing to measure the DOM element.

ksolanki7: @kurtdoherty Firefox now offers partial support for this feature. I tested it in the latest version of Firefox, and the `opacity` property appears to work correctly within @starting-style.

I'll update the PR

Issue #224, opened 2024-11-28
v5 new components: BottomBar


ss-dimasm:
**Abstract**

While the smaller part of the `BottomBar` is live in main (#192), we could start the next step by scaffolding the BottomBar based on the design spec. Further technical discussion is required before we proceed with crafting the production-grade implementation

**Specification**
- Is new component or existing? new
- Is the component being re-named? no
- Is the component design signed off by product and design? yes
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=34-7016&node-type=instance&m=dev)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-25700&node-type=symbol&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**
- The `BottomBar` should be hidden when users scroll down and should appear when users scroll up [Detail Link](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=34-7016&node-type=instance&m=dev)
- When the expanded menu is visible and the user scrolls down, the panel should close, and the bar should be hidden [Question Link](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=34-7014&m=dev#1040987505)

ss-dimasm:
Hi @kurtdoherty and @adrian-reapit , FYI I've raised this ticket to initiate the investigation of BottomBar implementation. I will provide updates in this ticket and create a new bare minimum proof-of-concept (POC) first. Then, I'll share the findings with you, including considerations for the next steps thanks 👍

ss-dimasm:
[...]

PR #296, opened 2025-01-30
fix: #224 bottom bar item design feedback

ss-dimasm:
# Pull request checklist

fixes: #
part of #224 

**Context:**
Following the unmatched style between the design and actual component feedback https://github.com/reapit/elements/issues/224#issuecomment-2620769647 we also need to little amendment for this one

### DIFF
--- a/src/components/bottom-bar/bottom-bar.atoms.tsx
+++ b/src/components/bottom-bar/bottom-bar.atoms.tsx
@@ -1,8 +1,9 @@
 import type { FC, ReactNode } from 'react'
 
-import { Menu, MenuItemProps } from '../menu'
+import { elWFull } from '../../styles/sizing'
 import { BottomBarItem } from '../bottom-bar-item'
 import { Icon } from '../icon'
+import { Menu, MenuItemProps } from '../menu'
 
 export interface BottomBarMoreMenuProps {
   children: ReactNode
@@ -10,7 +11,7 @@ export interface BottomBarMoreMenuProps {
 
 export const BottomBarMoreMenu: FC<BottomBarMoreMenuProps> = ({ children }) => {
   return (
-    <Menu data-alignment="right">
+    <Menu data-alignment="right" className={elWFull}>
       <Menu.Trigger>
         {({ getTriggerProps }) => (
           <BottomBarItem {...getTriggerProps()} icon={<Icon icon="more" />}>
--- a/src/components/bottom-bar/bottom-bar.atoms.tsx
+++ b/src/components/bottom-bar/bottom-bar.atoms.tsx
@@ -1,9 +1,9 @@
 import type { FC, ReactNode } from 'react'
 
-import { elWFull } from '../../styles/sizing'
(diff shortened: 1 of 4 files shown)


src/components/bottom-bar/styles.ts
```diff
@@ -17,6 +17,11 @@ export const ElBottomBar = styled.div`
   margin-top: auto;
   width: 100%;
 
+  // NOTE: this selector will force the first level of child to have the same width
```
kurtdoherty, line 20: **issue:** I think this is a code smell and suggests we're either missing styles elsewhere or that we are missing an atom that handles this for us. In this case, I think it's the former (we're missing styles elsewhere).

Specifically, I think this suggests our `MoreMenu` atom should be applying styles to the Menu atom it encapsulates to ensure that component renders to an element whose width matches what our other bottom bar items use.

kurtdoherty (commented):
Thanks @ss-dimasm, one blocker for me at the moment. If the approach I'm suggesting doesn't work, we can bail and run with what you have here.

src/components/bottom-bar/styles.ts
ss-dimasm: > I think this suggests our MoreMenu atom should be applying styles to the Menu atom it encapsulates

Looks more wise than before, would try this one 👍

src/components/bottom-bar/bottom-bar.atoms.tsx
```diff
@@ -1,16 +1,17 @@
...
+import { Menu, MenuItemProps } from '../menu'
 
 export interface BottomBarMoreMenuProps {
   children: ReactNode
 }
 
 export const BottomBarMoreMenu: FC<BottomBarMoreMenuProps> = ({ children }) => {
   return (
-    <Menu data-alignment="right">
+    <Menu data-alignment="right" className={elWFull}>
```
ss-dimasm, line 14: hey @kurtdoherty , I ended up by adding declared css variable to ensure it has 100% width; would love to hear your thought on this

kurtdoherty: **suggestion:** Hmm, I'm not sure if those utility classes only exist to support v4 components or if they're a long-term thing we're wanting v5 to provide. I think I'd probably feel more comfortable with us doing:

```javascript
// styles.ts
const ElBottomBarMoreMenu = styled(Menu)`
  width: 100%
`

// bottom-bar.atoms.tsx
export const BottomBarMoreMenu: ... = ({ ... }) => {
  return (
    <ElBottomBarMoreMenu ...>
       ...
    </ElBottomBarMoreMenu>
  )
}
```

Not a blocking concern though.

kurtdoherty (approved):

src/components/bottom-bar/bottom-bar.atoms.tsx
ss-dimasm: thanks @kurtdoherty , it's more solid would give a try

Issue #272, opened 2025-01-07
v5 component refactor: Label


ksolanki7:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing
- Is the component being re-named? No
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=3592-3299&m=dev)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/label--docs)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

ksolanki7:
@HollyJoyPhillips I have created this issue for new element Label which dev ready over figma.

rpt-rfoxy:
Hello Genks,

From the Figma design, the new Label design only requires adding new parameters, so there's no need to add deprecated info. Here's my proposed solution:

---

### **Updates to the Label Component**

#### **1. UI Updates**
- No new design is required, only two variants will be added.

#### **2. New Parameters**

- **intent**  
  (Should we retain this variant name as "intent," or has the team discussed a new name?)  
  - `soft` (default)  
  - `strong`

- **Label Size**  
  - `small`  
  - `medium`

- **Required Indicator**  
  For the required indicator, we will used ("*") as hardcoded mark.

- **Text Overflow**  
  By default, the text overflow will continue on a new line, depending on the wrapper.

---

### **Code Concept**

```typescript
// Label.tsx
export interface LabelProps extends LabelHTMLAttributes<HTMLLabelElement> {
  intent?: "soft" | "strong"
[...]

PR #298, opened 2025-02-03
feat: #272 v5-label-text component

rpt-rfoxy:
# Pull request checklist

As Perdiscussion, so we will used static rendered span, and rename the label as LabelText [Discussion](https://github.com/reapit/elements/issues/272#issuecomment-2620688093)

here the preview:
image


src/components/label-text/styles.ts
```diff
@@ -0,0 +1,16 @@
+import { styled } from '@linaria/react'
+
+export const ElLabelText = styled.span`
```
ksolanki7, line 3: @rpt-rfoxy 
**issue (blocking):** I believe the CSS properties for the label are not applied. Please refer to DS figma. https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=3592-3299&m=dev
image

rpt-rfoxy: hi @ksolanki7 i already update the styles as the requarement, please if you have freetime give me feedback

ksolanki7: @rpt-rfoxy
**issue (blocking):** As discussed yesterday, it is still missing the parent container styles.

rpt-rfoxy: hi @ksolanki7 i already update the label text,
image

i used span as the span container, so it makes the stylesheet more easy to implement to the children

image

and the style:

image

its will keep LabelText keep render as span

src/components/label-text/styles.ts
```diff
@@ -0,0 +1,25 @@
...
+  font-family: var(--font-family);
+  font-style: normal;
+  font-weight: var(--font-weight-regular);
+  line-height: var(--line-height-sm);
+  letter-spacing: var(--letter-spacing-sm);
+  display: inline-flex;
+  align-items: center;
```
ksolanki7, line 10: @rpt-rfoxy
**issue (non-blocking):** `align-items: center` CSS property is not required for LabelText component, I believe we discussed this over the meeting that Figma adds certain CSS properties which aren't required for actual Web implementation.

rpt-rfoxy: hi @ksolanki7 , thanks for the feedback, i lareday update the code as your feedback to remove `align-items: center` 

Screenshot from 2025-02-07 08-06-57

ksolanki7 (approved):
Looks good now.

ksolanki7:
**issue (non-blocking):** We may need to revisit on the stories, how we represent different stories and description for it.

Issue #138, opened 2024-07-30
v5 component refactor: BreadCrumbs


HollyJoyPhillips:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing 
- Is the component being re-named? Yes - currently 'BreadCrumbs' - DS is 'Breadcrumb trail'
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-
7363&t=8GcgX59FmafMRAda-4)
- [Link to Elements Audit in Figma](https://www.figma.com/design/J6EneNGfpk2GcUuymiOKW4/Elements-Gaps-%2B-Discrepancies?node-id=269-9991&t=pXCAT2b1xMUuVVdm-4)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/breadcrumb--docs)
- [Accessibility requirements](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

nurm717123:
Hi Kurt, and Adrian.

I've done to explore the approach to build the breadcrumbs, and here is what I come up with.

### Context
- the component structure is pretty simple but have some complexity at component overflow behaviour

### Requirements 
- in non mobile screen when the items is long enough to overflow the sreen, it should display ellipsis so it become shorter, but if its too tight, it allows to collapse to multiple line
- in mobile screen, it allows horizontal scroll but don't display the scroll 

[...]

PR #299, opened 2025-02-03
feat: #154 v5 new SideBar (menu-list & menu-item)

nurm717123:
## Context
In order to create a new SideBar for the v5 release, this PR, as part of #154, initended to initiate the `SideBarMenuList` and `SideBarMenuList` for SideBar.

_ps: I think that `SideBarMenuList` is not perfect without the `SideBarMenuItem` (for demo), so I make it as one pr_

## Pull request checklist
- [x] initiate `SideBarMenuList` or `SideBar.MenuList`
- [x] initiate `SideBarMenuItem`  (support anchor and button)

fixes: #

sidebar 2


src/components/side-bar-menu-item/styles.ts
```diff
@@ -0,0 +1,60 @@
...
+    .el-icon {
+      color: var(--text-action);
+      font-weight: 500;
+    }
+  }
+
+
+  .el-icon {
+    width: var(--icon-md);
+    height: fit-content;
```
nurm717123, line 48: the design define the height as 'var(--icon-md)`, but it causing result inconsistency in height (i.e there is more space above the icon)

image

in above example: 
the 1st list use `icon-md` as height
the 2nd list use `fit-content`

but the draw back is the height is off 2px
image

I think I need to ask the design team, because they use different icon that unavailable in elements

src/components/side-bar-menu-item/styles.ts
```diff
@@ -0,0 +1,58 @@
+import { styled } from '@linaria/react'
+
+const baseButtonOverride = `
+  cursor: pointer;
+  background: none;
+  border: none;
+  width: 100%;
+`
```
nurm717123, line 8: as in the poc, this can be used as base style of `ElSubMenuItemButton` later

src/components/side-bar-menu-item/side-bar-menu-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * @example
+   * <SideBarMenuItem isActive>This is active page link</SideBarMenuItem>
+   *
+   * @default false
+   */
+  isActive?: boolean
+  icon: ReactNode
+}
+
```
kurtdoherty, line 17: **question:** do we really need to support side bar items as buttons? Feels like everything in the sidebar should be a proper page in the product and will therefore have a URL... 🤔 

**suggestion:** Unless there's some compelling use-case for button-based side bar items, I think it would be worth simplifying things here and only supporting anchor-based items.

src/components/side-bar-menu-item/styles.ts
```diff
@@ -0,0 +1,58 @@
...
+
+
+  &:hover {
+    background: var(--fill-default-lightest);
+    font-weight: 500;
+    &, .el-icon {
+          color: var(--neutral-600);
+    }
+  }
+
```
kurtdoherty, line 47: **issue:** We should be handling icons in the same generic manner we have elsewhere (the icon element should be wrapped by a styled component "owned" by the side bar menu item, and we should have some well-commented CSS that forces our `ElIcon` class to inherit its `color`).

At the moment,
- You're _implicitly_ coupling this component to the `.el-icon` class name here. We want that to be explicit by importing the `ElIcon` styled component and using it here.
- You're only ensuring icons rendered specifically by our current `Icon` component will be coloured correctly. Since we're accepting any ReactNode as the `icon`, we're not enabling other components to be able to render correctly here.

kurtdoherty (commented):
Good start @nurm717123 👏 Thanks for the small PR!

Just the one key issue around the handling of icons for the menu items.

src/components/side-bar-menu-item/side-bar-menu-item.tsx
nurm717123: >question: do we really need to support side bar items as buttons? Feels like everything in the sidebar should be a proper page in the product and will therefore have a URL... 🤔

I think so, for `react-router` user, so they can use onClick to navigate

> suggestion: Unless there's some compelling use-case for button-based side bar items, I think it would be worth simplifying things here and only supporting anchor-based items.

should we use `onClick` on `anchor` 🤔

src/components/side-bar-menu-item/styles.ts
nurm717123: yeah, that's good point, I thought we're going to only support our `Icon` and the reason I don't import `elIcon` directly is it doesn't work with string variable (i.e it will work with `css` or `styled`), I will try to try other workaround to solve the issue that you've mention, and get back to you, thanks 👍

src/components/side-bar-menu-item/side-bar-menu-item.tsx
nurm717123: Ah yes, there is standard behaviour in anchor 🤦 , my previous concern is when we enforce user to use `href` (with plain anchor here) , for react app they will not use full adventage of reac-router, causing it to do normal navigation (with reload page animation) , I think we can still use anchor here, and for react-router user the can override the behaviour using onClick (but its still anchor element) to do `preventDefault` and navigate using proper navigate function 🤔 , .eg

```javacsript
<SideBarMenuItem
  onClick={e => {
       e.preventDefault()
       navigate('/page1')
  }}>
Page 1
</SideBarMenuItem>
```

nurm717123: > React Router works perfectly fine with plain <a> elements with href attributes 

I've tested in in my app, but its navigate just like plain anchor 🤔 

> (in fact, RR's Link component renders as a plain <a>).

yep, but with my current approach, can't do as it will result in nested `a`

> The navigate function is only needed when a navigation needs to happen imperatively in situations that do not involve an <a> element (e.g. a successful form submission that needs to result in a navigation)

agree. by the way, I've update the pr to exclude support for button

src/components/side-bar-menu-item/__tests__/__snapshots__/side-bar-menu-item.test.tsx.snap
```diff
@@ -1,11 +1,11 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`SideBarMenuItem > should render anchor item and match snapshot 1`] = `
+exports[`SideBarMenuItem > should render active item and match snapshot 1`] = `
 <DocumentFragment>
   <li>
     <a
```
kurtdoherty, line 6: **question:** why do we no longer see the href for an active item?

kurtdoherty (approved):
Looks good 👍 Thanks @nurm717123

src/components/side-bar-menu-item/__tests__/__snapshots__/side-bar-menu-item.test.tsx.snap
nurm717123: Thanks for the review Kurt 👍 

its because in that test case I don't pass href (as its not required - but in the other one I have passed it)
I think I should've update the test title to be "should render without href + active and match snapshot" 🤔 

image

kurtdoherty: Ah, shouldn't `href` be a required prop for a sidebar item?

nurm717123: I think so, if its not, then the default `anchor` behaviour wouldn't work 😅 , I will update it to be required, thanks for noticing it 👍

nurm717123: I will update the pr to use the pattern we've discussed in Breadcrumb poc [here](https://github.com/reapit/elements/issues/138#issuecomment-2609094056), where we export the className, so both party can use the same style

nurm717123: but now the problem is now we can't use react-hook-form's `setFocus` because the component is initially not in the dom and  we set `shouldUnregister`, I think need to discuss it first

kurtdoherty: Hmm, I don't think that pattern works very well here.

Firstly, we have a single component that represents the `<a>` element AND the layout of its descendants (the icon and the label), whereas the `Breadcumbs` split those two concerns between the `Breadcrumbs.Item` and the `Breadcrumbs.Link` atoms.

Consumers of React Router's `Link` component should not be responsible for the internal layout of the side bar item's icon and label. I don't think we should be doing anything significant to support RR's Link component right now, as I'm not sure heavy usage of it and its relative `to` prop should be encouraged in products. For Breadcrumbs, I think it makes sense because the relative `to` prop allows us to do `to="../"` and `to="../.."`, which is convenient. But for the sidebar menu items, I think it's probably better for products to use a proper `href`.

src/components/side-bar-menu-item/styles.ts
```diff
@@ -0,0 +1,53 @@
...
+    width: var(--icon-md);
+    height: var(--icon-md);
+    color: var(--neutral-400);
+    box-sizing: content-box;
+  }
+
+  padding: var(--spacing-half);
+`
+
+export const elSideBarMenuItemAnchor = css`
```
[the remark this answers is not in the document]
kurtdoherty: So wouldn't it be better to just remove `ElSideBarMenuItem`?

src/components/side-bar-menu-item/__tests__/__snapshots__/side-bar-menu-item.test.tsx.snap
nurm717123: I think we still need to support react-router link (I've tried plain `a` but doesn't smoothly navigate as good as RR's link),
then I think I have other idea. i.e

use optional props
```javascript
<SideBarMenuItem as={RouterLink} icon={<Customicon />} href="/details">
    Custom LInk
</SideBarMenuItem>
 ```

what do you think?

Comment (kurtdoherty) [reply src/components/side-bar-menu-item/__tests__/__snapshots__/side-bar-menu-item.test.tsx.snap:6]: > I've tried plain a but doesn't smoothly navigate as good as RR's link

When you have time, can you create a demonstration of this difference? I'm interested to dig into it and understand why...

> then I think I have other idea...

Hmm, yes that might work, but I've encountered challenges with dynamic `as` props, refs and prop types from a typing perspective. Does passing a `href` when `as={RouterLink}` result in a smoother navigation?

Comment (kurtdoherty) [reply src/components/side-bar-menu-item/__tests__/__snapshots__/side-bar-menu-item.test.tsx.snap:6]: Another similar option could be:

```javascript
<SideBar.MenuList>
  <SideBarMenuItem icon={<Icon/>} href="/details">
    Standard LInk
  </SideBarMenuItem>
  <Link to="details">
    <SideBarMenuItem as="span" icon={<Icon/>}>
      Custom LInk
    </SideBarMenuItem>
  </Link>
<SideBar.MenuList> 
```

In this case, `<SideBarMenuItem as="span">` would no permit an `href` prop (because its prop types would be `HTMLAttributes<HTMLSpanElement>`. Not 100% sure this would be visually equivalent to your suggestion though 🤔 

We may need to descope RR's Link support for now and look at that challenge holistically across all components that may want to render as a Link (e.g. buttons, menu items, normal links, these side bar menus, etc).

nurm717123: > When you have time, can you create a demonstration of this difference? I'm interested to dig into it and understand why...

sure, I've created a quick sandbox [here](https://codesandbox.io/p/sandbox/yzfj52)

> Hmm, yes that might work, but I've encountered challenges with dynamic as props, refs and prop types from a typing perspective. Does passing a href when as={RouterLink} result in a smoother navigation?

refs probably doesn't work (to RouterLink but, do we need it?), but passing href when `as={RouterLink}` I believe going to work as expected

nurm717123: > 
> In this case, `<SideBarMenuItem as="span">` would no permit an `href` prop (because its prop types would be `HTMLAttributes<HTMLSpanElement>`. Not 100% sure this would be visually equivalent to your suggestion though 🤔
> 

I think it can also work too (100%), and I think its better than previous approach, except it add addition hierarchy, but I think it doesn't matter right?

> I think we may need to descope RR's Link support for now and look at that challenge holistically across all components that may want to render as a Link (e.g. buttons, menu items, normal links, these side bar menus, etc).

> Another similar option could be:
> 
> ```js
> <SideBar.MenuList>
>   <SideBarMenuItem icon={<Icon/>} href="/details">
>     Standard LInk
>   </SideBarMenuItem>
>   <Link to="details">
>     <SideBarMenuItem as="span" icon={<Icon/>}>
>       Custom LInk
>     </SideBarMenuItem>
>   </Link>
> <SideBar.MenuList> 
> ```

ok. then can I use this pattern first, for initial approach, or should I revert my `classname` change (latest commit)? 🤔

kurtdoherty: As I said above, I think we should descope Link compatibility for now and deal with it holistically across all components that may want to leverage RR's Link.

For now, I think we just keep the changes you've made (`elSideBarMenuItem` as a class we export, `SideBarMenuItem` just renders an `<a>` element with `elSideBarMenuItem` as its class), but no `as` prop yet.

nurm717123: got it 👍

Issue #258, opened 2024-12-17
v5 component refactor: Chip


kurtdoherty:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing
- Is the component being re-named? No, existing component will be deprecated and new component built in its place.
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?m=dev&focus-id=3256-8443)
- Accessibility requirements: No known ARIA guidelines

**Developer Checklist**

- [x] Styles alignment between Design System and Elements
- [x] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [x] Align with accessibility standards / spec as per above
- [x] If relevant, break down component into `Styles Only` and `React` component structures
- [x] Ensure all variants of components are documented as appropriate 
- [x] Ensure unit test coverage is adequate for component
- [x] Update documentation in MDX file as per guidelines
- [x] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [x] Approved PR merged to main
- [x] Design & product review and feedback addressed by developer
- [x] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

kurtdoherty:
@andrei-dragut @fazil-99 Chip and ChipGroup are available for review:
- [Chip](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-chip--docs)
- [ChipGroup](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-chipgroup--docs)

**note:** This work was merged prior to the agreement in Figma to add a maxWidth prop to the Chip component. I'll create a separate ticket for that to be added once we've rolled out the new variables.

kurtdoherty:
#314 added `maxWidth` so Chip is now complete

PR #303, opened 2025-02-06
feat: #258 deprecate existing chip and chip group components

kurtdoherty:
### Context

- We want to deprecate this existing Chip component to make room for a new <button>-based version

### This PR

- Renames the existing Chip, ChipGroup and related styled components so they are all marked as "Deprecated".


ksolanki7 (approved):
@kurtdoherty Looks good to me.

PR #304, opened 2025-02-06
feat: #258 add new `Chip` component

kurtdoherty:
### Context

- #303 deprecated this existing Chip (and Chip Group) component to make room for a new <button>-based version

### This PR

- Adds the new Chip component (a new ChipGroup component will be added in a separate PR).

Screenshot 2025-02-07 at 2 39 39 PM

**note:** The "Simple Chip" detailed in the Figma designs is currently marked as out-of-scope, so only the "Clearable Chip" has been implemented.

### DIFF
--- a/src/components/chip/__tests__/chip.test.tsx
+++ b/src/components/chip/__tests__/chip.test.tsx
@@ -1,0 +1,74 @@
+import { Chip } from '../chip'
+import { expect, test, vi } from 'vitest'
+import { fireEvent, render, screen } from '@testing-library/react'
+
+test('renders as a <button> element', () => {
+  render(<Chip variant="filter">Label</Chip>)
+  expect(screen.getByRole('button', { name: 'Label' })).toBeVisible()
+})
+
+test('filter chip has `data-variant="filter"` attribute', () => {
+  render(<Chip variant="filter">Label</Chip>)
+  expect(screen.getByRole('button', { name: 'Label' })).toHaveAttribute('data-variant', 'filter')
+})
+
+test('selection chip has `data-variant="selection"` attribute', () => {
+  render(<Chip variant="selection">Label</Chip>)
+  expect(screen.getByRole('button', { name: 'Label' })).toHaveAttribute('data-variant', 'selection')
+})
+
+test('chip has `type="button"` attribute', () => {
+  render(<Chip variant="selection">Label</Chip>)
+  expect(screen.getByRole('button', { name: 'Label' })).toHaveAttribute('type', 'button')
+})
+
+test('chip has no `aria-disabled` attribute when enabled', () => {
+  render(<Chip variant="selection">Label</Chip>)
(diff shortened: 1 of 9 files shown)


src/components/chip/__tests__/chip.test.tsx
kurtdoherty, line 1: **note:** I decided against snapshot tests here, largely because I wanted the test descriptions to be specific (e.g. "has xxx attribute" rather than "renders correctly"). I also need to do some more thinking around snapshot tests in Elements, especially where we're encapsulating other Elements components as this results in the snapshot capturing implementation detail of those other Elements components...

src/components/chip/chip.stories.tsx
kurtdoherty, line 1: **note:** I've covered the following behaviours of the Chip in these stories:

- Filter chip variant
- Selection chip variant
- Disabled chip
- Overflow/wrapping
- Truncation

src/components/chip/chip.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const tooltip = useTooltip()
+    return (
+      <>
+        <Chip {...args} {...tooltip.getTriggerProps()} />
+        <Tooltip description="Because reasons" position="top" {...tooltip.getTooltipProps()} />
+      </>
+    )
+  },
+}
+
```
kurtdoherty, line 72: **note:** this `useNarrowParentDecorator` decorator is used on the next two stories concerned with text overflow behaviour. The decorator is used to constrain the maximum width to which the Chip is permitted to grow.

tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "target": "ES2022",
     "declaration": true,
     "declarationMap": true,
```
kurtdoherty, line 19: **note:** This informs TS about the extra matchers on Vitest's `expect` utility

vitest.setup.ts
kurtdoherty, line 1: **note:** This will decorate Vitest's `expect` utility with the additional DOM-specific matchers.

src/components/chip/styles.ts
```diff
@@ -0,0 +1,90 @@
...
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+    background: var(--fill-default-lightest);
+
+    &:hover {
+      background: var(--fill-default-lightest);
+    }
+  }
+
+  &:focus-visible {
```
kurtdoherty, line 46: **note:** I've used `outline` and `outline-offset` here, mostly because the `outline` property is the official CSS property for "focus" outlines. Unfortunately, Figma doesn't support the `outline-offset` property, so Design cannot specify the focus styles in the same way, hence they use `box-shadow`.

Not 100% sure how I feel about this. On the one hand, I think `outline` is the "right tool for the job", but it means we're not 1-1 with the Design spec (and the current focus styles require a 3px thickness, for which we have no CSS variable).

ksolanki7 (approved):
This looks great.

PR #306, opened 2025-02-10
feat: #258 add new `ChipGroup` component

kurtdoherty:
### Context

- #303 deprecated this existing Chip (and Chip Group) component to make room for a new <button>-based version
- #304 adds a new Chip component

### This PR

- Adds a new ChipGroup component.

Screenshot 2025-02-10 at 10 05 47 AM

Screenshot 2025-02-11 at 12 33 36 PM

### DIFF
--- a/src/components/chip-group/chip-group-item.tsx
+++ b/src/components/chip-group/chip-group-item.tsx
@@ -1,0 +1,17 @@
+import { ElChipGroupListItem } from './styles'
+import { Chip } from '../chip/chip'
+
+import type { ComponentProps } from 'react'
+
+interface ChipGroupItemProps extends ComponentProps<typeof Chip> {}
+
+/**
+ * A thin wrapper around a chip to ensure it is rendered as a list item inside the chip group.
+ */
+export function ChipGroupItem(props: ChipGroupItemProps) {
+  return (
+    <ElChipGroupListItem>
+      <Chip {...props} />
+    </ElChipGroupListItem>
+  )
+}
--- a/src/components/chip-group/chip-group.stories.tsx
+++ b/src/components/chip-group/chip-group.stories.tsx
@@ -1,0 +1,114 @@
+import { ChipGroup } from './chip-group'
+import * as ChipStories from '../chip/chip.stories'
+
+import type { Decorator, Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Components/ChipGroup',
+  component: ChipGroup,
+  argTypes: {
+    children: {
+      control: 'radio',
+      defaultValue: 'Filter Chips',
+      options: ['Filter Chips', 'Selection Chips'],
+      mapping: {
+        'Filter Chips': (
+          <>
+            <ChipGroup.Item {...ChipStories.FilterChip.args}>Apples</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.FilterChip.args}>Bananas</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.FilterChip.args}>Oranges</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.FilterChip.args}>Peanuts</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.FilterChip.args}>Strawberries</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.FilterChip.args}>Watermelons</ChipGroup.Item>
+          </>
+        ),
+        'Selection Chips': (
+          <>
+            <ChipGroup.Item {...ChipStories.SelectionChip.args}>Red</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.SelectionChip.args}>Blue</ChipGroup.Item>
+            <ChipGroup.Item {...ChipStories.SelectionChip.args}>Yellow</ChipGroup.Item>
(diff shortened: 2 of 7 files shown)


src/components/chip-group/chip-group-item.tsx
kurtdoherty, line 1: **note:** Since a `ChipGroup` represents a collection of chips, the chip group will render a list of items, where each item will contain a chip. The chips cannot be list items themselves because they already have a semantically meaningful role (`button`). As such, we need to wrap chips in a `<li>` so that our chip group's list has list items.

The `ChipGroupItem` component being tested here is responsible for rendering a chip inside an `<li>`.

src/components/chip-group/chip-group.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const meta = {
+  title: 'Components/ChipGroup',
+  component: ChipGroup,
+  argTypes: {
+    children: {
+      control: 'radio',
+      defaultValue: 'Filter Chips',
+      options: ['Filter Chips', 'Selection Chips'],
+      mapping: {
```
kurtdoherty, line 22: **note:** The two groups of chips defined here match the groups shown in Figma

src/components/chip-group/chip-group.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+      control: 'radio',
+      options: ['scroll', 'wrap'],
+    },
+  },
+} satisfies Meta<typeof ChipGroup>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
```
kurtdoherty, line 55: **note:** This decorator is used to constrain the width of the chip group. We give it the same border that is used in Figma.

src/components/chip-group/chip-group.tsx
kurtdoherty, line 1: **note:** A chip group renders its children (chip group item's) inside a list.

src/components/chip-group/styles.ts
```diff
@@ -0,0 +1,32 @@
+import { styled } from '@linaria/react'
+
+export const ElChipGroup = styled.div`
+  display: block;
+
```
kurtdoherty, line 5: **note:** The [:has](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) selector allow us to style this `ElChipGroup` element if it has a child with the `data-overflow='scroll'` attribute. While this attribute is in the "Newly Available" baseline, it's been available in all modern browsers since December 2023, so it's pretty well supported.

src/components/chip-group/chip-group.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+ */
+export const ChipSizing: Story = {
+  argTypes: {
+    children: {
+      control: false,
+    },
+  },
+  args: {
+    children: (
+      <>
```
ksolanki7, line 100: @kurtdoherty **suggestion (non-blocking):** Using an additional fragment (`<>...</>`) is necessary to comply with React’s `render()` method, as it must return a **single** React element. However, in Storybook, this fragment is exposed in the Show Code section as `<React.Fragment>...</React.Fragment>`. To prevent this and provide a more user-friendly code display, we can leverage Storybook’s parameters to customize the code shown under Show Code.

eg: 
```
parameters: {
    docs: {
      source: {
        code: `
<ChipGroup overflow="scroll">
    <ChipGroupItem
      variant="filter"
    >
      Apples
    </ChipGroupItem>
    <ChipGroupItem
      variant="filter"
    >
      Bananas
    </ChipGroupItem>
    <ChipGroupItem
      variant="filter"
    >
      Oranges
    </ChipGroupItem>
    <ChipGroupItem
      variant="filter"
    >
      Peanuts
    </ChipGroupItem>
    <ChipGroupItem
      variant="filter"
    >
      Strawberries
    </ChipGroupItem>
    <ChipGroupItem
      variant="filter"
    >
      Watermelons
    </ChipGroupItem>
</ChipGroup>
        `,
      },
    },
  },
```

Ref: https://storybook.js.org/docs/6/writing-docs/doc-blocks/doc-block-source#working-with-the-docspage:~:text=of%20a%20button.-,Working%20with%20the%20DocsPage,-Storybook%20automatically%20generates

src/components/chip-group/styles.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+`
+
+interface ElChipGroupListProps {
+  'data-overflow': 'scroll' | 'wrap'
+}
+
+export const ElChipGroupList = styled.ul<ElChipGroupListProps>`
+  display: flex;
+  gap: 8px;
```
ksolanki7, line 17: @kurtdoherty **issue(non-blocking):** I believe we have a CSS variable available for spacing i.e for 8px it is `--spacing-2`

ksolanki7 (approved):

src/components/chip-group/chip-group.stories.tsx
kurtdoherty: Yeah, I'm aware of that ability, but it means we have to remember to maintain the code snippet manually when making changes to the story. This means they're all but guaranteed to diverge over time.

Instead, I've changed it to use an array instead of a fragment. This seems to net us the best of both worlds 👍

src/components/chip-group/styles.ts
kurtdoherty: Ah, good catch! Will fix.

PR #307, opened 2025-02-10
feat: #218 v5-table-component table-part-[01]

rpt-rfoxy:
# Pull request checklist

- [x] Create table Component  (table) part

schreenshoot
image


src/components/table/table/__tests__/table.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { render } from '@testing-library/react'
+import { Table } from '../index'
+import { TableToolbar } from '../../table-toolbar'
+import { TableContainer } from '../../table-container'
+describe('Table', () => {
+  test('should match snapshot', () => {
+    const { asFragment } = render(
```
kurtdoherty, line 7: **issue:** It's important to design our tests to fail well. You can read more about this [here](https://consolegroup.atlassian.net/wiki/spaces/DP/pages/2678587420/Writing+good+unit+tests#Designing-tests-that-fail-well). Right now, this test is will fail:

1. ✅ when our subject changes (`TableContainer`); and,
2. ❌ when one of the subject's collaborators change (e.g. `Table`)

Failure (2) is not what we want.

**suggestion:** To avoid this happening, we can simply change our test as follows:

```javascript
render(<TableContainer>Fake child</TableContainer>)
```

This way, we're only snapshotting the subject (and it's fake child), not the subject AND a bunch of collaborators. Each collaborator will have it's own tests to validate they function correctly.

src/components/table/table/table.mdx
```diff
@@ -0,0 +1,29 @@
...
+import { RenderHtmlMarkup } from '../../../storybook/render-html-markup'
+import * as TableStories from './table.stories'
+
+<Meta of={TableStories} />
+
+# Table
+
+
+The Table component can be used from the Table molecule components, e.g. `TableRow`, `TableCell`, etc
+There are two visual styles for the table, dependant on screen width.
```
kurtdoherty, line 11: **issue:** Again, I don't see anything about this kind of responsive behaviour on the Figma design spec.

src/components/table/table/styles.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { styled } from '@linaria/react'
+
+export const ElTable = styled.table`
+  width: 100%;
+  border-collapse: collapse;
+  cursor: pointer;
+  padding: 0.75rem 0.5rem;
+  ${isBelowDesktop} {
+    display: flex;
+  }
```
kurtdoherty, line 11: **issue:** I don't see anything about this kind of responsive behaviour on the Figma design spec. It's really important that we do not add complexity to our components if we're not required to by the design spec.

Unless I'm mistaken, "responsiveness" will be done by hiding/showing columns; it will not be done by adjusting the way a table's cells are laid out.

src/components/table/table/styles.tsx
```diff
@@ -0,0 +1,12 @@
+import { isBelowDesktop } from '../../../styles/media'
+import { styled } from '@linaria/react'
+
+export const ElTable = styled.table`
```
kurtdoherty, line 4: **issue:** I do not think we should be using the default `table` display mode. Instead, I think we should be using a `flex` or `grid` display mode. That said, I think this concern was missed on the solution design as I can't see any details around how we plan to size columns in the table correctly.

Might need further discussion on this one 🤔

src/components/table/table/table.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { FC, TableHTMLAttributes } from 'react'
+import { ElTable } from './styles'
+
+export interface ITableProps extends TableHTMLAttributes<HTMLTableElement> {}
```
kurtdoherty, line 4: **todo:** Can you please keep the names for prop types and interfaces consistent with the rest of the repo? Specifically, we don't use prefixes like `I` to denote an interface.

src/components/table/index.ts
kurtdoherty, line 1: **issue:** I don't think it was discussed in the solution design, but do we really need the nested folder structure here? Feels like we could just have something like:

```
table/
  __tests__/
    table.tests.tsx
    table-container.tests.tsx
  styles.tsx
  table.tsx
  table.stories.tsx
  table-container.tsx
  table-container.stories.tsx
  ...
```

src/components/table/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './table'
+export * from './table-container'
+export * from './table-toolbar'
```
kurtdoherty, line 3: **todo:** the table container and table toolbar components don't exist yet so should not be exported.

kurtdoherty (changes requested):
Thanks @rpt-rfoxy. I've left a few comments; I think there'll need to be a bit of conversation on some, particularly the file structure and `display` mode.

rpt-rfoxy:
thanks @kurtdoherty for the feedback, i'll change the code

src/components/table/index.test.ts
kurtdoherty, line 1: **question:** why are we testing our barrel file?

kurtdoherty (approved):
Thanks @rpt-rfoxy, just the one question, otherwise looks good.

src/components/table/index.test.ts
rpt-rfoxy: its because when we do test --coverage, the result  funcs always bellow 84%, so i inisiative to add test barrel file to keep the test function in 84%++

image

kurtdoherty: Haha, okay, that's just "gaming the system." We don't want to add tests for the sake of our coverage metrics; that approach will just lead to low-value, high-maintenance tests.

Instead of simply adding test coverage against a module that does not need tests, we need to assess whether there's any _meaningful_ coverage gaps being reported. Barrel files are not a meaningful gap.

To help with this, I'll make a change to our coverage settings to exclude barrel files. Once that's merged in, rebase you changes and see what your coverage comes back as for the components you're adding in this PR.

kurtdoherty: I've raised #315 👍

src/index.ts
```diff
@@ -60,8 +60,7 @@ export * from './components/dialog'
 export * from './components/avatar'
 export * from './components/skeleton'
 export * from './components/tooltip'
-export * from './components/table/table-container'
-export * from './components/table/table-toolbar'
```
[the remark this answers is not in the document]
kurtdoherty: I didn't notice this change sorry 🙈 I agree with you, removing these doesn't seem to be necessary.

src/components/table/index.test.ts
kurtdoherty: #315 has now been merged.

src/index.ts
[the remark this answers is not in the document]
rpt-rfoxy: thanks for the feedback, i already remove the barrel

kurtdoherty (approved):

kurtdoherty (approved):

PR #312, opened 2025-02-10
feat: 218 v5 table component table row part [04]

rpt-rfoxy:
# Pull request checklist

- [x] Create table cell Component (tr)

schreenshoot
image


src/components/table/table-row/table-row.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { ElTableRow } from './styles'
+
+export interface TableRowProps extends TableHTMLAttributes<HTMLTableRowElement> {
+  isSelected?: boolean
+  children: ReactNode
+}
+
+export const TableRow: FC<TableRowProps> = ({ children, isSelected, ...rest }) => {
+  return (
+    <ElTableRow role="row" {...rest} data-selected={isSelected}>
```
kurtdoherty, line 11: **todo:** I've mentioned this in the past, but please remember that our data attributes should match our prop names; `data-selected` -> `data-is-selected`

src/components/table/table-row/styles.tsx
```diff
@@ -0,0 +1,7 @@
+import { styled } from '@linaria/react'
+
+export const ElTableRow = styled.tr`
```
kurtdoherty, line 3: **todo:** I don't see any styling here for the row's selected state.

**todo:** Also, our `TableRow` component is passing a data attribute to this atom. That data attribute should be modelled here as an expected prop for `ElTableRow`.

That is, I would want to see:

```typescript
interface ElTableRowProps {
  'data-is-selected': boolean
}

export const ElTableRow = styled.tr<ElTableRowProps>`
  // ...
`
```

kurtdoherty (changes requested):
Thanks @rpt-rfoxy, just a few small changes 👍

src/components/table/table-row/styles.tsx
rpt-rfoxy: hi @kurtdoherty about row, @DanishAli-Reapit will have sparated branch to create row part, its include selection row

src/components/table/table-row/table-row.tsx
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @component
+ * @param {boolean} [isSelected] - Whether the row is selected.
+ * @param {ReactNode} [children] - The content inside the row.
+ * @returns {JSX.Element} A `<tr>` element with optional selection state.
+ *
+ */
+export const TableRow: FC<TableRowProps> = ({ children, isSelected, ...rest }) => {
+  return (
+    <ElTableRow role="row" {...rest} data-is-selected={isSelected}>
```
kurtdoherty, line 32: **nit-pick:** I think I've mentioned this in previous PRs, but we don't want to assign explicit `role` attributes to elements when we're using elements like `<tr>` that have an implicit role.

kurtdoherty (approved):

src/components/table/table-row/table-row.tsx
rpt-rfoxy: thanks @kurtdoherty code already updated as request

PR #311, opened 2025-02-10
feat: 218 v5 table component TableHead and TableHeaderCell part [3]

rpt-rfoxy:
# Pull request checklist

- [x] Create table head Component  (thead)
- [x] Create table head cell Component  (th)

**Detail as per issue below (required):**
as perdiscusstion here [Discussion](https://github.com/reapit/elements/discussions/201) 
schreenshoot
image

PR #313, opened 2025-02-10
feat: 218 v5 table component  table cell part [05]

rpt-rfoxy:
# Pull request checklist

- [x] Create table cell Component (td)

schreenshoot
image


src/components/table/table-cell/table-cell.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { FC, ReactNode, TdHTMLAttributes } from 'react'
+import { ElTableCell, ElTableCellContent, TableCellCSSProps } from './styles'
+
+export interface TableCellProps extends TdHTMLAttributes<HTMLTableCellElement> {
+  children?: ReactNode
+  alignment?: 'left' | 'center' | 'right'
+  width?: string
+  minWidth?: string
+  maxWidth?: string
```
kurtdoherty, line 9: **question:** where does the design specification show the need for `flexDirection` and `isFlexWrap`?

src/components/table/table-cell/table-cell.tsx
```diff
@@ -0,0 +1,65 @@
...
+  flexDirection?: 'column' | 'row'
+  isFlexWrap?: boolean
```
kurtdoherty, line 11: **issue:** I don't think the special CSS style props defined for `ElTableCell` should be exposed here. For `TableCell`, the only way consumers should be able to specify the min/max/fixed width of the cell is via the explicit props for those values, not `style`.

kurtdoherty (changes requested):
Thanks @rpt-rfoxy. It feels like you're still stepping beyond the design specification too often. I shouldn't see components be capable of things that design haven't specified. If you have opinions or ideas you'd like to share, by all means raise them on the Figma design spec for the component. And if you do add things into the code that is not specified in the design spec, please highlight it on the PR via a comment.

src/components/table/table-cell/table-cell.tsx
rpt-rfoxy: hi @kurtdoherty for this question we already discuss here [here](https://github.com/reapit/elements/pull/311#discussion_r1959161857) and its needit since we have limited style properties for <td> the i add new wrapper <div> inside the td to give flexibility styling.

and if the contentwrapper are necessery we can remove it and also remove **flexDirection** and **isFlexWrap**.

src/components/table/table-cell/table-cell.tsx
rpt-rfoxy: hi @kurtdoherty  thanks for the feedback,
**Notes**:  the minWidth and maxWidth also alignment property is sugested from this discussion 
image
and i just following the past discussion and for `flexDirection` and `isFlexWrap` we can remove it since its part of tablecellcontent

src/components/table/table-cell/table-cell.tsx
kurtdoherty: I'm confused. The two comments you have referred to don't seem to have anything to do with this component's need for `flexDirection` and `isFlexWrap`...

src/components/table/table-cell/table-cell.tsx
kurtdoherty: I'm confused here too. My comments above are focused on your use of the `TableCellCSSProps` type for this React component's `style` prop. This `TableCellCSSProps` type includes CSS variable property definitions; these should not be part of `TableCell`'s prop interface, instead they should only be part of `ElTableCell`'s prop interface.

src/components/table/table-cell/styles.tsx
```diff
@@ -0,0 +1,74 @@
+import { styled } from '@linaria/react'
+import { CSSProperties } from 'react'
+
+export interface TableCellCSSProps extends CSSProperties {
```
kurtdoherty, line 4: **issue:** please remove the export of this interface and rename the interface as `ElTableCellCSSProperties`

src/components/table/table-cell/table-cell.tsx
rpt-rfoxy: aah i see, im sorry my bad, i'll remove the style tyle, 
i **thougt** since the <td> have child  <ElTableCellContent> so i thougt it will be good if `style` props will ocused only for `ElTableCellContent`.

now its clear, i'll remove it

**question**: should we remove ElTableCellContent or just keep it?

src/components/table/table-cell/styles.tsx
foxdreamstudio: image
its laready ermoved

src/components/table/table-cell/table-cell.tsx
foxdreamstudio: hi @kurtdoherty i already cahnge the code, and keep the props as DOC from [here](https://github.com/reapit/elements/discussions/201#discussion-7461375)

rpt-rfoxy: hi @kurtdoherty i already cahnge the code, and keep the props as DOC from [here](https://github.com/reapit/elements/discussions/201#discussion-7461375)

src/components/table/table-cell/styles.tsx
rpt-rfoxy: image
its laready ermoved

kurtdoherty (approved):
Thanks @rpt-rfoxy. I still don't understand the need for `ElTableCellContent` and I don't think you've provided a clear demonstration for _why_ it's necessary, but I'm pretty fatigued from the number rounds all these table-related PRs have taken so I think the best path forward is to just get this merged in and move on.

PR #320, opened 2025-02-19
feat: #154 v5 new SideBar (Collapse and Expand feature)

nurm717123:
## Context
In order to create a new SideBar for the v5 release, this PR, as part of #154, initended to add Collapse and Expand feature in SideBar.

## Pull request checklist
- [x] SideBarMenuItem collapsed/expanded style
- [x] SideBar collapse/expand hook
- [x] SideBar collapse/expand component

fixes: #

sidebar collapse


src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
kurtdoherty, line 1: **question:** Can you help me understand why you decided to place the collapse button implementation here (within `components/side-bar-menu-item`) and not in it's own folder (e.g. `components/side-bar-collapse-button`)?

src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
```diff
@@ -0,0 +1,20 @@
...
+  ElSideBarCollapseButtonContainer,
+  ElSideBarCollapseText,
+  elSideBarMenuItemAnchor,
+} from './styles'
+
+export const SideBarCollapse = () => {
+  const { getTriggerProps, isOpen } = useSideBarContext()
+  return (
+    <ElSideBarCollapseButtonContainer data-is-expanded={isOpen}>
+      <ElSideBarCollapseButton className={elSideBarMenuItemAnchor} {...getTriggerProps()}>
```
kurtdoherty, line 14: **question:** Does the design specification create this explicit relationship between the side bar menu items and the collapse button? Or do they just look similar and you've decided to reuse the styles?

**issue:** I also find it awkward that we're using a class with `Anchor` in it's name on a `<button>` element. That feels like a clear signal that something with our code design is not quite right.

src/components/side-bar-menu-item/styles.ts
```diff
@@ -51,3 +52,42 @@ export const elSideBarMenuItemAnchor = css`
...
+`
+
+export const ElSideBarCollapseButton = styled.button`
+  background: var(--fill-white);
+  border: none;
+  cursor: pointer;
+
+  border-radius: var(--corner-lg);
+  display: flex
+    flex-direction: row;
```
kurtdoherty, line 73: **todo:** `display: flex` is missing a semi-colon (`;`) which is throwing the indentation out of whack.

src/components/side-bar/__test__/__snapshots__/side-bar-context.test.tsx.snap
kurtdoherty, line 1: **question:** what value is this test providing?

src/components/side-bar/side-bar.tsx
```diff
@@ -1,16 +1,27 @@
...
+import { SideBarProvider, useSideBarContext } from './side-bar-context'
 import { ElSideBar, ELSideBarMenuList } from './styles'
 
 type SideBarFC = FC<HTMLAttributes<HTMLElement>> & {
-  MenuList: typeof ELSideBarMenuList
+  CollapseButon: typeof SideBarCollapse
+  MenuList: FC<HTMLAttributes<HTMLUListElement>>
 }
 
 const SideBar: SideBarFC = ({ children, ...props }) => (
```
kurtdoherty, line 11: **issue:** I'm finding the `useSideBar` hook and it's use by `SideBarProvider` a little over-engineered given they are only facilitating a single `isExpanded` piece of state. I think you could simplify the implementation greatly by doing the following:

```tsx
// side-bar.tsx
const SideBar = (...) => {
  const [isExpanded, setIsExpanded] = useState(false)
  return (
    <IsSideBarExpandedContext.Provider value={{ isExpanded, setIsExpanded }}>
      <ElSideBar ...>{children}</ElSideBar>
    </IsSideBarExpandedContext.Provider>
  )
}

// is-side-bar-expanded-context.ts
interface IsSideBarExpandedContextValue {
  isExpanded: boolean
  setIsExpanded: Dispatch<SetStateAction<boolean>>
}
export const IsSideBarExpandedContext = createContext<IsSideBarExpandedContextValue | undefined>()

export const useIsSideBarExpandedContext = () => { ... }

// side-bar-collapse-button.tsx
const SideBarCollapseButton = (...) => {
  const { isExpanded, setIsExpanded } = useIsSideBarExpandedContext()
  return ...
}
```

Later, when you go to add in the keyboard navigation handling, that can all be implemented in a unit separate to the expanded/collapsed UI state.

src/components/side-bar/use-side-bar.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface useSideBar {
+  getTriggerProps: (props?: Partial<HtmlButtonAttr>) => HtmlButtonAttr
+  isOpen: boolean
+  openSideBar: VoidFunction
+  closeSideBar: VoidFunction
+}
+
+export const useSideBar = () => {
+  const [isOpen, setIsOpen] = useState(true)
```
kurtdoherty, line 13: **issue:** since the mental model for the side bar is expanded vs collapsed, we should be using that terminology instead of open vs closed.

src/components/side-bar/__test__/side-bar-context.test.tsx
```diff
@@ -0,0 +1,19 @@
+import { render } from '@testing-library/react'
+import { SideBarProvider, useSideBarContext } from '../side-bar-context'
+
+describe('SideBarProvider and useSideBarContext', () => {
+  it('should render SideBarProvider and match snapshot', () => {
+    const { asFragment } = render(<SideBarProvider>Foo</SideBarProvider>)
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('should throw an error when useSideBarContext is used outside of SideBarProvider', () => {
```
kurtdoherty, line 10: **question:** Does the version of `@testing-library/react` we're using have the `renderHook` utility available? If so, that would make this test a bit simpler.

src/components/side-bar/styles.ts
```diff
@@ -21,4 +22,11 @@ export const ELSideBarMenuList = styled.ul`
   & li {
     width: 100%;
   }
+
+  &[data-is-expanded='false'] {
+    width: fit-content;
+    ${ElSideBarMenuItemText} {
```
kurtdoherty, line 28: **issue:** I feel like this should be handled by the `ElSideBarMenuItemText` atom, not `ElSideBar`.

kurtdoherty (changes requested):
Thanks for you patience with this one @nurm717123 🙏 I've left a few questions and identified some things I think need to change. Happy to discuss further if you disagree with my suggestions 👍

src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
nurm717123: the main reason is because it has the same style with `SideBarMenuItem`, the different is the `CollapseButton` doesn't have right icon, I wonder if its need to be separated 🤔 

question: should we provide separate stories for each small components which will only used in the bigger component e.g `CollapseButton` to `SideBar`

src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
nurm717123: > question: Does the design specification create this explicit relationship between the side bar menu items and the collapse button? Or do they just look similar and you've decided to reuse the styles?

no, its not explicitly defined, but the styles looks quite the same. so I think that we can reuse it

> issue: I also find it awkward that we're using a class with Anchor in it's name on a <button> element. That feels like a clear signal that something with our code design is not quite right.

yeah, for the reminder, we provide that classname just for `react-router` consumer

src/components/side-bar-menu-item/styles.ts
nurm717123: 👍

src/components/side-bar/__test__/side-bar-context.test.tsx
nurm717123: yeah, I will try that 👍

src/components/side-bar/styles.ts
nurm717123: got it

src/components/side-bar/side-bar.tsx
nurm717123: its will be new pattern as I just make it the same as `useMenu` pattern, but I think it will works, I will try it 👍

src/components/side-bar/__test__/__snapshots__/side-bar-context.test.tsx.snap
nurm717123: previously just to ensure its not crashed 😄

src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
kurtdoherty: Unless Design make an explicit relationship in a design spec, we should not be coupling components via shared styles. Since Design have modelled the collapse button as an independent component to the side bar menu items, we should also model them as independent components in our code.

src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
kurtdoherty: > yeah, for the reminder, we provide that classname just for react-router consumer

Yep, I understand the context, but either that class name is poorly named (i.e. is not specific to anchors), or we're misusing it here. Either way, something needs to change.

nurm717123: apologize, turns out the collapse button have a relationship with the sidebar menu items

image

when you click the CollapseButton's `layers.label.item` in figma, the component indicate thats it has the link to the menu item, yesterday I click only either its container or its text that's why the indicator doesn't show up 🤦 

given this situation I think I will make it still under `side-bar-menu-item` and make it use shared style instead of shared class, how do you think @kurtdoherty ?

src/components/side-bar-menu-item/side-bar-menu-item.molecules.tsx
kurtdoherty: Hmm, side bar menu items for us are anchor elements, not buttons. I think what you've suggested in the other comment thread is best for now: we treat the collapse button and side bar items as completely separate components.

nurm717123: got it 👍

src/components/side-bar/styles.ts
```diff
@@ -18,7 +18,11 @@ export const ELSideBarMenuList = styled.ul`
   // TODO: remove default value once the token is available
   width: var(--size-64, 256px);
 
-  & li {
```
nurm717123, line 21: without "<" would still work for now, just to make it save here, as in future there will be a nested list (nested menu item)

kurtdoherty (approved):

PR #323, opened 2025-02-20
feat: #138 init new breadcrumb

nurm717123:
## Context

As there is new v5 breadcrumb #138 that will replace the previous (v4) breadcrumb component. this pr is to create the new Breadcrumbs based on a POC that has been discussed [here](https://github.com/reapit/elements/issues/138#issuecomment-2592060984)

note: depends on #322 

# Pull request checklist
[x] Breadcrumb component
[x] Desktop and Mobile overflow behaviour
 
breadcrumb


src/components/breadcrumbs/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  &:hover {
+    text-decoration: underline;
+  }
+`
+
+export const ElBreadcrumbs = styled.ul`
+  display: flex; /* Makes the list horizontal */
+  list-style: none;
+  overflow: visible !important;
```
nurm717123, line 58: so that the `:focus` style (shadow) is visible

.eslintrc.cjs
kurtdoherty, line 1: **note:** same here. Can you please provide some context for why this change is being made.

.editorconfig
```diff
@@ -10,4 +10,4 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 
 [*.md]
```
kurtdoherty, line 12: **note:** can you please provide some context for this change? Generally, we should avoid changes like this that are completely unrelated to the topic of the PR.

.yarnrc.yml
```diff
@@ -4,8 +4,8 @@ enableGlobalCache: false
 
 nodeLinker: node-modules
 
-npmAuthToken: "${NPM_TOKEN}"
+# npmAuthToken: "${NPM_TOKEN}"
```
kurtdoherty, line 7: **issue:** This doesn't seem to be something we should be doing. Can you explain why this change has been made, including why you needed to turn this off in the first place?

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type BreadcrumbsFC = FC<HTMLAttributes<HTMLUListElement>> & {
+  Item: FC<HTMLAttributes<HTMLLIElement>>
+  Link: FC<AnchorHTMLAttributes<HTMLAnchorElement>>
+}
+
+/**
+ * Breadcrumbs are used to indicate to the user their flow in the application
+ * and provide a navigation back step to previous pages.
+ */
```
kurtdoherty, line 12: **issue:** Can you help me understand why `ElBreadcrumbs` is a `ul` and not the `nav`? Feels like we may be missing an atom such as `ElBreadcrumbsList` 🤔

src/components/breadcrumbs/styles.ts
kurtdoherty: **question:** why do you need the `!important` here?

vite.config.ts
```diff
@@ -95,4 +95,7 @@ export default defineConfig({
     include: ['src/**/*.test.ts?(x)'],
     setupFiles: ['vitest.setup.ts'],
   },
```
kurtdoherty, line 97: **issue:** why is this change necessary? Why do we want to turn off `hmr` in Vite's dev server?

kurtdoherty (changes requested):
Thanks @nurm717123. A few issues/questions to discuss.

Overall, if you aren't already doing so, I'd encourage you to perform your own review of your PRs before you add other reviewers. If you do that with the correct mindset (i.e. if you look at the changes from your reviewer's perspective), I think that will help you communicate the reason behind certain changes a lot better. I will also help you identify changes that you've accidentally included.

vite.config.ts
nurm717123: sometimes I got an error after the fast refresh, so I just add it, I don't mean it to be part of the pr sorry 🙏

.yarnrc.yml
nurm717123: I don't mean it to be part of the pr sorry 🙏

.eslintrc.cjs
nurm717123: I don't mean it to be part of the pr 🙏

.editorconfig
nurm717123: looks like I've accidentally add some of unrelated changes 🙏

src/components/breadcrumbs/breadcrumbs.tsx
nurm717123: I think that to include the nav into `Breadcrumb` is simpler as it doesn't have additional style, so that the consumer just need to remember 3 item, i.e `Breadcrumbs`, `Breadcrumbs.Item`, and `Breadcrumbs.Link`, we did that too in `TopBar` where `nav` is not part of the style 🤔 

image

src/components/breadcrumbs/styles.ts
nurm717123: right, after I tried it again, we don't even need that, it might because to fix false issue at that time when I work on it

I will add a placeholder page content example to storybook to make it better to test the ui, thanks 👍

nurm717123:
Thanks @kurtdoherty for the review again, I've resolve all the feedbacks except 1 related to the `nav` (we can continue this on the thread)

in addition I've updated the storybook to have a page content placeholder to make the test more valid

updated storybook breadcrumb

>Overall, if you aren't already doing so, I'd encourage you to perform your own review of your PRs before you add other reviewers. If you do that with the correct mindset (i.e. if you look at the changes from your reviewer's perspective), I think that will help you communicate the reason behind certain changes a lot better. I will also help you identify changes that you've accidentally included

apologise for some of unrelated file changes 🙏 , I will be more careful in the future.

src/components/breadcrumbs/breadcrumbs.tsx
kurtdoherty: Yeah okay. I think you're right that REact consumers should probably only have to deal with `Breadcrumbs`, `Breadcrumbs.Item`, and `Breadcrumbs.Link`, but what I find awkward about the current implementation of `Breadcrumbs` is the prop interface communicates we are rendering a `<ul>` element, but that's not actually the root element we render.

Generally speaking, I think our components should be typed according to the primary semantic element they render as. For the Breadcrumbs, this feels like it should be the `<nav>`, not the `<ul>` (I think we'll more likely want to apply additional attributes to the nav element, not the list element). For an Input component, the primary element would be the `<input>` element, not any surrounding `<div>` or `<label>`.

This makes me think `ElBreadcrumbs` should actually be `ElBreadcrumbsList` and then we have a new `ElBreadcrumbs` atom that is just a styled `<nav>` element (with no actual styles needed). This `Breadcrumbs` component can then extend `HTMLAttributes<HTMLElement>` instead of `HTMLAttributes<HTMLUListElement>`.

Does that sound reasonable?

nurm717123: Ok, I guess I got it now, i.e the name of the component should represent the semantic component it render.  I will rethink about this to apply in side-bar's pr too, and I will update this pr. Thanks for the explanation.

kurtdoherty: I'm not sure there's any changes needed in the sidebar work 🤔

nurm717123: After I re-review my sidebar pr, turns out it already correct about this 😄

kurtdoherty: @nurm717123, I think there's been some miscommunication here, sorry! My comments above were trying to explain the following:

1. I agree with you about the number of React components consumers should have to deal with;
2. Felt like `ElBreadcrumbs` should have been `ElBreadcrumbsList` and that we were missing `ElBreadcrumbs` as a styled `<nav>`
3. The `Breadcrumbs` React component should probably be typed based on the `<nav>` element, not the `<ul>` element.

You've done (2) and (3) well, but I think (1) has been misunderstood. All I was suggesting was this:

```jsx
const Breadcrumbs = ({ children, ...props }) => {
  return (
    <ElBreadcrumbs {...props}>
      <ElBreadcrumbsList>
        {children}
      </ElBreadcrumbsList>
    </ElBreadcrumbs>
  )
}
```

Sorry for all the confusion. If you just want to move on, we can; I don't think it's a big deal for consumers to have to manually compose `Breadcrumbs.List`.

kurtdoherty (approved):

PR #324, opened 2025-02-24
feat: #238 Added support conditional displaying tooltip on truncated text element

ksolanki7:
# Pull request checklist

- Added support for conditionally displaying the tooltip for truncated text.

**Note:** I have provide inline comments to explain the approach and some recommendation of what needs to change and how to utilise this conditional support for internal elements.

### DIFF
--- a/src/components/tooltip/tooltip.mdx
+++ b/src/components/tooltip/tooltip.mdx
@@ -1,4 +1,4 @@
-import { Meta, Canvas, Controls, Description } from '@storybook/blocks'
+import { Meta, Canvas, Controls, Description, Source } from '@storybook/blocks'
 import { RenderHtmlMarkup } from '../../storybook/render-html-markup'
 import * as TooltipStories from './tooltip.stories'
 
@@ -12,5 +12,31 @@ import * as TooltipStories from './tooltip.stories'
 
 <Controls />
 
+# Basic Usage
+
 <Description of={TooltipStories.BasicUsage} />
 <Canvas of={TooltipStories.BasicUsage} />
+
+# Conditional Display
+
+<Description of={TooltipStories.ConditionalDisplay} />
+<Canvas of={TooltipStories.ConditionalDisplay} />
+
+The following example demonstrates how to truncate text using CSS.
+<Source code={`
+// HTML
+<div class="container">
+  <span class="truncate-text">Text is truncated here</span>
+</div>
+
+// CSS styles
+.container {
+  width: '136px';
+  display: 'flex'; // Can use grid
+}
+.truncate-text {
+  overflow: 'hidden';
+  text-overflow: 'ellipsis';
+  white-space: 'nowrap'
+}
+`} language="tsx" />
--- a/src/components/tooltip/use-tooltip.ts
+++ b/src/components/tooltip/use-tooltip.ts
@@ -1,12 +1,34 @@
-import { HTMLAttributes, useEffect, useState } from 'react'
+import { HTMLAttributes, useEffect, useRef, useState } from 'react'
 import { useId } from '#src/storybook/random-id'
 
 export const useTooltip = () => {
-  const tooltipId = `tooltip-id-${useId()}`
+  const tooltipId = useRef(`tooltip-id-${useId()}`).current
+  const truncatedElementTooltipId = useRef(`truncation-target-${useId()}`).current
   const [isVisible, setIsVisible] = useState(false)
 
-  const show = () => setIsVisible(true)
+  const show = (shouldCheckTruncation: boolean) => {
+    if (checkTextTruncation(shouldCheckTruncation)) {
+      setIsVisible(true)
+    }
+  }
   const hide = () => setIsVisible(false)
+
+  const checkTextTruncation = (shouldCheckTruncation: boolean) => {
+    const triggerElement = document.querySelector<HTMLElement>(`[data-visible-id="${tooltipId}"]`)
+
+    if (!shouldCheckTruncation) {
+      return true // Always show tooltip if truncation check is disabled
+    }
+
+    // Check if the trigger itself has `data-will-truncate="true"`
+    // OR if one of its descendants has it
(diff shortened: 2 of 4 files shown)


src/components/tooltip/tooltip.mdx
```diff
@@ -12,5 +12,31 @@ import * as TooltipStories from './tooltip.stories'
...
+
 <Description of={TooltipStories.BasicUsage} />
 <Canvas of={TooltipStories.BasicUsage} />
+
+# Conditional Display
+
+<Description of={TooltipStories.ConditionalDisplay} />
+<Canvas of={TooltipStories.ConditionalDisplay} />
+
+The following example demonstrates how to truncate text using CSS.
```
ksolanki7, line 25: This is to educate the user on truncating the text using CSS.

src/components/tooltip/use-tooltip.ts
```diff
@@ -134,5 +153,11 @@ export const useTooltip = () => {
     isVisible,
   })
 
-  return { getTriggerProps, getTooltipProps }
+  const getTruncationTargetProps = (props?: HTMLAttributes<HTMLElement>) => ({
+    ...props,
+    id: truncatedElementTooltipId,
+    'data-will-truncate': true, // boolean
```
ksolanki7, line 159: This will add the attribute `data-will-truncate` which can be utilized by internal elements to truncate the text of the target using CSS.

ksolanki7:
@kurtdoherty Thank you for your feedback—I truly appreciate it. I’ll take some time to review and digest your suggestions, and then I’ll get back to you after running a POC. I also encountered some reference issues with other solutions I tested. That was the reason I ended up with this solution. 
I'll explore this further and follow up soon.

ksolanki7:
@kurtdoherty Thank you for looking into this PR. I have updated the PR with solution 2 which is more robust and requires fewer consumer side changes.

ConditionalTooltip

src/components/tooltip/use-tooltip.ts
```diff
@@ -1,13 +1,30 @@
...
+  truncationTargetId?: string
+}
+
+export const useTooltip = ({ truncationTargetId }: UseTooltipOptions = {}) => {
+  const tooltipId = useRef(`tooltip-id-${useId()}`).current
   const [isVisible, setIsVisible] = useState(false)
 
-  const show = () => setIsVisible(true)
+  const show = () => {
+    if (!truncationTargetId || checkTextTruncation()) {
```
kurtdoherty, line 13: **suggestion:** we're checking the existence of `truncationTargetId` here, but we're also checking it in `checkTextTruncation`. Feels like we should do it one place, not both?

src/components/tooltip/use-tooltip.ts
```diff
@@ -1,13 +1,30 @@
...
+      setIsVisible(true)
+    }
+  }
   const hide = () => setIsVisible(false)
 
+  const checkTextTruncation = () => {
```
kurtdoherty, line 19: **suggestion:** `checkTextTrunction` feels a little vague and too much like a command (i.e. do something) rather than a query (i.e. tell me something). A name like `isTargetTruncated` could be better?

src/components/tooltip/use-tooltip.ts
```diff
@@ -1,13 +1,30 @@
...
+    if (!truncationTargetId) return false // If no truncation ID is provided, assume no truncation
+
+    const target = document.getElementById(truncationTargetId)
+    if (!target) return false // // If the target element is not found, no truncation
+
+    return target.scrollWidth > target.clientWidth // Check if text is truncated
```
kurtdoherty, line 25: **thought:** some of the comments here feel a little unnecessary to me as they're just restating what the code is already doing. I think it would be worth focusing on the "why" more. For example, why does the `scrollWidth` being larger than the `clientWidth` imply truncated text?

kurtdoherty (approved):

src/components/tooltip/use-tooltip.ts
ksolanki7: Makes sense, I'll provide more context reasoning why than what.

src/components/tooltip/use-tooltip.ts
ksolanki7: Sure, will rename this. Good catch.

ksolanki7: @kurtdoherty I'll rename it to a more appropriate name `shouldShowTooltip`, as the function determines whether to show the tooltip rather than merely checking the target truncation.

src/components/tooltip/use-tooltip.ts
ksolanki7: @kurtdoherty I agree, unnecessaryly added a check where I can manage this in `checkTextTruncation`. I'll have amended this. Please review it.

src/components/tooltip/use-tooltip.ts
```diff
@@ -1,13 +1,47 @@
...
+    if (!truncationTargetId) return true
+
+    const target = document.getElementById(truncationTargetId)
+    /**
+     * Fallback mechanism:
+     * If the target element is not found in the DOM, return `true` to ensure
+     * the Tooltip is displayed by default.
+     */
+    if (!target) return true
+
```
kurtdoherty, line 31: **note:** the JS doc comment below reads like it's an overall comment for the `shouldShowTooltip` function. Not sure if you intentionally put on the function's return statement or not, but I'm not sure it works that way...

kurtdoherty (approved):

src/components/tooltip/use-tooltip.ts
ksolanki7: Oops, nice catch. My bad!
I initially updated it as an inline comment, made it JSDOC, and staged it before moving it to the function. I'll push my uncommitted changes. 

This is what I call a terrible result of context switching. Thank you again for picking it up.

ksolanki7:
@kurtdoherty I'll go ahead and merge this PR now.

Issue #325, opened 2025-02-24
v5 Refactor: Button component - Disabled button still allowing onClick


ksolanki7:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behaviour eg:
1. Go to '[Disabled Button](https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-button--button-disabled)'
2. Click on 'Button Text' button
3. Check the Actions tab in Storybook
4. See error onClick eventhandler being triggered

**Expected behaviour**
Disabled button shouldn't trigger onClick event.

PR #327, opened 2025-02-25
fix: #325 Fix issue with disabled button triggering onClick event

ksolanki7:
# Pull request checklist

- Rename the `disabled` prop to `isDisabled`
- Fix the issue with a disabled button which was triggering the `onClick` event.
- Added unit test to test disabled functionality over Button component


src/components/button/button.tsx
```diff
@@ -147,22 +147,24 @@ export const Button: FC<ButtonProps> = (props) => {
   } else {
     return (
       <ElButton
+        {...(rest as ButtonHTMLAttributes<HTMLButtonElement>)}
```
ksolanki7, line 150: The issue was due to the ordering of props; the ...rest spread was overriding the onClick prop specified in the component declaration.

src/components/button/button.tsx
```diff
@@ -147,22 +147,24 @@ export const Button: FC<ButtonProps> = (props) => {
...
         data-variant={variant}
         data-has-no-padding={hasNoPadding}
         className={combinedClassName}
+        aria-label={ariaLabel}
+        aria-disabled={isDisabled}
+        role="button"
         onClick={(e) => {
```
kurtdoherty, line 157: **suggestion:** would be good to make this look exactly the same as what I did for `Chip`...

```typescript
const handleClick = useCallback<MouseEventHandler<HTMLButtonElement>>(
  (event) => {
    // We are not using <button>'s `disabled` attribute because disabled buttons are bad for a11y.
    // Rather, we keep the <button> enabled and available in the a11y tree, but mark it as disabled using
    // `aria-disabled`. This means click events will still be fired, so we need to prevent any default action
    // for the button from occuring, stop it propagating to ancestors and avoid calling the consumer-supplied
    // `onClick` callback.
    if (isDisabled) {
      event.preventDefault()
      event.stopPropagation()
      return
    }
    onClick?.(event)
  },
  [isDisabled, onClick],
)
```

kurtdoherty (approved):

src/components/button/button.tsx
ksolanki7: @kurtdoherty **question:** Do you think we should extract this event handler handling the disable functionality into utils, so that it can be used across the elements project?

kurtdoherty: Probably, but I don't want to have to think about that right now 😅

src/components/button/button.tsx
```diff
@@ -120,23 +130,37 @@ export const Button: FC<ButtonProps> = (props) => {
...
   const combinedClassName = cx(className, sizeClass, miscellaneousClass)
 
+  const handleClick = useCallback(
+    (event: MouseEvent<HTMLButtonElement> | MouseEvent<HTMLAnchorElement>) => {
+      // We are not using <button>'s `disabled` attribute because disabled buttons are bad for a11y.
+      // Rather, we keep the <button> enabled and available in the a11y tree, but mark it as disabled using
+      // `aria-disabled`. This means click events will still be fired, so we need to prevent any default action
+      // for the button from occuring, stop it propagating to ancestors and avoid calling the consumer-supplied
+      // `onClick` callback.
+      if (isDisabled && isButtonAsButtonElement(props)) {
```
ksolanki7, line 140: @kurtdoherty Can I get this reviewed again.

Added conditional check `isButtonAsButtonElement(props)` to check that it is a native button and not an anchor explicitly.

**Note:** `isDisabled` prop is set to `never` type for anchor, so IDE will prompt TS error anyways for this prop. I have added the above check so that button as anchor still allows onClick regardless of `isDisabled` prop on the safer side.

kurtdoherty: Ahh, great pick up with the button vs anchor difference here 👏

kurtdoherty (approved):

Issue #330, opened 2025-03-05
Make new design tokens available


kurtdoherty:
The Design System is refactoring all its variables.

We need to get those variables available to our v5 components. We also need to prepare the ground for the introduction of a first-class PayProp theme.

### Scope

1. Get the updated variables exported from Figma and applied to Elements;
2. Get the new variables included in Elements' CSS;
3. Implement a new, simpler HTML/CSS-first approach to selecting a theme.

**note:** Ideally, we would have the Reapit/PayProp theming be a build-time concern. At this stage, it is not clear how we would implement this for Elements given our use of Linaria. As such, we will accept the fact that our CSS artefact will be slightly bloated with CSS variable definitions products will simply never need.

kurtdoherty:
**note:** Token Studio could not be used because it does not appear to work with Figma file branches (without paying for a pro licence). As such, the variable exports will be done via a different Figma plugin—Design Tokens Manager—which provides JSON files that adhere to [Design Tokens Format Module](https://tr.designtokens.org/format/). This format is supported by `style-dictionary` and therefore fits well with Elements' existing design token build pipeline (though we need to upgrade [style-dictionary@4](https://styledictionary.com/version-4/migration/)).

**note:** The new approach I'll be presenting for theming an application is demonstrated, via it's most basic HTML/CSS-only form, in [CSS theming via data-attribute](https://codesandbox.io/p/sandbox/css-theming-via-data-attribute-6rkygv) CodeSandbox. Any React components that we provide through Elements to set the theme will simply write the theme name to the `data-theme` attribute on the document's root element, as shown in the POC.

kurtdoherty:
Closed by #331

PR #331, opened 2025-03-05
chore: #330 Add new design tokens, new theme provider and upgrade style-dictionary

kurtdoherty:
### Context

The Design System is refactoring all its variables.

We need to get those variables available to our v5 components. We also need to prepare the ground for the introduction of a first-class PayProp theme.

### This PR

**resolves:** #330 

- Adds new JSON files for the exported variables from Figma;
- Updates the `tokens/build.mjs` script we use to build the CSS variables output files. In doing this:
  - I've removed the TS output format as there's no known consumers of this format. Console Cloud may need it in the future, but not yet, and when we do, I'm pretty confident we'll want it in a different TS/JS format that what's being generated at the moment.
  - I've changed the output folder for the CSS artefacts built by `style-dictionary` to be called `dist`. To allow this folder to still be tracked by Git, I've also updated our `.gitignore` file.
- Deprecates the existing theming provider component and its related documentation; and,
- Implements a new, HTML/CSS-first approach to selecting a theme.

**note:** The line count for this PR is inflated by the two new JSON files with our token data exported from Figma, as well as the yarn.lock changes due to the upgrade of `style-dictionary`

**note:** While we support a PayProp theme, there variables for that theme have not yet been defined in Figma. As such, opting to apply that theme will not currently result in any visual changes.

Screenshot 2025-03-05 at 1 07 09 PM


src/components/theme-provider/use-theme.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Sets the theme for the current document. Note, this will override any `data-theme` value
+ * that may already exist on the document's root element.
+ *
+ * @param theme The theme the current document should use.
+ */
+export function useTheme(theme: Theme) {
+  useLayoutEffect(
+    function setThemeForDocument() {
```
kurtdoherty, line 12: **note:** See [MDN's docs for documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement) for more context. The TLDR is that this is the root element of any document. For HTML documents, this will be the `<html>` element

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
-import StyleDictionaryPackage from 'style-dictionary'
+// @ts-check
```
kurtdoherty, line 2: **note:** I've turned TS type checking on in this file via the `@ts-check` directive above. This allows TS to provide type checking (at least in the IDE) when working in this file based on the JS Doc comments. That is why I've also added the JS Doc comments on some of the variables/functions in this file.

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
...
-const getStyleDictionaryConfig = (theme) => {
-  const lowerCasedTheme = theme.toLowerCase()
+import StyleDictionary from 'style-dictionary'
+
+/** @type {Theme[]} */
+const themes = ['payprop', 'reapit']
+
+/** @returns {Config} */
+function getConfig(themeName) {
   return {
```
kurtdoherty, line 15: **note:** Instead of a single `tokens.json` file, we now have two—`Primitives.Value.tokens.json` and `Semantics.Reapit.tokens.json`—that we source the design tokens from. Both of these files use follow the [Design Tokens Format Module](https://tr.designtokens.org/format/) spec and were obtained using the Design Tokens Manager plugin via Figma (i.e. not Tokens Studio).

Because these files follow the DTFM spec and style-dictionary supports that spec, we no longer need the custom parser.

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
...
-            console.log(error)
-          }
-        },
-      },
+    source: [
+      './src/tokens/Primitives.Value.tokens.json',
+      `./src/tokens/Semantics.${themeName.toUpperCase()}.tokens.json`,
     ],
     platforms: {
       css: {
```
kurtdoherty, line 21: **note:** We now output to a `dist` folder with each file named according to the theme.

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
...
-        buildPath: `src/tokens/${lowerCasedTheme}/`,
+        buildPath: 'src/tokens/dist/',
         files: [
           {
-            destination: 'tokens.css',
+            destination: `${themeName}.css`,
```
kurtdoherty, line 25: **note:** As per the code comment, we only output the semantic tokens, not the primitives.

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
...
+              // We only want to include semantic tokens in our output because they are resolved to explicit values
+              // instead of referencing the primitives. This is important because we do not want to expose the
+              // primitives to consumers, (1) to prevent their misuse, and (2) to minimise the number of CSS variables
+              // at play.
+              return token.filePath.includes('Semantics')
+            },
             format: 'css/variables',
           },
         ],
-      },
```
kurtdoherty, line 37: **note:** We no longer output TS files. We may introduce some kind of JS/TS output back in later if Console Cloud actually needs it.

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
...
-      ts: {
-        buildPath: `src/tokens/${lowerCasedTheme}/`,
-        files: [
-          {
-            destination: 'tokens.ts',
-            format: 'javascript/es6',
-          },
-        ],
-        transforms: ['name/cti/camel'],
+        options: {
```
kurtdoherty, line 36: **note:** we customise the selector written to the CSS output file based on the theme. By default, the selector used is `:root`, but we want the Reapit theme to use `:root, :root[data-theme="reapit"]` and the PayProp theme to use `:root[data-theme="payprop"]` because we're including both themes within the one CSS file for Elements.

In future, we'll want to shift the theming to be a build-time concern, instead of a runtime concern.

src/tokens/build.mjs
```diff
@@ -1,56 +1,48 @@
...
+          selector: themeName === 'reapit' ? ':root, :root[data-theme="reapit"]' : `:root[data-theme="${themeName}"]`,
+        },
+        transformGroup: 'web',
       },
     },
   }
 }
```
kurtdoherty, line 44: **note:** these changes are related to breaking changes introduced in `style-dictionary@4`

src/tokens/payprop/linaria-tokens.ts
kurtdoherty, line 1: **note:** The existing `payprop/*` files have been deleted. They are no different to the `legacy-reapit/*` files.

src/tokens/legacy-reapit/linaria-tokens.ts
kurtdoherty, line 1: **note:** I've move the legacy token output files to live in `legacy-reapit/*`. Once we've migrated all the v5 components to use the new CSS variables, we'll delete these artefacts.

kurtdoherty:
> **praise:** Thank you for getting this done, I was under the impression that you'll be just looking into the approach and I'll take care of the implementation. But thanks for getting this through.

No worries. Turned out to be one of those things where, to work out what we need to do, it was easiest to just do it 😄

PR #333, opened 2025-03-11
feat: #154 v5 new SideBar (Menu Group and Group Item)

nurm717123:
## Context
In order to create a new SideBar for the v5 release, this PR, as part of #154, initended to add the MenuGroup and MenuGroupItem component in SideBar.

image

## Pull request checklist
- [x] SideBarMenuGroup
- [x] SideBarMenuGroupItem
- [x] integration with SideBar's collapsed/expanded state

fixes: #

side-bar-menu-group

### DIFF
--- a/src/components/side-bar-menu-group/side-bar-menu-group.tsx
+++ b/src/components/side-bar-menu-group/side-bar-menu-group.tsx
@@ -1,0 +1,54 @@
+import type { AnchorHTMLAttributes, FC, HTMLAttributes, ReactNode } from 'react'
+import {
+  SideBarMenuGroupContainer,
+  SideBarMenuGroupItemTrigger,
+  SideBarMenuGroupList,
+} from './side-bar-menu-group.molecules'
+import { elSideBarMenuGroupItemAnchor, ElSideBarMenuGroupItem, ElSideBarMenuGroupListItem } from './styles'
+
+export interface SideBarMenuGroupItemProps extends Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'onClick'> {
+  children?: ReactNode
+  /**
+   * Whether the page represented by this link is currently active.
+   *
+   * @default false
+   */
+  isActive?: boolean
+}
+
+export const SideBarMenuGroupItem: FC<SideBarMenuGroupItemProps> = ({ children, isActive, ...props }) => {
+  return (
+    <ElSideBarMenuGroupListItem>
+      <ElSideBarMenuGroupItem
+        className={elSideBarMenuGroupItemAnchor}
+        aria-current={isActive ? 'page' : undefined}
+        {...props}
+      >
+        {children}
+      </ElSideBarMenuGroupItem>
+    </ElSideBarMenuGroupListItem>
+  )
+}
+
+export interface SideBarMenuGroupProps extends Omit<HTMLAttributes<HTMLButtonElement>, 'onClick'> {
+  children?: ReactNode
+  /**
+   * Whether the page represented by this group contain active link.
+   *
+   * @default false
+   */
+  isActive?: boolean
+  icon: ReactNode
+  label: string
+}
+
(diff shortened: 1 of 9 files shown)


src/components/side-bar-menu-group/side-bar-menu-group.tsx
```diff
@@ -0,0 +1,54 @@
...
+   */
+  isActive?: boolean
+  icon: ReactNode
+  label: string
+}
+
+export const SideBarMenuGroup: FC<SideBarMenuGroupProps> = ({ label, icon, isActive, children }) => {
+  return (
+    <SideBarMenuGroupContainer isActive={isActive}>
+      <SideBarMenuGroupItemTrigger isActive={isActive} icon={icon}>
```
nurm717123, line 48: can't use context's expanded state, because expanded menu-group doesn't mean it contain active link

nurm717123:
hi @kurtdoherty this pr is ready for review now, a few notes for some of my reasoning for some part is

- I'm not use snapshot test, to avoid larger file change, instead I use normal test just like `Chip` component
- I avoid combine the component fully with the SideBar so it became .eg: `<SideBar.Group.List/>` and/or `<SideBar.Group.Item/>` because I'm not sure if this is what we going to plan to have 2 nested sub component, I think it will be better to do in separate pr 🤔 
- the naming looks very long, I think it can be better if we use sub components for all of it 🤔 
- `.molecules` only used in main file `side-bar-menu-group`, that's why I didn't export it in `index.ts` 

let me know if there is any concern, thanks

src/components/side-bar-menu-group/styles.ts
```diff
@@ -0,0 +1,120 @@
...
+export const ElSideBarMenuGroupItemText = styled.span`
+  font-family: var(--font-family);
+  font-size: var(--font-size-sm);
+  font-style: normal;
+  font-weight: var(--font-weight-regular);
+  line-height: var(--line-height-sm);
+  letter-spacing: var(--letter-spacing-sm);
+`
+
+export const ElSideBarMenuGroupTriggerIcon = styled.span`
```
nurm717123, line 24: using span, to allow consumer to use custom icon

src/components/side-bar/side-bar.stories.tsx
```diff
@@ -60,6 +61,10 @@ export const Default: Story = {
             <SideBarMenuItem isActive icon={<Icon icon="property" />} href="#">
               SideBar Item (active)
             </SideBarMenuItem>
+            <SideBarMenuGroup isActive label="Menu item Expandable" icon={<Icon icon="property" />}>
+              <SideBarMenuGroupItem isActive>Sub Menu Item 1</SideBarMenuGroupItem>
+              <SideBarMenuGroupItem>Sub Menu Item 2</SideBarMenuGroupItem>
+            </SideBarMenuGroup>
```
nurm717123, line 67: this kind of pattern I still not sure,  should we provide this shorthand

or just let user use the molecules, e.g:

```javascript
<SideBarMenuGroupContainer isActive >
  <SideBarMenuGroupItemTrigger isActive icon={</Icon  icon="property">}>Menu item Expandable</SideBarMenuGroupItemTrigger>
  <SideBarMenuGroupList>
    <SideBarMenuGroupItem isActive>Sub Menu Item 2</SideBarMenuGroupItem>
    <SideBarMenuGroupItem>Sub Menu Item 3</SideBarMenuGroupItem>
  </SideBarMenuGroupList>
</SideBarMenuGroup>
```

kurtdoherty: I think this is fine for now. Not hard for us to move to the use of the molecules later if we discover that's better in practice.

kurtdoherty (approved):

src/components/side-bar/side-bar.stories.tsx
nurm717123: yeah👍

Issue #245, opened 2024-12-09
v5 component refactor: Pagination


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing
- Is the component being re-named? No
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=1663-23201&m=dev)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/pagination--docs)
- [Accessibility requirements](https://design-system.w3.org/components/pagination.html)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

ss-dimasm:
Proposed Solution Design - Pagination

## Component Structure

Pagination is better suited for use with the Provider pattern. This pattern requires cloning the state to synchronize details across components, and to ensure the composed component should be used within a correct parent. Here is [a simple POC](https://codesandbox.io/p/sandbox/elements-pagination-solution-design-3g58fy) for Pagination using the Provider pattern

**1. Pagination**
Outlines the core concept of a Pagination component, typically implemented using a `nav` element in the DOM

* **Purpose:** 
    * Setup and manages the internal state of pagination that will be shared across the composed component
    * Serves as the entry point for other, more specific composed components (e.g., "Backward," "Forward"  buttons)

* **Key Properties:**

| Prop          | Description                                                                     | 
|---------------|------------------------------------------------------------------------------|
| `initialPage` | The initial page to display when the pagination component first mount       | 
[...]

PR #335, opened 2025-03-17
chore: #245 improve documentation for pagination

ss-dimasm:
# Pull request checklist

Part of #245, Resolve the misunderstanding on how the Pagination should behave in Storybook based on the QA feedback

here is the preview addition documentation

image


kurtdoherty:
**question:** @ss-dimasm what was the misunderstanding?

ss-dimasm:
> what was the misunderstanding?

https://github.com/user-attachments/assets/29ef2c68-6637-4973-9f82-48dc41afa857

re @kurtdoherty , feedback highlighted that the Control jsx in Storybook doesn't represent 100% on how the Pagination component works, it leads to confusion 😕 for `currentPage` prop

The Storybook example used an external state to make it look functional. Unfortunately, the control and external props weren't linked after it mounted

QA and @ksolanki7 found this issue a week ago. We decided to add usage snippets to make it clearer how the Pagination component should be used

kurtdoherty:
Okay, no worries 👍  This is fine for now, but I don't think we should make a habit of taking this approach.

The main drawback with manually crafted code snippets like this is the snippets can easily diverge from the real component's API. That is, nothing is enforce the snippet code to actually be correct; as soon as we introduce a prop interface change, or a DOM structure change, these snippets will be out-of-date and we'll be relying on the dev making those changes to notice these snippets need updating.

Again though, what you've done is fine for now.

kurtdoherty (approved):

Issue #281, opened 2025-01-16
v5 component refactor: Icon


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing 
- Is the component being re-named? No
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=4252-8141&m=dev)
- [Link to Elements](https://elements.reapit.cloud/?path=/docs/icon--docs)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1052-22272&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

kurtdoherty:
**suggestion:** I think this work should include the following requirements:

- Each icon should be independently importable. Consumers that only need a specific set of icons should not be forced to bundle all of them
- All icons should, by default, inherit their colour from their ancestor.

lilykirkwood:
@HollyJoyPhillips Hi Holly, I'm looking to confirm the v5 icon set. The Elements docs outline an icon set, and there is also an icon set in [Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=10-2608&p=f&t=c5Za4io4UoGIYxvf-0). However, the two sets do not match. Could you please help with the following, or tag the relevant person?
1. There are 43 icons in Figma that are not present in Storybook – just checking, are these new icons that need to be added?
[...]

PR #338, opened 2025-03-18
feat: 281 v5 icon component

lilykirkwood:
# Pull request checklist

* New v5 icon set: [Figma production file](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=6696-323&t=jfzqk06TvFLwFdMm-0)
* Icons now inherit colour by default
* Icons are individually importable to avoid large bundle size

**Explanation of the per-icon component solution:**
* `scripts/generate-icons.ts` runs over all SVGs inside `assets/icons` and creates per-icon React component files.
* Each icon component focuses solely on rendering its SVG, while shared logic is abstracted away – they all then use `createIconComponent` in order to centralise common aspects (same props and styling as the existing Icon component).
* For now the components output in `src/iconsNew` ahead of removing the old Icon component.
* We should consider whether we want these generated files committed at all?
* Icons are imported like so `import { WalkingIcon } from '../../iconsNew/walking'` or via the barrel file.
* Consider whether we want a barrel file, or if it defeats the point? I don't know a lot about tree shaking at the moment so would want to understand the impact of importing like this instead: `import { WalkingIcon } from '../../iconsNew'`.

**WIP:**
* Tests!
* Implement new icon components across repo
* Change intents to variants?
* Check that icons adhere to new [sizing guide](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=945-21415&m=dev)
* Update Storybook with icon component changes

fixes: https://github.com/reapit/elements/issues/281


kurtdoherty:
@lilykirkwood how are you going with this? Do you need any assistance getting this over the line or would you like to discuss the feedback above any further?

kurtdoherty:
@lilykirkwood thanks for all the initial work here; was a great reference point for the work I did to finish the implementation of new icons 🙏  I'm closing this PR out now though, as it's no longer needed.

Issue #340, opened 2025-03-19
v5 Beta Change log update and release


ksolanki7:
**Background Context:**
The last beta release was made in October 2024. Since then, several v5 elements have been developed, but no new release has been published.

**Specification / Tasks:**
Prepare and publish a new beta release incorporating all work completed to date.

PR #343, opened 2025-03-20
chore: #340 Updated Prerelease test script to vitest

ksolanki7:
**Background:**

We recently replaced Jest with Vitest and missed updating the prerelease GitHub workflow to use Vitest.

fixes: #340 `5.0.0-beta.18` release - Replaced jest test script with vitest

### DIFF
--- a/.github/workflows/release-prerelease.yml
+++ b/.github/workflows/release-prerelease.yml
@@ -61,7 +61,7 @@ jobs:
 
       - name: Test
         run: |
-          yarn test --watch=false --ci --forceExit --detectOpenHandles --runInBand --passWithNoTests
+          yarn vitest run --watch=false --passWithNoTests
 
       - name: Package Version Bump
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release-prerelease.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
       - name: Test
         run: |
-          yarn test --watch=false --ci --forceExit --detectOpenHandles --runInBand --passWithNoTests
+          yarn vitest run --watch=false --passWithNoTests
```
kurtdoherty, line 64: **note:** `yarn test` should still work. It'll run Vitest for us. The `run` option implies `--watch=false`.

kurtdoherty (approved):

.github/workflows/release-prerelease.yml
ksolanki7: I had the same thought, then I just followed the previous implementation. Anyways will update and push it.

PR #344, opened 2025-03-20
feat: #154 SideBar Accessibility (a11y)

nurm717123:
## Context
In order to create a new SideBar for the v5 release, this PR, as part of #154, initended to add the accessibility support and keyboard interaction in SideBar.

for this work I referring to https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation following the pattern in MobileNav ticket previously.

## Pull request checklist
- [x] support `arrowUp` and `arrowDown` keys to navigate between items
- [x] support `Enter` or `Space` key to click item
- [x] support `Escape` key to unexpand the group and move focus to the trigger
- [x] ensure the correct style applied

fixes: #

sidebar-a11y

### DIFF
--- a/src/components/side-bar-menu-group/styles.ts
+++ b/src/components/side-bar-menu-group/styles.ts
@@ -45,7 +45,6 @@ export const ElSideBarMenuGroupItemTrigger = styled.button`
 
   display: flex;
   width: 100%;
-  justify-content: flex-start;
   height: 40px;
   padding: var(--spacing-none) var(--spacing-2);
   align-items: center;
@@ -54,7 +53,7 @@ export const ElSideBarMenuGroupItemTrigger = styled.button`
   border-top-left-radius: inherit;
   border-top-right-radius: inherit;
 
-  &[data-expanded='false'] {
+  &[aria-expanded='false'] {
     border-radius: inherit;
   }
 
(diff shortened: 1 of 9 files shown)


src/components/side-bar-menu-group/styles.ts
```diff
@@ -18,6 +18,8 @@ export const ElSideBarMenuGroupItemText = styled.span`
   font-weight: var(--font-weight-regular);
   line-height: var(--line-height-sm);
   letter-spacing: var(--letter-spacing-sm);
+  flex: 1;
+  text-align: left;
 `
```
nurm717123, line 23: after I update the story's trigger label to be shorter, I found that the UI is not right, this is to fix that case

src/components/side-bar-menu-group/styles.ts
```diff
@@ -42,6 +44,7 @@ export const ElSideBarMenuGroupItemTrigger = styled.button`
   border: none;
 
   display: flex;
+  width: 100%;
```
nurm717123, line 47: following this fix

> after I update the story's trigger label to be shorter, I found that the UI is not right, this is to fix that case

kurtdoherty (approved):

Issue #314, opened 2025-02-12
v5 component refactor: Add `maxWidth` to new `Chip` component


kurtdoherty:
> @andrei-dragut @fazil-99 Chip and ChipGroup are available for review:
> - [Chip](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-chip--docs)
> - [ChipGroup](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-chipgroup--docs)
> 
> **note:** This work was merged prior to the agreement in Figma to add a maxWidth prop to the Chip component. I'll create a separate ticket for that to be added once we've rolled out the new variables. 

 _Originally posted by @kurtdoherty in [#258](https://github.com/reapit/elements/issues/258#issuecomment-2652373816)_

PR #345, opened 2025-03-21
feat: #314 add `maxWidth` prop to `Chip` and use new design tokens

kurtdoherty:
fixes: #314

- Adds `maxWidth` prop to `Chip`
- Updates `Chip` to use the new design tokens

The image below shows chips with a max width applied. Originally, this was achieved by rendering them inside a parent element that had its own max width. That parent is no longer present and the Chip's are handling the max width themselves.

Screenshot 2025-03-21 at 10 06 36 AM

Here, we can see the Chip's max width handling at work within a Chip group.

Screenshot 2025-03-21 at 10 08 46 AM

### DIFF
--- a/src/components/chip/chip.stories.tsx
+++ b/src/components/chip/chip.stories.tsx
@@ -2,7 +2,7 @@ import { Chip } from './chip'
 import { Tooltip } from '../tooltip'
 import { useTooltip } from '../tooltip/use-tooltip'
 
-import type { Decorator, Meta, StoryObj } from '@storybook/react'
+import type { Meta, StoryObj } from '@storybook/react'
 
 const meta = {
   title: 'Components/Chip',
@@ -70,21 +70,13 @@ export const Disabled: Story = {
   },
 }
 
-const useNarrowParentDecorator: Decorator = (Story) => {
-  return (
-    <div style={{ width: '300px' }}>
-      <Story />
-    </div>
-  )
-}
-
 /** By default, long labels will wrap if there is not enough space is available. */
 export const Wrapping: Story = {
   args: {
     ...FilterChip.args,
     children: "This very long label will wrap because it's parent is not wide enough",
+    maxWidth: '--size-80',
   },
-  decorators: [useNarrowParentDecorator],
 }
 
 /**
@@ -95,7 +87,7 @@ export const Truncation: Story = {
   args: {
     ...FilterChip.args,
     children: 'Truncation can be applied when necessary',
+    maxWidth: '--size-80',
     willTruncateLabel: true,
   },
-  decorators: [useNarrowParentDecorator],
 }
--- a/src/components/chip/chip.tsx
+++ b/src/components/chip/chip.tsx
@@ -13,14 +13,15 @@ type ElementAttributesToOmit = Extract<keyof ElementAttributes, 'disabled' | 'ty
 interface ChipProps extends Omit<ElementAttributes, ElementAttributesToOmit> {
   children: ReactNode
   isDisabled?: boolean
-  willTruncateLabel?: boolean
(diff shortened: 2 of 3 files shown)


src/components/chip/chip.stories.tsx
```diff
@@ -70,21 +70,13 @@ export const Disabled: Story = {
   },
 }
```
kurtdoherty, line 72: **note:** this decorator was rendering the story inside a parent with a fixed width. This was how we were originally demonstrating the Chip label's wrapping vs truncation behaviour. Now that Chip's accept a max width themselves, I've removed this decorator and simply provide a max width via the Story's args.

src/components/chip/chip.tsx
```diff
@@ -13,14 +13,15 @@ type ElementAttributesToOmit = Extract<keyof ElementAttributes, 'disabled' | 'ty
 interface ChipProps extends Omit<ElementAttributes, ElementAttributesToOmit> {
   children: ReactNode
   isDisabled?: boolean
```
kurtdoherty, line 15: **note:** I've typed this new `maxWidth` prop as tightly as we can right now. In future, I hope to have proper TS types based on the actual design tokens.

src/components/chip/chip.tsx
```diff
@@ -39,7 +40,16 @@ export function Chip({ children, isDisabled, onClick, variant, willTruncateLabel
   )
 
   return (
-    <ElChip {...rest} type="button" aria-disabled={isDisabled} data-variant={variant} onClick={handleClick}>
+    <ElChip
+      {...rest}
+      type="button"
+      aria-disabled={isDisabled}
+      data-variant={variant}
+      onClick={handleClick}
```
kurtdoherty, line 48: **note:** This CSS variable approach is the same pattern used for TextArea's resizing behaviour. As the code comment points out, when browser support the modern [attr()](https://developer.mozilla.org/en-US/docs/Web/CSS/attr) syntax, we'll be able to replace this pattern with a `data-max-width` attribute.

src/components/chip/styles.ts
```diff
@@ -17,29 +26,36 @@ export const ElChip = styled.button<ElChipProps>`
   padding-block: var(--spacing-1);
   padding-inline: var(--spacing-4) var(--spacing-2);
   width: fit-content;
```
kurtdoherty, line 28: **note:** If no `maxWidth` prop is provided, there'll be no `--chip-max-width` CSS variable in scope, and thus, we'll fallback to `auto` for the max width.

src/components/chip/styles.ts
```diff
@@ -17,29 +26,36 @@ export const ElChip = styled.button<ElChipProps>`
...
+  max-width: var(--chip-max-width, auto);
+
```
kurtdoherty, line 30: **note:** the rest of the changes here are related to replacing the "old" tokens with the new component-specific tokens.

src/components/chip/styles.ts
```diff
@@ -1,14 +1,23 @@
 import { styled } from '@linaria/react'
 import { Icon } from '../icon'
+import type { CSSProperties } from 'react'
+
+interface ElChipCSSProperties extends CSSProperties {
+  /** Used to determine the maximum width of the chip because the browser does not support
+   * [CSS' advanced attr() syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/attr) */
+  '--chip-max-width'?: `var(--size-${string})`
```
ksolanki7, line 8: **praise:** I like that you use a token variable here instead of some hard-coded values for the prop. This is something we should audit in the future and update on other elements too.

ksolanki7 (approved):
Looks good.

Issue #348, opened 2025-03-24
v5 - Beta release 19 issue with importing elements type to consumer's project


ksolanki7:
**Describe the bug**
We recently released 5.0.0-beta.18 and 5.0.0-beta.19 and found that consumers are not able to import elements types.

**To Reproduce**
Steps to reproduce the behavior eg:

- Update reapit elements version to 5.0.0-beta.19
- Import elements type to extend for your project
- Error will appear for the import.

After updating reapit elements beta version from 18->19, reapit elements types import should work as is.

PR #349, opened 2025-03-24
fix: #348 Updated vite config to add the export types to production build

ksolanki7:
# Pull request checklist

fixes: #348 Updated the vite config to export types so that current consumers who import types avoid any import errors.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -3,6 +3,7 @@ import { defineConfig } from 'vite'
 import { fileURLToPath, URL } from 'node:url'
 import react from '@vitejs/plugin-react'
 import svgr from 'vite-plugin-svgr'
+import dts from 'vite-plugin-dts'
 import wyw from '@wyw-in-js/vite'
 import packageManifest from './package.json'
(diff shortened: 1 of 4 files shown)


vite.config.ts
```diff
@@ -60,6 +61,10 @@ export default defineConfig({
         presets: ['@babel/preset-typescript', '@babel/preset-react'],
       },
     }),
+    dts({
+      outDir: 'dist/js/src',
+      copyDtsFiles: true,
```
ksolanki7, line 66: - I have explicitly provided `outDir` to place the src files in correct folder
- copyDtsFiles set to true - this will generate the `.d.ts` file

ss-dimasm: I think to make it aligned with the exposed entry point in [package.json#28](https://github.com/reapit/elements/blob/main/package.json#L28), we need to change the `outDir` path to be `dist/types` , so it will expose the types with the correct reference

ss-dimasm (approved):

PR #353, opened 2025-03-24
chore: #154 v5 sidebar update css var

nurm717123:
## Context

_As part of the v5 Elements release, there is a design token (CSS variable) update. Each existing component that still uses old tokens needs to be updated. This ticket specifically only for Sidebar's work._

also contain little style fix (I'll provide a comment for detailed description)

# Pull request checklist
- [x] SideBarMenuItem
- [x] SideBarMenuGroup
- [x] SideBarCollapseButton

### DIFF
--- a/src/components/side-bar-collapse-button/styles.ts
+++ b/src/components/side-bar-collapse-button/styles.ts
@@ -30,15 +30,14 @@ export const ElSideBarCollapseButton = styled.button`
   color: var(--neutral-600);
   border-radius: var(--corner-lg);
 
-  &:hover {
-    background: var(--fill-default-lightest);
-    font-weight: 500;
+  &:hover,
+  &:focus-visible {
+    background: var(--colour-fill-neutral-light);
   }
 
   &:focus-visible {
     outline: none;
     border: var(--border-width-double) solid var(--colour-border-focus);
-    background: var(--color-fill-neutral-light);
   }
 `
 
--- a/src/components/side-bar-menu-group/styles.ts
+++ b/src/components/side-bar-menu-group/styles.ts
@@ -57,11 +57,11 @@ export const ElSideBarMenuGroupItemTrigger = styled.button`
     border-radius: inherit;
   }
 
-  color: var(--component-text-colour-text-side-bar-menu-item-default, #506478);
+  color: var(--neutral-600);
 
   &[aria-current='page'] {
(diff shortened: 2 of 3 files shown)


src/components/side-bar-collapse-button/styles.ts
```diff
@@ -30,15 +30,14 @@ export const ElSideBarCollapseButton = styled.button`
   color: var(--neutral-600);
   border-radius: var(--corner-lg);
 
-  &:hover {
-    background: var(--fill-default-lightest);
-    font-weight: 500;
```
nurm717123, line 35: font-weight only for selected style, while collapse button doesn't have

src/components/side-bar-menu-item/styles.ts
```diff
@@ -35,22 +35,19 @@ export const elSideBarMenuItemAnchor = css`
...
 
-  &:hover,
-  &:focus-visible {
-    background: var(--colour-fill-neutral-light);
-  }
-
-  &[aria-current='page'] {
-    background: var(--fill-default-lightest);
-  }
-
```
nurm717123, line 46: move the hover style after `:selected` style, so the hover style will still applied i,e not overriden

src/components/side-bar-menu-group/styles.ts
```diff
@@ -101,7 +101,7 @@ export const elSideBarMenuGroupItemAnchor = css`
   display: flex;
   align-items: center;
   text-align: start;
-  color: var(--component-text-colour-text-side-bar-menu-item-default, #506478);
```
nurm717123, line 104: I didn't see this variable in the latest token update, so I update it using this

based on:
image

ksolanki7: @nurm717123 Please check with the design team, as we shouldn't be assuming and replacing variables as such with fallbacks. There are some bugs in Figma which makes things harder for both Devs and Designer to move forward if one is not aware of such issues resulting in bugs.

If we make the design team aware of such issues/bugs in Figma, they can go back and fix the issue in Figma manually and then they can provide us with the correct variables.

nurm717123: yeah, I wondering why it isn't generated, I've raised the question [here](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=16-5010&m=dev#1185396885)

Issue #371, opened 2025-03-27
Breadcrumbs Element: Unintended black border on focus state & incorrect svg size on chevronRight icon


schristiani-rpt:
**Describe the bug**
Issues on breadcrumbs element:
- Unintended black border when on focus state (see screenshot below)
- Incorrect size on chevronRight icon svg, both height & width should be in size 12 px but found 1em, and the viewBox is "0 0 24 24" instead of "0 0 12 12"

**To Reproduce**
***Unintended black border on Focus State (Default):***
1. Go to breadcrumbs element on storybook (https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-breadcrumbs--default)
2. Press "Tab" button on the keyboard
3. Notice that there's black border inside the element

***Unintended black border on Focus State (Mobile viewport):***
1. Go to breadcrumbs element on storybook (https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-breadcrumbs--default)
2. Change the viewport to "Small Mobile"
3. Press "Tab" button on the keyboard
4. Notice that there's black border inside the element

***Incorrect svg's width & height on chevronRight icon:***
1. Go to breadcrumbs element on storybook (https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-breadcrumbs--default)
2. Inspect the chevronRight icon
3. On devTool, go to svg part under the span of `el-breadcrumbs-item-chevron el-icon`
4. Notice that the width, height, and viewBox is different than the figma

**Expected behaviour**
- On focus state, both of the default & mobile viewport view, should not have another black border inside the element
- On svg part, it should be as per figma, both width & height value should be 12 and the viewBox is "0 0 12 12"

**Screenshots**
***Focus state-Default***
![Image](https://github.com/user-attachments/assets/6e25df56-9742-4c71-acd9-11095bed5729)

**_Focus state - Mobile Viewport_**
![Image](https://github.com/user-attachments/assets/63ef123a-8699-4cf4-96d0-0c9a050bc6d8)

***Incorrect svg size***
Actual:
![Image](https://github.com/user-attachments/assets/12f1b79d-746d-4771-8df5-2d485760bb38)

Expected:
![Image](https://github.com/user-attachments/assets/42cb9fd2-050c-425d-ab52-049174f2d2c4)

**Device (please complete the following information for web issues):**
Tested on:
 - Device: Laptop
 - OS: Windows 11
 - Browser: Chrome

**Additional context**
[...]

PR #372, opened 2025-03-27
fix: #371 v5 breadcrumbs focus style on mobile screen

nurm717123:
## Context
_QA reported several issues in #371. This PR fixes all the mentioned issues to ensure expected functionality._

the summary of the issue is

1. mobile screen focus style broken
   The focus box shadow was not visible due to overflow=auto on mobile screens.
   Reference: [CodeSandbox Example](https://codesandbox.io/p/sandbox/8f35r7).
2. Default HTML outline visible on focus
3. Icon size mismatch with parent span
   While this didn’t break the UI (since the wrapper size is defined), updating it ensures future consistency in case the Icon component changes.

# Pull request checklist
- [x] fix mobile screen focus style
- [x] remove focus default html outline
- [x] force icon svg to have same size

fixes: #

breadcrumb-mobile-focus


src/components/breadcrumbs/styles.ts
```diff
@@ -21,8 +21,11 @@ export const ElBreadcrumbsItemChevron = styled(Icon)`
   padding: 0px var(--spacing-1);
 
   box-sizing: content-box;
-  width: var(--icon-xs);
-  height: var(--icon-xs);
+  &,
+  & svg {
+    width: var(--icon-size-xs);
```
ksolanki7, line 26: @nurm717123 
**issue(blocking):** The token variables are updated now. Can you please pull `main` to your branch and update the CSS variable with the updated once.

ksolanki7 (changes requested):

adrian-reapit (approved):

src/components/breadcrumbs/styles.ts
```diff
@@ -21,30 +21,37 @@ export const ElBreadcrumbsItemChevron = styled(Icon)`
   padding: 0px var(--spacing-1);
 
   box-sizing: content-box;
-  width: var(--icon-xs);
-  height: var(--icon-xs);
+  &,
+  & svg {
+    width: var(--icon_size-xs);
+    height: var(--icon_size-xs);
+  }
```
ksolanki7, line 28: @nurm717123 **suggestion:**  I believe Icon has a prop for font-size, that should be used in the story and usage rather than hardcoding it via CSS.

src/components/breadcrumbs/styles.ts
```diff
@@ -21,30 +21,37 @@ export const ElBreadcrumbsItemChevron = styled(Icon)`
...
+  font-weight: var(--font-sm-regular-weight);
+  line-height: var(--font-sm-regular-line_height);
+  letter-spacing: var(--font-sm-regular-letter_spacing);
 
   cursor: pointer;
   white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;
 
+  padding: 1px; // to imitate the link focus size
```
ksolanki7, line 47: @nurm717123 **question:** Is there a specific reason not to use outline and represent the outline using border and padding?

src/components/breadcrumbs/styles.ts
nurm717123: yes, but the Figma use width and height for it, idk what the reason, probably to prevent the case when consumer use a custom Icon which has width and height defined, in that case fontSize are ignored

image

src/components/breadcrumbs/styles.ts
nurm717123: I've tried it, the main reason is because the figma uses it and even I use outline we have to add additional padding there (the padding is to match the figma style where they use additional absulte position container to style the border)

nurm717123: i.e we can update it using outline here, but the padding is still needed, what do you think?

src/components/breadcrumbs/styles.ts
ksolanki7: @nurm717123  I believe for custom icon, we need to leave it up to the consumer, we shouldn't anyhow force consumer's UI if its not elements icon. The font-size in return will manage the height and width of the icon. The new icon element will have better proposal of defining font-size rather than hardcoding the values in. Can discuss this in detail if require.

src/components/breadcrumbs/styles.ts
ksolanki7: @nurm717123 **suggestion:** You can use outline-offset - https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset to manage it.

src/components/breadcrumbs/styles.ts
nurm717123: I got it 👍 

aside: would be great if figma also use font-size to avoid confusion in the future

src/components/breadcrumbs/styles.ts
nurm717123: I just knew this property 😆, then I think we can use it

nurm717123: @ksolanki7 I've tried using outline, but I just remember it has an issue with parent's overflow behaviour, I've make a sandbox for it https://codesandbox.io/p/sandbox/n9wxd9, so I revert it back to use border

ksolanki7: @nurm717123 That's all good, however I'll try to find some solution to work out with the outline.

ksolanki7 (approved):

Issue #286, opened 2025-01-23
v5 component refactor: Checkbox


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? Existing 
- Is the component being re-named? Yes - currently 'Input checkbox - rename to 'Checkbox' 
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://elements.reapit.cloud/?path=/story/input--input-checkbox)
- [Link to Elements](https://elements.reapit.cloud/?path=/story/input--input-checkbox)
- [Accessibility requirements related docs](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/)
- [Link to Design Guidelines](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1052-22272&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

DanishAli-Reapit:
Hi Team,

After studying the Figma design below is my proposed design solution:

---

## Checkbox Design Solution

The Checkbox will have the following main prop interface

| Prop | Value | Default | Usage |
| --- | --- | --- | --- |
| children?: ReactNode | | | For nesting checkbox component |
| label?: string | | | checkbox label |
| supplementaryInfo?: string | | | checkbox supplementary Info label |
| checked?: boolean | 'true', 'false' | false | to handle checkbox check icon |
| indeterminate?: boolean | 'true', 'false' | false | to handle checkbox indeterminate icon |
| onChange?: void | | | onChange event to render checkbox |
| value: string | | | checkbox value |
[...]

PR #373, opened 2025-03-28
feat: #286 created the v5 checkbox component

DanishAli-Reapit:
# Description

- Created a new v5 Checkbox component to reflect DS Figma
- Added accessibility support to the checkbox


src/components/checkbox/styles.css
```diff
@@ -0,0 +1,98 @@
+.el-checkbox-label {
+  display: grid;
+  gap: 2px 8px;
```
ksolanki7, line 3: **suggestion:** Use token variables instead of hardcoding.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Interface for the Checkbox component props.
+ *
+ * Optional label text to display next to the checkbox.
+ * Provides a user-friendly description of the checkbox's purpose.
+ *
+ * Optional supplementary information to display below the label.
+ * Offers additional context or details related to the checkbox.
+ *
+ * Optional required to make input element required.
```
ksolanki7, line 22: **issue (non-blocking):** Ensure when trying JSDocs, you include all related information. I see there is no mention of `isIndeterminate` type. This may misguide the DX.

src/components/checkbox/styles.css
```diff
@@ -0,0 +1,98 @@
+.el-checkbox-label {
```
ksolanki7, line 1: **issue (blocking):** Ensure we follow the same pattern for styles, i.e. `styles.ts` and use Linaria to define the HTML elements.

ksolanki7 (changes requested):
I have left some comments to resolve, so stopping any further review until its' fixed.

src/components/checkbox/checkbox.tsx
DanishAli-Reapit: Updated

ksolanki7 (approved):
looks good.

src/components/checkbox/styles.css
DanishAli-Reapit: Updated

Issue #374, opened 2025-03-28
v5 DS Token Variable - Issue fix with variable builder replacing '_' with '-' and update tokens variable with latest


ksolanki7:
**Describe the bug**
Currently, the DS token variable builder replaces the '_' with '-' for all the token variable respective of the naming convention coming from json file.

**To Reproduce**
Steps to reproduce the behaviour eg:
1. Check Sementic json file
2. Check the generated reapit.css and compare it with Figma naming. (Explicitly check the folder/variable name with '_')

**Expected behaviour**
If json is
```
{
  "icon_size": {
      "xs": {
        "$type": "dimension",
        "$value": "16px",
        "$description": "Used only for icon sizing"
      }
  },
  "text": {
    "primary": {
      "$type": "color",
      "$value": "#fff"
    },
    "accent_1": {
      "$type": "color",
      "$value": "#000"
    }
  } 
}
```

Then the output should be 
```
--icon_size-xs: 16px
--text-primary-default: #fff
--text-accent_1: #350 
```
Current generated CSS variable which is wrong as below:
```
--icon-size-xs: 16px
--text-primary-default: #fff
--text-accent-1: #350 
``` 

**Tasks:**

- Fix the token variable builder
- Update Sementic and Primitive(if any) new updates are available from DS Figma

PR #375, opened 2025-03-31
fix: #374 Fix the style dictionary builder to support the correct naming convention coming from Figma

ksolanki7:
# Pull request checklist

fixes: #374 

- Fix the builder to generate the CSS variable with the correct naming convention.
- Updated the Semantic variables with latest Figma


src/tokens/build.mjs
```diff
@@ -10,6 +10,26 @@ import StyleDictionary from 'style-dictionary'
 /** @type {Theme[]} */
 const themes = ['payprop', 'reapit']
 
+StyleDictionary.registerTransform({
```
ksolanki7, line 13: Refer: https://styledictionary.com/reference/api/#registertransform:~:text=%7D)%3B-,registerTransform,-StyleDictionary.registerTransform(transform for custom variable naming convention.

adrian-reapit (approved):

Issue #361, opened 2025-03-25
Incorrect pagination button size


schristiani-rpt:
**Describe the bug**
Pagination button size is incorrect, it is expected to have size-8 (32px) but on actual it use size-9 (36px)

**To Reproduce**
1. Go to Pagination Element on Storybook (https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-pagination--docs)
2. Inspect on the pagination button
3. Find button element with class ".el-button.el-button-icon-only"
4. Check on the css styles, on width property, notice that it use size-9 instead of size-8

**Expected behaviour**
Button size should be using size-8 for both of width and height

**Screenshots**
Actual:
![Image](https://github.com/user-attachments/assets/c967a359-b032-45b1-8df8-c747b9686c88)
![Image](https://github.com/user-attachments/assets/89e638b5-3c0f-4333-a9fb-407f039e1e6e)

Expected on [Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=4505-3411&m=dev):
![Image](https://github.com/user-attachments/assets/a49b7d46-10f9-40f8-86ad-42b5489e6cfa)

**Device (please complete the following information for web issues):**
Tested on:
 - Device: Laptop
 - OS: Windows 11
 - Browser: Chrome

ss-dimasm:
hi @schristiani-rpt , would you mind to revalidate the issue in the latest url?

schristiani-rpt:
Hi @ss-dimasm 
I've tested in the latest url, the button size is now correct as per design 👌🏻

**Status PASS**
Width & Height already correct as per design, (--size-8, 32 px)

![Image](https://github.com/user-attachments/assets/3b548222-333e-49d0-85e5-70753871a0ae)
![Image](https://github.com/user-attachments/assets/d7df2a8c-3a41-46bd-b8c3-33d831832e07)

Tested on:
- Chrome browser
- [Storybook Link](https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-pagination--default)

PR #381, opened 2025-04-02
fix: #361 (pagination) change button size and use semantic CSS variable

ss-dimasm:
# Pull request checklist

fixes: #361 

- Change button size in pagination component
- use latest semantic CSS variable token for pagination


src/components/pagination/__tests__/pagination.test.tsx
```diff
@@ -7,10 +7,6 @@ describe('Pagination', () => {
     vi.clearAllMocks()
   })
 
-  it('should match a snapshot', () => {
-    expect(renderComponent().asFragment()).toMatchSnapshot()
-  })
-
```
ss-dimasm, line 13: note: following the Chip component unit test structure it doesn't have the snapshot test; and the pagination already have functionality test case

PR #384, opened 2025-04-03
feat: #236 (mobile-nav-menu) initialize mobile-nav-menu component

ss-dimasm:
# Pull request checklist

Part of #236 
To continue the development for MobileNavMenu after we've settle an approach in SideBar component

This PR is deliberately to ship the MobileNavMenu based on [last agreement](https://github.com/reapit/elements/issues/236#issuecomment-2550558575) with @kurtdoherty  and also already use the latest token from the Design

There would be another PR that focusing on a11y and keyboard interaction based on @ksolanki7  [suggestion](https://github.com/reapit/elements/issues/236#issuecomment-2635465523)

Please to note, this PR rely on #264 , would be better if we also look at that first

## Attachment

https://github.com/user-attachments/assets/9e3fab54-4e9b-48d8-a29a-9ff3925f2cd0


ss-dimasm:
need to sunset any PR that related with MobileNavMenu, already raise the new version called TopBarMenu

Issue #383, opened 2025-04-02
Sidebar: Incorrect default state on some breakpoints


schristiani-rpt:
**Describe the bug**
On [Design Guideline](https://www.figma.com/design/pJ96bxoEFStjjIbAPy0Jaj/Design-Guidelines?node-id=1661-6502&m=dev), there'requirement for sidebar default state:
- On breakpoints <1440 px -> default state is collapsed
- On breakpoints >=1440 px -> default state is expanded

On actual, there's no different between any breakpoints, the default state is always expanded

**To Reproduce**
1. Go to [Sidebar element](https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-side-bar--default)
2. Inspect and turn on toggle device toolbar
3. Change the breakpoints and notice that the default state is always expanded

**Expected behaviour**
- On breakpoints <1440 px -> default state should be collapsed
- On breakpoints >=1440 px -> default state should be expanded

**Screenshots**
1024 px still expanded
![Image](https://github.com/user-attachments/assets/ba547147-675e-4f8f-9655-61ad85e1d542)

Screen record:
https://github.com/user-attachments/assets/c2cf948b-189c-4fbe-a83d-ccbf1d6d4252

**Device (please complete the following information for web issues):**
Tested on:
 - Device: Laptop
 - OS: Windows 11
 - Browser: Chrome

nurm717123:
fixed in https://github.com/reapit/elements/issues/443

PR #388, opened 2025-04-10
fix: #383 v5 update SideBar default expanded state

nurm717123:
## Context
_QA reported several issues in #371. This PR fixes all the mentioned issues to ensure expected functionality._

to fix this issue, I utilize existing `useMediaQuery` hook, so the user have to wrap the `SideBar` with the `MediaStateProvider` to make it work (I also have add a note in the documentation for it).

# Pull request checklist
- [x] On breakpoints <1440 px -> default state is collapsed
- [x] On breakpoints >=1440 px -> default state is expanded

fixes: #

sidebar default expanded state

### DIFF
--- a/src/components/side-bar/side-bar.tsx
+++ b/src/components/side-bar/side-bar.tsx
@@ -2,6 +2,7 @@ import { useState, type FC, type HTMLAttributes, type KeyboardEventHandler } fro
 import { SideBarCollapseButton } from '../side-bar-collapse-button'
 import { IsSideBarExpandedContext, useIsSideBarExpandedContext } from './is-side-bar-expanded-context'
 import { ElSideBar, ELSideBarMenuList } from './styles'
+import { useMediaQuery } from '#src/hooks/use-media-query/index'
 
 type SideBarFC = FC<HTMLAttributes<HTMLElement>> & {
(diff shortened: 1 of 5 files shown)


src/components/side-bar/side-bar.tsx
```diff
@@ -2,14 +2,16 @@ import { useState, type FC, type HTMLAttributes, type KeyboardEventHandler } fro
...
 
 type SideBarFC = FC<HTMLAttributes<HTMLElement>> & {
   CollapseButon: typeof SideBarCollapseButton
   MenuList: FC<HTMLAttributes<HTMLUListElement>>
 }
 
 const SideBar: SideBarFC = ({ children, ...props }) => {
-  const [isExpanded, setIsExpanded] = useState(true)
+  const { isWideScreen, isSuperWideScreen, is4KScreen } = useMediaQuery()
+  const [isExpanded, setIsExpanded] = useState(!!(isWideScreen || isSuperWideScreen || is4KScreen))
```
adrian-reapit, line 14: it would be good to allow for an override from a user preference. If the user opens the sidebar at 1300px wide it should be able to stay open if they refresh the page

nurm717123:
fixed in #443

Issue #394, opened 2025-04-14
v5 DS Token Variable: Issue with font weight value


ksolanki7:
**Describe the bug**
The CSS builder from DS token json adding `px` to font weight values.

**To Reproduce**
Steps to reproduce the behaviour eg:
1. Check reapit.css file

**Expected behaviour**
The transformation for json variable values should be exact and should be transform it to any condition.

PR #395, opened 2025-04-14
fix: #394 Fix the build script for token variables to opt out css tra…

ksolanki7:
…nsform for fontweight

# Pull request checklist

fixes: #394


src/tokens/build.mjs
```diff
@@ -59,7 +59,6 @@ function getConfig(themeName) {
         options: {
           selector: themeName === 'reapit' ? ':root, :root[data-theme="reapit"]' : `:root[data-theme="${themeName}"]`,
         },
-        transformGroup: 'web',
```
ksolanki7, line 62: Note: By defining the web, the library was converting all the numeric values to px and converting colors from rgba to hex. So removing this flag to avoid any transformation.

nurm717123 (approved):
looks good to me 👍

PR #397, opened 2025-04-15
feat: #206 (top-bar-menu-item) ship top bar menu item

ss-dimasm:
# Pull request checklist

fixes: 
- #206

### DIFF
--- a/src/components/top-bar-menu-item/top-bar-menu-item.tsx
+++ b/src/components/top-bar-menu-item/top-bar-menu-item.tsx
@@ -1,0 +1,62 @@
+import type { FC, MouseEventHandler, AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react'
+
+import {
+  ElTopBarMenuItemAnchor,
+  ElTopBarMenuItemBadge,
+  ElTopBarMenuItemButton,
+  ElTopBarMenuItemListItem,
+} from './styles'
+
+type CommonTopBarMenuItemProps = {
+  label: string
+  isActive?: boolean
+  hasBadge?: boolean
+}
+
+interface TopBarMenuItemAsButtonElementProps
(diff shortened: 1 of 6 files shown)


adrian-reapit:
Menus should use a consistent pattern. Can we follow

 <Menu.List>
      <Menu.Group label="Group Title">
           <Menu.Item onClick={console.log}>Menu Item</Menu.Item>
            <Menu.Item href="/#">Menu Item as anchor</Menu.Item>
            <Menu.Item closeMenu={false}>Menu Item (keep open)</Menu.Item>
      </Menu.Group>
</Menu.List>

src/components/top-bar-menu-item/top-bar-menu-item.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return 'onClick' in props
+}
+
+export const TopBarMenuItem: FC<TopBarMenuItemProps> = (props) => {
+  if (isTopBarMenuItemButtonElement(props)) {
+    const { isActive, label, hasBadge, 'aria-label': ariaLabel, ...rest } = props ?? {}
+
+    return (
+      <ElTopBarMenuItemListItem>
+        <ElTopBarMenuItemButton {...rest} aria-label={ariaLabel ?? label} aria-current={isActive ? 'true' : undefined}>
```
adrian-reapit, line 44: the label does not need to go in aria-label. only need an aria-label if the content is different to the label.

ss-dimasm: thanks Adrian, consumer now could pass aria-label when they need so we're not forcing to supply with that data attribute

ss-dimasm:
a valid point, now it should be used with this structure

image

PR #402, opened 2025-04-15
feat: #399 export v5 top-bar and side-bar

nurm717123:
exporting the top-bar and side-bar so others can try it early

# Pull request checklist

fixes: #


src/components/top-bar/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './styles'
 export * from './top-bar'
+export { default as MenuIcon } from './icons/menu-icon.svg?react'
```
ksolanki7, line 3: @nurm717123 **issue (non-blocking):** We should leverage the Icon element and pass in icon prop to define which icon to use for the menu, rather than directly using the SVG into the Navigation.
Reason: The new v5 Icon element will have additional props that can be utilised directly rather than writing custom CSS classes and styles when required.

ksolanki7 (approved):
I have just one non-blocking issue; we can get that resolved in any follow-up ticket or the ticket in backlog for navigation, as it's just the story and the export that requires an update. Consumer can still use the Icon element at their end.

src/components/top-bar/index.ts
nurm717123: yeah, I will note this for future works

Issue #405, opened 2025-04-16
NavItem still imported the old version [5.0.0-beta.23]


rpt-rfoxy:
**Describe the bug**
NavItem component still imported the old one and its will have gap for the first child. its look good in storybook [link](https://elements-beta.dev.paas.reapit.cloud/?path=/docs/components-top-bar--docs) but when we import the NavItem directly its imported the old version.

**Figma Design **
[Figma CRM](https://www.figma.com/design/KArHs5M7g6NkMgfTVfUaS5/Navigation-updates?node-id=1-560&m=dev)

**Related PR**
[PR](https://github.com/reapit-global/anz-sales-crm-react-app/pull/9)

**Expected behaviour**
there should no gap between first child of NavItem

**Screenshots**
![Image](https://github.com/user-attachments/assets/a6feb8d2-7ad9-4716-99b3-250af31fb413)

nurm717123:
thanks , I will deprecate the older nav and secondary nav (in this ticket) to avoid future confusion, because some of it has same component name

PR #407, opened 2025-04-16
fix: #405 Deprecate the old nav components

nurm717123:
fix the components export conflict by deprecating older `Nav` components, `NavResponsive` also deprecated to avoid future confusion

# Pull request checklist

- [x] Deprecate older `Nav` components
- [x] update package version and changelog

fixes: #


ksolanki7 (approved):
Looks good.

Issue #279, opened 2025-01-14
v5 component refactor: Tooltip viewport overflow 


ksolanki7:
The initial tooltip component has been completed in [#235](https://github.com/reapit/elements/issues/235),

The work remaining to support the positioning of the tooltip in case of overflowing over the viewport is remaining.

Until [#276](https://github.com/reapit/elements/issues/276) is addressed, please add support for dynamically positioning the tooltip if it overflows the viewport.

**Task:**

- Review the design system guidelines provided in [Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=1903-4669&m=dev), particularly focusing on handling viewport overflow scenarios.
- Implement dynamic positioning logic to ensure the tooltip adjusts its position appropriately when an overflow is detected.
- This enhancement is a temporary measure until the broader issue is resolved in [#276](https://github.com/reapit/elements/pull/276).

PR #410, opened 2025-04-21
feat: #279 Introduce dynamic display of popover element

ksolanki7:
# Pull request checklist

fixes: #279 Added helper function to calculate the position of the popover element relative to the trigger.

**Note:**

- The initial usage of this popover is done in the Tooltip component, which uses `createPortal` to position the popover element at the bottom of the body. 

**To do:**

- The current calculation only supports the positioning of the popover element for the 1 level of scrollable element, i.e., window scroll.
- Still need to provide a fix for nested scrollable elements, i.e., Usage of trigger in a scrollable container within a scrollable container. A good example would be a scrollable Table within a scrollable Page.

### DIFF
--- a/src/helpers/calculatePopoverPosition.ts
+++ b/src/helpers/calculatePopoverPosition.ts
@@ -136,32 +136,50 @@ const calculatePopoverPosition = ({
     }
   }
 
+  const adjustWhenOverflowTop = () => {
+    if (top - scrollTop < parentRect.top) {
+      top = triggerRect.bottom + scrollTop + popoverOffset
+    }
+  }
+
+  const adjustWhenOverflowBottom = () => {
+    if (top + popoverRect.height - scrollTop > parentRect.bottom) {
+      top = triggerRect.top + scrollTop - popoverRect.height - popoverOffset
+    }
+  }
+
+  const adjustWhenOverflowLeft = () => {
+    if (left - scrollLeft < parentRect.left) {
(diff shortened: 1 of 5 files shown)


src/helpers/calculatePopoverPosition.ts
```diff
@@ -0,0 +1,261 @@
...
+ * - Takes into account scrollbars and adjusts position accordingly.
+ * - Includes future TODO to use `useRef` once React 19 is adopted.
+ *
+ * @param {HTMLElement} triggerElement - The element that triggers the popover.
+ * @param {HTMLElement} popoverElement - The popover element to be positioned.
+ * @param {string} [position='bottom-start'] - Preferred position of the popover.
+ * @param {number} [popoverOffset=4] - The spacing between the trigger and popover.
+ */
+
+const calculatePopoverPosition = ({
```
nurm717123, line 54: suggestion: I think it will be hard to test this type of big helpers, In my opinion I think it can be split into two parts

1. the `calculatePopoverPosition` fn, which returning the object
2. the `setCalculatedPopoverPosition` fn, which set correct props given with required params

I think this way it will be much easier to provide unit test into it 🤔

nurm717123 (approved):
as this pr is to unblock the upcoming Menu's work by moving the reusable logic into helpers. as we've discussed internally (there will be separate ticket to address the unit test), now I think the pr is good enough to me, thanks 👍 .

Issue #409, opened 2025-04-17
v5 css variable update: Top Bar


schristiani-rpt:
**Abstract**

_As part of the v5 Elements release, there is a design token (CSS variable) update. Each existing component that still uses old tokens needs to be updated. This ticket spesifically address the **Top Bar (Navigation)** component_

**Specification**

- Is new component or existing?
- Is the component being re-named? no
- Is the component design signed off by product and design? yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=18-7578&m=dev)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

kurtdoherty:
Closed by #412

PR #412, opened 2025-04-24
chore: #409 v5 top-bar css update var

nurm717123:
## Context

_As part of the v5 Elements release, there is a design token (CSS variable) update. Each existing component that still uses old tokens needs to be updated. This ticket specifically addresses the TopBar components (and its sub components)_

also contain a css tokens file updates (generated from figma in collaboration with @ksolanki7)

# Pull request checklist
- [x] Avatar button
- [x] Nav Dropdown button
- [x] Nav Icon item
- [x] Nav item
- [x] Nav Search button
- [x] TopBar

topbar with new var


src/tokens/Semantics.Reapit.tokens.json
```diff
@@ -1,3157 +0,0 @@
-{
```
ksolanki7, line 1: @nurm717123 Seems like you accidentally deleted this file.

ksolanki7 (changes requested):

src/tokens/Semantics.Reapit.tokens.json
nurm717123: ah yes, must be some accident, thanks for catching it

nurm717123: after I re-trace it, it happen when I failed to run `yarn build:sd`, then I look at the code and I found the issue at
https://github.com/reapit/elements/blob/28334aca5e82011435c36bf77e0d9fc79c3c246e/src/tokens/build.mjs#L41
while the file we want to read is `Semantic.Reapit.tokens.json`, it failed because the `themeName` is capital case not an upper case, then after I rename the file to upper-case and the command succeed (after that I missed to discard it the changes to restore deleted file 🤦 )

ksolanki7: @nurm717123 Thanks for the feedback on the build script. I'll look into this.

ksolanki7 (approved):
Looks good.

Issue #356, opened 2025-03-24
Missing variable value on close chip icon's width & height


schristiani-rpt:
**Describe the bug**
On figma, close chip icon should has both of width and height to be: `var(--icon_size-s, 16px)`  but the actual value is  `var(--size-icon-sm)` and the variable value is not defined

**To Reproduce**
Steps to reproduce the behaviour eg:
1. Go to Chip element on story book -> https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-chip--filter-chip
2. Inspect the clear button filter
3. On devtools, go to".el-chip-clear-icon"
4. See that height & width value for variable `var(--size-icon-sm)`  is not defined

**Expected behaviour**
Icon's width and height should be `var(--icon_size-s)` and the variable should be defined with value 16 px

**Screenshots**

![Image](https://github.com/user-attachments/assets/b95c84c1-ba86-4cb1-af4f-390ab16e4aee)

**Device (please complete the following information for web issues):**
Tested on
OS: Windows 11
Browser: Chrome

kurtdoherty:
@schristiani-rpt thanks for picking this up. I suspect this regression occurred when we updated our CSS variables to match those in Figma 👍 Has now been fixed.

schristiani-rpt:
hi @kurtdoherty 
i've retested this issue and it's already correct as per figma 👍 

**Test Result**
- There's no missing variable value on `.el-chip-clear-icon`
- Correct size as per figma: `var(--icon_size-s, 16px)`

![Image](https://github.com/user-attachments/assets/d9de439b-4198-4360-b73b-aaf7ce46db0b)

Issue #386, opened 2025-04-07
Chip element: Text overflow on very long content if the content is longer than maxWidth prop


schristiani-rpt:
**Describe the bug**
On wrapped chip, if the content element is longer than `maxWidth`, the text will be overflowed out of chip container
To resolve this issue, based on [this comment on figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=3256-8443&m=dev#1126892683), we need to add logic so that `maxWidth `props is only works when truncated prop is provided

**To Reproduce**
Steps to reproduce the behaviour eg:
1. Go to [Wrapping Chip storybook](https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-chip--wrapping)
2. On storybook control, set children with a long text, for example "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
3. Notice that the text will be overflowed out of container

**Expected behaviour**
- `maxWidth` props should only works when truncated prop is provided

**Screenshots**
![Image](https://github.com/user-attachments/assets/2b273f0d-3de2-4b42-af8d-6cafc6b4e26d)

**Device (please complete the following information for web issues):**
 - Device: Laptop
 - OS: Windows 11
 - Browser: Google Chrome

kurtdoherty:
@schristiani-rpt this has been fixed now 👍 You can see this demonstrated via story here: https://elements-beta.dev.paas.reapit.cloud/?path=/story/components-chip--long-words

schristiani-rpt:
hi @kurtdoherty 

I've retested this issue and it's working fine for a long word 👍 

**Test Result**
- Test with --size-20, long word text is not overflowed ✅ 
![Image](https://github.com/user-attachments/assets/f89382f4-1605-4f6b-8445-a527c4c850fd)

- Test with --size-80, longs word text is not overlowed ✅ 
![Image](https://github.com/user-attachments/assets/4624bc0d-5575-46cc-bed6-581c79cb4150)

PR #420, opened 2025-05-08
fix: #356 #386 Chip style updates

kurtdoherty:
fixes: #356 and #386 

- Replaces incorrect icon size CSS variables with the correct ones
- Updates the Chip label styles to allow line breaks mid-word
- Adds new stories for Chip and ChipGroup demonstrating correct wrapping behaviour for long words

Screenshot 2025-05-08 at 12 19 15 pm
Screenshot 2025-05-08 at 12 18 48 pm
Screenshot 2025-05-08 at 12 18 54 pm

### DIFF
--- a/src/components/chip-group/chip-group.stories.tsx
+++ b/src/components/chip-group/chip-group.stories.tsx
@@ -141,6 +141,7 @@ export const ChipSizing: Story = {
       <ChipGroup.Item key="8" {...ChipStories.Wrapping.args}>
         Or, you can avoid truncation and allow a long chip label to wrap to multiple lines
       </ChipGroup.Item>,
+      <ChipGroup.Item key="9" {...ChipStories.LongWords.args} />,
     ],
     overflow: 'wrap',
   },
--- a/src/components/chip/chip.stories.tsx
+++ b/src/components/chip/chip.stories.tsx
@@ -79,6 +79,15 @@ export const Wrapping: Story = {
   },
 }
 
+/** Line breaks within an otherwise unbreakable string are also permitted to prevent text from overflowing the chip. */
+export const LongWords: Story = {
+  args: {
+    ...FilterChip.args,
+    children: 'Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu, NZ',
+    maxWidth: '--size-80',
+  },
+}
+
 /**
  * Truncation is an optional behaviour that can be enabled to prevent the label from
  * wrapping on multiple lines
--- a/src/components/chip/styles.ts
+++ b/src/components/chip/styles.ts
@@ -23,10 +23,10 @@ export const ElChip = styled.button<ElChipProps>`
   gap: var(--spacing-2);
   grid-template-columns: auto min-content;
   height: min-content;
+  max-width: var(--chip-max-width, auto);
   padding-block: var(--spacing-1);
   padding-inline: var(--spacing-4) var(--spacing-2);
   width: fit-content;
-  max-width: var(--chip-max-width, auto);
 
   &[aria-disabled='true'] {
     cursor: not-allowed;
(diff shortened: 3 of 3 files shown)


src/components/chip-group/chip-group.stories.tsx
```diff
@@ -141,6 +141,7 @@ export const ChipSizing: Story = {
       <ChipGroup.Item key="8" {...ChipStories.Wrapping.args}>
         Or, you can avoid truncation and allow a long chip label to wrap to multiple lines
       </ChipGroup.Item>,
```
kurtdoherty, line 143: **note:** using the new Chip story in this existing ChipGroup story.

src/components/chip/chip.stories.tsx
```diff
@@ -79,6 +79,15 @@ export const Wrapping: Story = {
   },
 }
```
kurtdoherty, line 81: **note:** new story demonstrating how long words are now broken if they will otherwise overflow

src/components/chip/styles.ts
```diff
@@ -23,10 +23,10 @@ export const ElChip = styled.button<ElChipProps>`
   gap: var(--spacing-2);
   grid-template-columns: auto min-content;
   height: min-content;
```
kurtdoherty, line 25: **note:** 🪄 Alphabetise 🪄

src/components/chip/styles.ts
```diff
@@ -74,6 +74,9 @@ interface ElChipLabelProps {
 export const ElChipLabel = styled.span<ElChipLabelProps>`
   color: var(--text-primary);
```
kurtdoherty, line 76: **note:** what the code comment says. This allows long words to be broken. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap). Note that the [word-break: break-word](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) property is deprecated, which is why it has not been used.

src/components/chip/styles.ts
```diff
@@ -99,8 +102,8 @@ export const ElChipClearIcon = styled(Icon)`
    * does not allow consumer-supplied styles to have a higher specificity */
   color: var(--comp-chip-colour-icon-active) !important;
   font-size: 1rem;
```
kurtdoherty, line 104: **note:** fixing the use of invalid CSS variables.

src/components/chip/styles.ts
paulbevis: Looking at the MDN link you provided that has 3 examples, interacting with them `overflow-wrap: break-word;` would appear to give us the designed effect, whereas the choice you have made `overflow-wrap: anywhere;`, doesn't seem to work as well in their example. But works perfectly in the storybook examples you have provided

I guess that's because of our max-width (80)?

I couldn't make the LongWords story change much though, even when I removed the max width value 🤷

paulbevis: Update looks good though :)

paulbevis (approved):
Changes look good! Just the one non blocking question for me!

Code checked out fine, and I was able to run storybook locally, so all good (fr me)

src/components/chip/styles.ts
kurtdoherty: Yeah, it's a bit tricky to know which is going to be best... `overflow-wrap: break-word` didn't seem to do much when I tested it; the long word remained unbroken in all cases 🤷 I landed on `overflow-wrap: anywhere` based on it seeming to prefer placing words on new lines rather than breaking them after only a single character.

Screenshot 2025-05-13 at 8 55 16 am
Screenshot 2025-05-13 at 8 55 50 am

I think the main difference here is I'm applying the `overflow-wrap` to a span of text that includes multiple words, whereas the MDN docs are applying it to a single word.

PR #451, opened 2025-05-29
chore: #443 refactor sidebar components and update styles (part 2)

kurtdoherty:
### Context
- We somehow managed to use component names that don't match what Design are using 🤦 ;
- There's some styles that aren't quite right (e.g. focus outline, SideBar doesn't fill vertical space);
- The components are consistently available via SideBar itself (i.e. inconsistent application of the compound component pattern);
- State management is a little buggy (e.g. an active menu group can be collapsed)

### Scope
- Refactor component names/responsibilities to match Design
- Fix up styles to ensure correct visual behaviour
- Apply compound component pattern consistently
- Fix bugs in state management
- Improve Storybook documentation for all the components involved in the SideBar

### Preceding PRs
- #450 

### This PR

- Adds a new `SideBarSubmenu` component. This includes a new subcomponent that wraps the `SideBarSubmenuItem` introduced in #450 in an `<li>` element.

Screenshot 2025-05-29 at 12 49 38 pm
Screenshot 2025-05-29 at 12 49 52 pm
Screenshot 2025-05-29 at 12 49 59 pm

PR #450, opened 2025-05-28
chore: #443 refactor sidebar components and update styles (part 1)

kurtdoherty:
### Context
- We somehow managed to use component names that don't match what Design are using 🤦 ;
- There's some styles that aren't quite right (e.g. focus outline, SideBar doesn't fill vertical space);
- The components are consistently available via SideBar itself (i.e. inconsistent application of the compound component pattern);
- State management is a little buggy (e.g. an active menu group can be collapsed)

### Scope
- Refactor component names/responsibilities to match Design
- Fix up styles to ensure correct visual behaviour
- Apply compound component pattern consistently
- Fix bugs in state management
- Improve Storybook documentation for all the components involved in the SideBar

### This PR

- Adds a new `SideBarContext` object and associated hook that is used to provide the side bar's state and state setters to child components. This replaces the existing `IsSideBarExpandedContext`, which was unfortunately responsible for providing more than just an `isSideBarExpanded` boolean.
- Adds a new `useSideBar` hook that manages the side bar's collapsed/expanded state.
- Adds a new `SideBarSubmenuItem` component which matches the [Figma component with the same name](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=12295-18961&m=dev). This replaces the existing `SideBarMenuGroupItem`.
- Adds dedicated storybook-based documentation for this new submenu item.
- Adds some new storybook decorators that will be used by many of the side bar component stories.

Screenshot 2025-05-29 at 12 26 44 pm
Screenshot 2025-05-29 at 12 26 56 pm
Screenshot 2025-05-29 at 12 27 08 pm
Screenshot 2025-05-29 at 12 27 17 pm

Issue #443, opened 2025-05-21
chore: refactor SideBar component names and fix styles


kurtdoherty:
### Context

1. We somehow managed to use component names that don't match what Design are using 🤦 ;
2. There's some styles that aren't quite right (e.g. focus outline, SideBar doesn't fill vertical space);
3. The components are consistently available via `SideBar` itself (i.e. inconsistent application of the compound component pattern);
4. State management is a little buggy (e.g. an active menu group can be collapsed)

### Scope

- Refactor component names/responsibilities to match Design
- Fix up styles to ensure correct visual behaviour
- Apply compound component pattern consistently
- Fix bugs in state management
- Improve Storybook documentation for all the components involved in the SideBar

PR #452, opened 2025-05-29
chore: #443 refactor sidebar components and update styles (part 3)

kurtdoherty:
### Context
- We somehow managed to use component names that don't match what Design are using 🤦 ;
- There's some styles that aren't quite right (e.g. focus outline, SideBar doesn't fill vertical space);
- There's some behaviour that isn't quite right (e.g. can close an active menu group)
- The components are consistently available via SideBar itself (i.e. inconsistent application of the compound component pattern);
- State management is a little buggy (e.g. an active menu group can be collapsed)

### Scope
- Refactor component names/responsibilities to match Design
- Fix up styles to ensure correct visual behaviour
- Apply compound component pattern consistently
- Fix bugs in state management
- Improve Storybook documentation for all the components involved in the SideBar

### Preceding PRs
- #450 
- #451 

### This PR

- ...


src/components/side-bar/menu-item/menu-item.tsx
kurtdoherty, line 1: **note:** simple, presentational menu item that will replace the existing menu item implementation. Similar to the submenu item introduced in #450, this component only handles rendering the `<a>` element. A separate component will be introduced in a future PR that will wrap this in an `<li>` element.

src/components/side-bar/menu-item/__test__/menu-item.test.tsx
```diff
@@ -0,0 +1,36 @@
+import { composeStories } from '@storybook/react'
+import { render, screen } from '@testing-library/react'
+import * as stories from '../menu-item.stories'
+import { elSideBarMenuItem } from '../styles'
+
+const MenuItemStories = composeStories(stories)
```
ksolanki7, line 6: **praise:** This is good, we should let other developers know about it too. I didn't know that we could do this.

src/components/side-bar/menu-item/menu-item.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+} satisfies Meta<typeof SideBarMenuItem>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Example: Story = {
+  args: {
+    children: 'Menu item',
+    href: globalThis.top?.location.href!,
```
ksolanki7, line 36: **praise:** I like the level of detailing here with optional chaining and non-null assertion.

src/components/side-bar/menu-item/menu-item.tsx
ksolanki7: **question:** I may need more context here for the approach and why we are separating this from the existing menu.
- Is the design different?
- Is the approach to submenu and its requirement different from the actual Menu component?

ksolanki7 (approved):

Issue #316, opened 2025-02-14
v5 component refactor: Menu


HollyJoyPhillips:
_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? New
- Is the component being re-named? N/A
- Is the component design signed off by product and design? Yes
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF/Reapit-Elements-production-ready-components?node-id=3772-3618&m=dev)
- [Accessibility requirements related doc](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

nurm717123:
Hi @adrian-reapit and @kurtdoherty ,  I’ve reviewed the design spec and come up with the following solution:

### Context
- the previous Menu #163 will not be used (which not released yet to production)
- this will be the new Menu component replacing #163 
- the new Menu basically has 2 version of item (let say for now, simple/text and the complex one)
- the new Menu have almost the same scrolling behaviour with the new `Tooltip`

### Proposal
Given the context above I propose to divide the work into 3 main work:
1. Deprecate previous component (there is still a few beta consumer that using it)
2. New Menu - Simple version (will be broken down in the following section)
3. implement scrolling behaviour  using the reusable `helpers`
4. New Menu - Complex version (will be broken down in the following section)
5. support a11y

### Component Interface
I think that the existing interface is still good for the new component, i.e
```javascript
<Menu data-alignment="right">
  <Menu.Trigger>
[...]

PR #461, opened 2025-06-03
feat: #316 new Menu - support maxHeight and adjust MenuGroup style

nurm717123:
## Context

As there is redesign in new v5 Menu #316 . this pr is to refactor existing meu based on a POC that has been discussed [here](https://github.com/reapit/elements/issues/316#issuecomment-2853782912) which focusing on adding support for `maxHeight` and also adjusting the content so the `MenuItemGroupTitle` not being part of scrollable content [reference](https://www.figma.com/design/XJ6qcAV8gHscsUodqJMNEF?node-id=4747-25373&m=dev#1240491371)

note: some of big changes is from snapshots

# Pull request checklist
- [x] add `maxHeight` props to `Menu.List` and `Menu.Group`
- [x] update storybook example
- [x] update changelog 
 
menu title   max-height


ksolanki7 (approved):
Looks good

PR #467, opened 2025-06-05
feat: #218 added v5 Table Row Selection Component with Table Provider

DanishAli-Reapit:
# Description

- Created the v5 Table Row Selection Component
- Created a Table Provider Component for state management.

https://github.com/user-attachments/assets/d5ee8b0f-bda5-4542-90e4-9cb8e46d03d7


DanishAli-Reapit:
> Thanks @DanishAli-Reapit 👏 I think this is good enough for now. There's probably some improvements we could make, but I'd prefer to defer those until I sweep through all the table components holistically.

Thanks @kurtdoherty for the feedback.

Issue #474, opened 2025-06-05
bug: `SideBar` list elements affected by user agent styles


kurtdoherty:
**Describe the bug**
SideBar lists (main and submenu) have user agent `margin-block` and `padding-inline` styles that break the layout when used in Reapit PM.

**Expected behaviour**

User agent styles wouldn't impact the SideBar's visual appearance.

**Screenshots**

<img width="414" alt="Image" src="https://github.com/user-attachments/assets/299037b7-3ebf-4f53-85c0-6e7c2431a809" />

PR #475, opened 2025-06-05
fix: #474 `SideBar` list elements no longer impacted by user-agent styles

kurtdoherty:
fixes: #474

**note:** I have no idea why these user agent styles don't surface in Storybook. When I inspect the side bar list elements in Storybook, the same user agent styles are being applied, but they seem to be having no effect 🤷 Will need to dig into this later I think.


paulbevis (approved):
Cool! Thanks for fixing that up!

PR #488, opened 2025-06-10
chore: #483 refactor `AvatarButton` into top bar folder

kurtdoherty:
### Context

- The Top Bar components are all separated in their own folders. This was largely done because they were implemented in isolation from each other.
- Now, we want to co-locate them all within the `/top-bar` folder.
- We also want to apply the same patterns used in the `SideBar` to these `TopBar` components.

### Scope of work in #483

- Move `/nav-item` to `/top-bar/nav-item`
- Move `/nav-dropdown-item` to `/top-bar/nav-dropdown-item`
- Move `/nav-icon-item` to `/top-bar/nav-icon-item`
- Move `/avatar-button` to `/top-bar/avatar-button`
- Add new `TopBar.MainNav` component in `/top-bar/main-nav`
- Add new `TopBar.SecondaryNav` component in `/top-bar/secondary-nav`
- Update docs for all `TopBar` components

### This PR

Part 4 of #483.
Follows on from #485, #486, and #487.

- Moves `/avatar-button` to `/top-bar/avatar-button` and renames it to `TopBarAvatarButton`
- Scouts style updates to the depended-on `Avatar` component
- Updates testing approach for the component
- Updates Storybook docs for the component
- Updates usage of the component in the `AppSwitcher` implementation and the `TopBar` stories

Screenshot 2025-06-10 at 4 44 21 pm
Screenshot 2025-06-10 at 4 44 37 pm

PR #487, opened 2025-06-10
chore: #483 refactor `NavIconItem` into top bar folder

kurtdoherty:
### Context

- The Top Bar components are all separated in their own folders. This was largely done because they were implemented in isolation from each other.
- Now, we want to co-locate them all within the `/top-bar` folder.
- We also want to apply the same patterns used in the `SideBar` to these `TopBar` components.

### Scope of work in #483

- Move `/nav-item` to `/top-bar/nav-item`
- Move `/nav-dropdown-item` to `/top-bar/nav-dropdown-item`
- Move `/nav-icon-item` to `/top-bar/nav-icon-item`
- Add new `TopBar.MainNav` component in `/top-bar/main-nav`
- Add new `TopBar.SecondaryNav` component in `/top-bar/secondary-nav`
- Update docs for all `TopBar` components

### This PR

Part 2 of #483.
Follows on from #485 and #486.

- Moves `/nav-icon-item` to `/top-bar/nav-icon-item` and renames it to `TopBarNavIconItem`
- Updates testing approach for the component
- Updates Storybook docs for the component
- Updates usage of the component in the `AppSwitcher` implementation and the `TopBar` stories

Screenshot 2025-06-10 at 3 23 16 pm
Screenshot 2025-06-10 at 3 23 25 pm
Screenshot 2025-06-10 at 3 25 25 pm
Screenshot 2025-06-10 at 3 24 01 pm
Screenshot 2025-06-10 at 3 23 53 pm

PR #486, opened 2025-06-10
chore: #483 refactor `NavDropdownButton` into top bar folder

kurtdoherty:
### Context

- The Top Bar components are all separated in their own folders. This was largely done because they were implemented in isolation from each other.
- Now, we want to co-locate them all within the `/top-bar` folder.
- We also want to apply the same patterns used in the `SideBar` to these `TopBar` components.

### Scope of work in #483

- Move `/nav-item` to `/top-bar/nav-item`
- Move `/nav-dropdown-item` to `/top-bar/nav-dropdown-item`
- Move `/nav-icon-item` to `/top-bar/nav-icon-item`
- Add new `TopBar.MainNav` component in `/top-bar/main-nav`
- Add new `TopBar.SecondaryNav` component in `/top-bar/secondary-nav`
- Update docs for all `TopBar` components

### This PR

Part 2 of #483

- Moves `/nav-dropdown-item` to `/top-bar/nav-dropdown-item` and renames it to `TopBarNavDropdownButton`
- Updates testing approach for the component
- Updates Storybook docs for the component

Screenshot 2025-06-10 at 2 19 10 pm
Screenshot 2025-06-10 at 2 19 20 pm
Screenshot 2025-06-10 at 2 19 27 pm
Screenshot 2025-06-10 at 2 19 34 pm

Issue #483, opened 2025-06-09
chore: refactor `TopBar` subcomponents inside `/top-bar` folder


kurtdoherty:
**Background context:**

- The Top Bar components are all separated in their own folders. This was largely done because they were implemented in isolation from each other.
- Now, we want to co-locate them all within the `/top-bar` folder.
- We also want to apply the same patterns used in the `SideBar` to these `TopBar` components.

**Specification or tasks to perform:**

- Move `/nav-item` to `/top-bar/nav-item`
- Move `/nav-dropdown-item` to `/top-bar/nav-dropdown-item`
- Move `/nav-icon-item` to `/top-bar/nav-icon-item`
- Move `/avatar-button` to `/top-bar/avatar-button`
- Add new `TopBar.MainNav` component in `/top-bar/main-nav`
- Add new `TopBar.SecondaryNav` component in `/top-bar/secondary-nav`
- Update docs for all `TopBar` components

PR #491, opened 2025-06-10
chore: #483 Add new `MainNav` component to top bar

kurtdoherty:
### Context

- The Top Bar components are all separated in their own folders. This was largely done because they were implemented in isolation from each other.
- Now, we want to co-locate them all within the `/top-bar` folder.
- We also want to apply the same patterns used in the `SideBar` to these `TopBar` components.

### Scope of work in #483

- Move `/nav-item` to `/top-bar/nav-item`
- Move `/nav-dropdown-item` to `/top-bar/nav-dropdown-item`
- Move `/nav-icon-item` to `/top-bar/nav-icon-item`
- Move `/avatar-button` to `/top-bar/avatar-button`
- Add new `TopBar.MainNav` component in `/top-bar/main-nav`
- Add new `TopBar.SecondaryNav` component in `/top-bar/secondary-nav`
- Update docs for all `TopBar` components

### This PR

Part 5 of #483.
Follows on from #485, #486, #487, #488.

- Adds new `TopBar.MainNav` component. This component takes the same approach as `SideBar.Submenu` and `SideBar.MenuList` insomuch as it provides `<li>` wrapped versions of `TopBarNavItem` and `TopBarNavDropDownButton`.
- Updates `TopBar` to render as a `<header>` instead of a `<nav>`. This is because `<header>`, when used in the manner we are, has the `banner` role, which matches what the Top Bar is for our product UI. See MDN's [header](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/banner_role) element and [banner](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/banner_role) role docs.

Screenshot 2025-06-11 at 7 17 20 am
Screenshot 2025-06-11 at 7 17 28 am
Screenshot 2025-06-11 at 7 17 35 am


src/components/top-bar/main-nav/styles.ts
```diff
@@ -0,0 +1,24 @@
+import { styled } from '@linaria/react'
+
+export const ElTopBarMainNav = styled.nav`
```
kurtdoherty, line 3: **note:** these padding and container properties were originally handled by the `ElTopBarMainNavContainer` styled element, but now that we have this dedicated MainNav component, it makes more sense to have them here.

src/components/top-bar/top-bar.tsx
```diff
@@ -65,7 +67,11 @@ export function TopBar({ appSwitcher, avatar, logo, mainNav, menu, search, secon
   )
 }
```
kurtdoherty, line 69: **note:** expose the main nav component and its subcomponents via TopBar. Also add in the missing dropdown button atom

PR #499, opened 2025-06-13
chore: Update CSS variables

kurtdoherty:
Updates the design tokens/CSS variables to match the latest values in Figma.


src/tokens/Semantics.PayProp.tokens.json
```diff
@@ -1269,1111 +1269,927 @@
     }
   },
   "comp": {
```
kurtdoherty, line 1271: **note:** unfortunately, design have re-ordered some of the variable folders in Figma, so we're getting messy diffs for our JSON files 😞 This makes it really hard for us to review, so I think I'll add an item of work to see if we can easily preprocess the JSON we get from Figma so it has alphabetically ordered keys or something...that way, we could ensure the diffs are much nicer

ksolanki7 (approved):

Issue #500, opened 2025-06-13
chore: alphabetically sort keys in token files


kurtdoherty:
**Background context:**

Every now and then, we need to export design tokens/variables from Figma to Elements. This is a reasonably trivial process, except that Design can sometimes reorder the variables in Figma. When they do, this causes massive diffs (see #499 for a recent example) and the massive diffs make it really hard to detect errors in the Figma variables.

**Specification or tasks to perform:**

Our copy of the variables (currently stored across three files: `Primitives.Values.tokens.json`, `Semantics.PayProp.tokens.json` and `Semantics.Reapit.tokens.json`) could be preprocessed prior to a PR being raised so that the keys in the JSON object are alphabetically sorted.

This would mean we are decoupled from Figma's order, and thus we'd be able to ensure better diffs in the subsequent PRs.

PR #501, opened 2025-06-13
chore: #500 preprocess token files

kurtdoherty:
### Context

Every now and then, we need to export design tokens/variables from Figma to Elements. This is a reasonably trivial process, except that Design can sometimes reorder the variables in Figma. When they do, this causes massive diffs (see https://github.com/reapit/elements/pull/499 for a recent example) and the massive diffs make it really hard to detect errors in the Figma variables.

### This PR

**fixes:** #500 

- Adds a new `preprocess.js` module that reads in all `.tokens.json` files within it's directory, sorts all keys alphabetically, and writes the changes back to the file.
- Updates the `package.json` script, `build:sd`, to now be named `build:tokens` and to run the new `build:tokens:preprocess` script _before_ running the `build:tokens:css`.


package.json
```diff
@@ -50,7 +50,9 @@
     "build": "yarn build:storybook & yarn build:types & yarn build:dist",
     "build:storybook": "rimraf public/dist && NODE_ENV=production storybook build -o public/dist",
     "build:dist": "vite build",
```
kurtdoherty, line 52: **note:** by ensuring the preprocessing happens as part of the manual build step devs must do to generate the CSS files, nothing needs to change w.r.t the process we follow to update the tokens 🥳

src/tokens/preprocess.js
```diff
@@ -0,0 +1,73 @@
+import path from 'node:path'
+import url from 'node:url'
+import fs from 'node:fs/promises'
+import util from 'node:util'
+
```
kurtdoherty, line 5: **note:** here's the new script that we run in the new `yarn build:tokens:preprocess` script.

This script was written by Claude and reviewed/modified by Gemini. All I did was verify it all worked and tweaked the console logging.

Issue #145, opened 2024-07-30
v5 component refactor: Drawer


HollyJoyPhillips:
**Abstract**

_As part of the v5 Elements release, each component will be reviewed and refactored to ensure best practice and design system alignment_

**Specification**

- Is new component or existing? New
- Is the component being re-named? New 'Drawer' component 
- Is the component design signed off by product and design? Yes - but UI is still being updated "The UI is ready to be updated - but the usage and behaviour is still being documented. Core UI team are working on this at the moment - i.e. drawer stacking, URLs etc."
- [Guide to component structure & documentation in storybook](https://github.com/reapit/elements/wiki/Developer-Contribution-Guide)
- [Link to DS in Figma](https://www.figma.com/design/6CaivqdlTX0UkFYJkpBKDu/Reapit-DS?node-id=428-7762&t=dnxaJYB5fA7ExrG5-4)
- [Link to Elements Audit in Figma](https://www.figma.com/design/J6EneNGfpk2GcUuymiOKW4/Elements-Gaps-%2B-Discrepancies?node-id=284-5489&t=Y1ennycgce64B1Qq-4)
- [Accessibility requirements](TBC) 

**Developer Checklist**

- [ ] Styles alignment between Design System and Elements
- [ ] Check design tokens in Figma and implement CSS variable tokens if available for relevant component
- [ ] Align with accessibility standards / spec as per above
- [ ] If relevant, break down component into `Styles Only` and `React` component structures
- [ ] Ensure all variants of components are documented as appropriate 
- [ ] Ensure unit test coverage is adequate for component
- [ ] Update documentation in MDX file as per guidelines
- [ ] Changelog updated to reflect a single beta version per component ideally

**Release Checklist**

- [ ] Approved PR merged to main
- [ ] Design & product review and feedback addressed by developer
- [ ] Beta release by product / engineering lead to next beta version

**Additional Context or Information**

kurtdoherty:
**suggestion:** It would be great if we could align Element's Drawer component with the Console Cloud team's `react-ui-core` [Drawer](https://dev.saas-dev.console.com.au/client-apps/design-system/react-ui-core/?path=/docs/overlay-drawer--docs), as that will make migration for us _much_ simpler. The component API we have has also been battled tested via 100s of drawers in Console Cloud, so we're pretty confident about it's ability to service the drawer pattern defined by the Design System.

[...]

PR #516, opened 2025-06-18
chore: #145 Deprecate existing `Drawer` component

kurtdoherty:
### Context

- We're implementing new a `Drawer` component that will not be backwards-compatible with the existing one.
- The existing `Drawer` needs to be deprecated and moved.

### This PR

Part 1 of #145

- Deprecates the existing Drawer component, renaming it and all it's related components and classes to include `deprecated` in their name.


src/index.ts
```diff
@@ -13,12 +13,20 @@ export * from './components/button'
 export * from './components/button-group'
 export * from './components/split-button'
 export * from './components/card'
```
kurtdoherty, line 15: **note:** small scout to not just update the export for the now deprecated drawer, but also group all the other exports of deprecated components

src/index.ts
```diff
@@ -32,28 +40,20 @@ export * from './components/secondary-nav'
...
 export * from './components/accordion'
-export * from './components/deprecated-tag'
 export * from './components/tag'
-export * from './components/deprecated-badge'
-export * from './components/deprecated-chip'
 export * from './components/tile'
-export * from './components/deprecated-avatar'
 export * from './components/key-value-list'
 export * from './components/page-header'
 export * from './components/deprecated-breadcrumb'
```
ksolanki7, line 59: **issue(blocking):** I believe there are a few elements that have not been exported yet - for example, `chip` and `chip-group`. This currently prevents consumers from using these components. Could we get these exported correctly as well as part of this PR or do you want to play it separately?

kurtdoherty: I'll do it separately 👍 Raising a PR now.

ksolanki7 (approved):

PR #526, opened 2025-06-19
chore: bump Vite, Vitest, TypeScript, Linaria et al

kurtdoherty:
Bumps the following dependencies:
- **Vite:** v5 → v6. No relevant breaking changes for us. Also upgraded `@vitejs/plugin-react`.
- **Vitest:** v2 → v3. No relevant breaking changes for us. Also upgraded `@vitest/coverage-v8`.
- **TypeScript:** v5.6 → v5.8
- **React Testing Library:** v15 → v16. The `renderHook` utility is now part of RTL itself, so we no longer need `@testing-library/react-hooks`. New peer dependency on `@testing-library/dom`.
  - This (and potentially the Happy DOM upgrade) have resulted in some test updates, particular around snapshots (invalid DOM warnings) and the use of `act` in some tests.
  - There are also some type-related changes in some tests using `renderHook`. Specifically, we no longer need to specify any generic types.
- **Happy DOM:** v15 → v18. No relevant breaking changes for us.
- **Linaria (and related libs):** v6.2 → v6.3. The related `@wyw-in-js/vite`, `@wyw-in-js/babel-preset` and the `@babel/...` dependencies have all been bumped as well.

I also removed the use of `vite-plugin-dts` entirely. We are building our TS types separately (via `tsc`), so there's no need to have Vite emit types as well. Our package exports now ensure the types point to what `tsc` emits.


package.json
```diff
@@ -25,7 +25,7 @@
   },
   "exports": {
     ".": {
```
kurtdoherty, line 27: **note:** now that we're no longer emit types via both Vite and `tsc`, we want our types export to point to the `dist/types` files (which is what `tsc` emits)

src/components/avatar-rectangle/__tests__/__snapshots__/avatar-rectangle.test.tsx.snap
```diff
@@ -102,31 +102,29 @@ exports[`AvatarRectangle > renders correctly with residential variant and small
 
 exports[`AvatarRectangle > simulate error and renders correctly when image fails to load 1`] = `
 <DocumentFragment>
```
kurtdoherty, line 104: **note:** the original snapshot wasn't correct. If an error occurs when loading the image, we should display one of the placeholder SVGs. That wasn't happening correctly in this test. Now that we've wrapped the dispatch of the error event in `act`, we can now see the component is behaving correctly and rendering one of the SVG placeholders.

src/components/avatar-rectangle/__tests__/avatar-rectangle.test.tsx
```diff
@@ -30,11 +30,13 @@ describe('AvatarRectangle', () => {
     const { asFragment } = render(
       <AvatarRectangle variant="commercial" size="medium" src="invalid-image.jpg" alt="Test Image" />,
     )
```
kurtdoherty, line 32: **note:** since there was no `act`, the snapshot was effectively being taken _before_ the component re-render after catching the error.

src/components/bottom-bar/__tests__/use-bottom-bar-visibility.test.ts
```diff
@@ -1,6 +1,6 @@
 import { RefObject } from 'react'
 import { handleChangeBottomBarVisibility, useBottomBarVisibility } from '../use-bottom-bar-visibility'
```
kurtdoherty, line 2: **note:** you'll see a number of import updates like this for any tests that were using the `renderHook` utility

src/components/menu/__tests__/use-menu.test.tsx
```diff
@@ -1,46 +1,57 @@
-import { renderHook } from '@testing-library/react-hooks'
+import { act, renderHook } from '@testing-library/react'
 import { useMenu } from '../use-menu'
 
 describe('useMenu', () => {
   it('should open menu when openMenu is called', () => {
     const { result } = renderHook(() => useMenu())
```
kurtdoherty, line 7: **note:** need to wrap this kind of stuff in `act` as it causes a state change and we need React to re-render correctly before we make any assertions.

src/components/multi-select/__tests__/__snapshots__/multi-select.test.tsx.snap
```diff
@@ -49,6 +49,7 @@ exports[`MultiSelectInput > should match a snapshot and render children 1`] = `
       <input
         aria-checked="true"
         aria-hidden="true"
```
kurtdoherty, line 51: **note:** not 100% sure on the cause for `checked` showing up in the snapshot now, but I assume it's either Happy DOM or React Testing Library related.

src/components/reapit-logo/__tests__/__snapshots__/reapit-logo.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`Reapit logo > should match snapshot 1`] = `
 <DocumentFragment>
   <img
```
kurtdoherty, line 5: **note:** the use of the data URL is what we expect in our production build. I think Vite 5.x or Vitest 2.x behaved differently within a test environment. Post upgrade of both Vite and Vitest, we now see consistent behaviour in test as we expect in our production build, which is nice!

src/components/select/__tests__/__snapshots__/select.test.tsx.snap
```diff
@@ -2,6 +2,9 @@
 
 exports[`Select component > should match a snapshot 1`] = `
 <DocumentFragment>
```
kurtdoherty, line 4: **note:** original snapshot was a bit borked. Again, I suspect Happy DOM or React Testing Library is now behaving better

src/components/table/table-body/__tests__/__snapshots__/table-body.test.tsx.snap
```diff
@@ -2,10 +2,16 @@
 
 exports[`TableBody > should match snapshot 1`] = `
 <DocumentFragment>
```
kurtdoherty, line 4: **note:** lots of the table atom snapshots were spitting out invalid DOM warnings. They are now all fixed.

vite.config.ts
```diff
@@ -12,6 +11,7 @@ export default defineConfig({
     copyPublicDir: false,
     emptyOutDir: true,
     lib: {
```
kurtdoherty, line 13: **note:** post Vite, Vitest and wyw-in-js upgrades, the CSS dist file was now `elements.css`. This config option gets it back to the existing `style.css`.

vite.config.ts
```diff
@@ -57,7 +57,6 @@ export default defineConfig({
         presets: ['@babel/preset-typescript', '@babel/preset-react'],
       },
     }),
```
kurtdoherty, line 59: note: We don't need type declaration files to be emitted by Vite. We have a separate build script that runs `tsc` to get these, so we can save a bunch of time on the actual JS production build by dropping `dts` entirely.

src/components/reapit-logo/__tests__/__snapshots__/reapit-logo.test.tsx.snap
paulbevis: That is good, although a little more verbose ;)

PR #645, opened 2025-07-30
fix: border radius on Menu

kurtdoherty:
The new Menu's border radius is not working as expected. The popover's background is not transparent as intended, perhaps because of the box-shadow? 🤷 Either way, we need to shift the border-radius from the ElMenuContent styled element to the popover element itself.

Screenshot 2025-07-31 at 8 22 28 am

### DIFF
--- a/src/core/menu/menu.stories.tsx
+++ b/src/core/menu/menu.stories.tsx
@@ -1,5 +1,5 @@
 import { Badge } from '#src/core/badge'
-import { getMenuTriggerProps, Menu } from './menu'
+import { Menu } from './menu'
 import { StarIcon } from '#src/icons/star'
 import { useId } from 'react'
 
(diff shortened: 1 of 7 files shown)


src/core/menu/menu.stories.tsx
```diff
@@ -110,7 +110,7 @@ const meta = {
           aria-controls={props.id}
           aria-haspopup="menu"
           id={props['aria-labelledby']}
```
kurtdoherty, line 112: **note:** small scout. Making this helper a property of the Menu component means its future deprecation and removal will not require any changes to import statements in consumers.

src/core/menu/menu.tsx
```diff
@@ -1,14 +1,12 @@
 import { AnchorMenuItem, MenuItem } from './item'
 import { ElMenuContent } from './styles'
+import { getPopoverTriggerProps as getMenuTriggerProps } from '#src/utils/popover'
 import { MenuDivider } from './divider'
 import { MenuGroup } from './group'
 import { Popover } from '#src/utils/popover'
 
 import type { HTMLAttributes } from 'react'
```
kurtdoherty, line 9: **note:** this is now being exposed as a property on the Menu component

PR #649, opened 2025-07-31
chore: Deprecate `Tooltip`

kurtdoherty:
### Context

- We want to leverage the new Popover utility component for our tooltips.
- The existing Tooltip component has a number of API choices that make it awkward to maintain backwards compatibility with if it's internal implementation moves to Popover.
- Thus we want to deprecate the existing Tooltip and implement a new replacement.

### This PR

- Deprecates `Tooltip` and renames it to `DeprecatedTooltip`. All its classes/styled elements have also been renamed and it's now part of the `@reapit/elements/deprecated/*` entry point.

PR #651, opened 2025-07-31
chore: use new `Tooltip` in core components

kurtdoherty:
### Context

- We want to leverage the new Popover utility component for our tooltips.
- The existing Tooltip component has a number of API choices that make it awkward to maintain backwards compatibility with if it's internal implementation moves to Popover.
- Thus we want to deprecate the existing Tooltip and implement a new replacement.
- Previous work includes:
  - #649 
  - #650 

### This PR

- Uses new `Tooltip` component in all core components that require tooltips.
- Scouts some test improvements and name changes for the top bar nav icon item components and replaces the top bar's  snapshot test with "DOM interrogation" tests.


src/core/badge/badge.tsx
```diff
@@ -28,24 +28,41 @@ export function Badge({
...
-  const tooltip = useDeprecatedTooltip()
+  const badgeId = id ?? useId()
+  const tooltipId = useId()
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- We don't need the aria-describedby prop
-  const { 'aria-describedby': _, ...triggerProps } = tooltip.getTriggerProps(rest)
+  // For icon-only badges, the `aria-label` will be used as the tooltip's text rather than the
+  // badge's `aria-label` label because the tooltip will label the badge via `aria-labelledby`.
+  // For badges with textual content, we just pass the `aria-label` through untouched.
+  //
```
kurtdoherty, line 45: **note:** as the code comment below highlights, there's a bit of awkwardness going on here with the component interface.

I think it's fine for now. I'm 60/40 on whether:

```tsx
<Badge aria-label="Badge" icon={...} />
```

shows clearer intent than:

```tsx
<Badge icon={...} isIconOnly>Badge</Badge>
```

I think I lean slightly towards `aria-label` because its "close to the web" and it's good for devs to think about ARIA labels for elements that have no visual label (even though in this case it technically does).

src/core/features/item/item.tsx
```diff
@@ -14,20 +15,17 @@ interface FeaturesItemProps extends HTMLAttributes<HTMLDivElement> {
...
-  // interferes with the a11y.
-  const { getTooltipProps, getTriggerProps } = useDeprecatedTooltip()
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- we don't want aria-describedby on the trigger
-  const { 'aria-describedby': _, ...triggerProps } = getTriggerProps(rest)
+export function FeaturesItem({ icon, id, label, value, ...rest }: FeaturesItemProps) {
+  const tooltipId = useId()
+  const triggerId = id ?? useId()
 
   return (
```
kurtdoherty, line 22: **note:** no use of `Tooltip.getTooltipTriggerProps` here because the trigger element is separate to the element labelled by the tooltip.

src/core/top-bar/nav-icon-item/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from './nav-icon-item-anchor'
+export * from './nav-icon-item'
```
kurtdoherty, line 1: **note:** small scout to align this with other button/anchor component pairs insomuch as the main component in the pair has no element type in its name.

PR #655, opened 2025-08-01
chore: upgrade Storybook, Vite and TypeScript

kurtdoherty:
What it says on the tin.

Also updates the Storybook config to collapse the "Deprecated" root node by default.


.storybook/manager.ts
```diff
@@ -4,6 +4,9 @@ import { create } from 'storybook/theming'
 import brandImage from './public/reapitLogo.svg'
 
 addons.setConfig({
```
kurtdoherty, line 6: **note:** this makes the sidebar far more approachable

PR #647, opened 2025-07-31
fix: Avoid empty classes/styled elements

kurtdoherty:
Linaria 0.7.0 does not play nicely with empty classes/styled elements, sometimes throwing build errors like `[wyw-in-js] Cannot use 'in' operator to search for 'initialCode' in undefined`. There is an [issue raised](https://github.com/Anber/wyw-in-js/issues/144) and a [PR to fix it](https://github.com/Anber/wyw-in-js/pull/147), but it remains unmerged.

This PR adds a CSS comment to all currently empty classes/styled elements. This appears to avoid the build issues, though given the build issues are intermittent anyway, it's hard to know if this will stand up over time.

PR #669, opened 2025-08-07
fix: `Menu.AnchorItem` displays selected state correctly

kurtdoherty:
`Menu.AnchorItem`'s selected state relies on `[aria-current='page']:enabled` as the selector, but `:enabled` will never match an anchor element because they cannot be disabled. Instead, the selector should be `[aria-current='page']:not[aria-disabled='true']`.

Also downgrades wyw-in-js to 0.6.0 because builds were continuing to intermittently fail, even after #647.


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/baac1445-3fe8-417c-b705-41f3d1fdcae0?refs=d13a0985d11b94a4cf8dafbf8aa493a62e1a7d8e&source=prChecksComment)

src/core/menu/item/styles.ts
```diff
@@ -40,7 +40,7 @@ export const elMenuItem = css`
   }
 
   &[aria-checked='true']:enabled,
```
kurtdoherty, line 42: **note:** anchor element's can't be `disabled`, so they will never match `:enabled`. They can be `aria-disabled` though, so we rely on that instead.

PR #698, opened 2025-08-21
feat: #496 Added Search Input v5 element

DanishAli-Reapit:
# Pull request checklist

fixes: #496 

- Added v5 SearchInput element into Lab folder to re-rendered by @kurtdoherty when he pick this ticket again

https://github.com/user-attachments/assets/8ff6da36-0c7b-4b1f-a0e0-04ddb66d6e00


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/e7357bb6-e6d6-4715-afb9-4880c1044940?refs=cc37e6744369706bc0c91f48f4b245502dad0cff&source=prChecksComment)

kurtdoherty:
@DanishAli-Reapit can you please include some screenshots of the component/storybook docs for reference in the PR? 🙏

src/lab/search-input/__tests__/search-input.test.tsx
```diff
@@ -0,0 +1,58 @@
+import { render, screen, fireEvent } from '@testing-library/react'
+import { SearchInput } from '../index'
+
+describe('SearchInput', () => {
+  test('should match snapshot (default)', () => {
+    const { asFragment } = render(<SearchInput />)
```
kurtdoherty, line 6: **suggestion:** I generally prefer to avoid snapshots these days in Elements and instead use React Testing Library to assert the DOM has what we expect, similar to what you're doing in the later unit tests; e.g. `expect(screen.getByRole('input', { name: '...' })).toBeVisible()` and so on.

Snapshots like what you have here capture information that we don't care about in the context of the `SearchInput` component (e.g. changes to the Icon component's SVG attributes). By capturing this kind of detail in our `SearchInput` unit tests, they may fail when internal changes are made to the icon component, which gives us false signals that something is wrong. See our docs on [Writing tests that fail well](https://consolegroup.atlassian.net/wiki/spaces/DP/pages/2678587420/Writing+good+unit+tests#Designing-tests-that-fail-well).

To be clear though, this is just a suggestion, not a blocker. We can improve the tests later if need be.

src/lab/search-input/search-input.tsx
```diff
@@ -0,0 +1,67 @@
+import { useState, FC, ChangeEvent, InputHTMLAttributes } from 'react'
+import { ELSearchInput, ELInput } from './styles'
+import { SearchIcon } from '#src/icons/search'
+import { CloseIcon } from '#src/icons/close'
+import { Button } from '#src/core/button'
+
+/**
+ * Props for the SearchInput component.
+ *
```
kurtdoherty, line 9: **todo:** please move these JS Doc comments for each prop to the actual property in the interface.

e.g.
```ts
interface MyInterface {
  /** This JS Doc will describe the `foo` property */
  foo: string
}
```

src/lab/search-input/search-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const handleClear = () => {
+    setValue('')
+    if (onSearch) onSearch('')
+  }
+
+  return (
+    <ELSearchInput data-size={inputSize} aria-disabled={isDisabled}>
+      {!value && <SearchIcon size="sm" color="primary" />}
+      <ELInput
```
kurtdoherty, line 54: **suggestion:** I think we should probably use a `type="search"` input (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/search)). This will likely introduce additional user agent styles though that would need to be worked around, hence why this is a suggestion and not a todo.

src/lab/search-input/styles.ts
```diff
@@ -0,0 +1,81 @@
+import { font } from '#src/core/text/index'
+import { styled } from '@linaria/react'
+
+export const ELInput = styled.input`
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 1;
```
kurtdoherty, line 7: **question:** why are do you need to use `display: -webkit-box` etc?

src/lab/search-input/styles.ts
```diff
@@ -0,0 +1,81 @@
...
+  flex: 1 0 0;
+  border: none;
+  outline: none;
+  background: transparent;
+  overflow: hidden;
+  color: var(--comp-input-colour-text-default-input);
+  text-overflow: ellipsis;
```
kurtdoherty, line 14: **question:** I don't see anything in the design spec that says we should be truncating text... am I missing something?

src/lab/search-input/styles.ts
```diff
@@ -0,0 +1,81 @@
+import { font } from '#src/core/text/index'
+import { styled } from '@linaria/react'
+
+export const ELInput = styled.input`
```
kurtdoherty, line 4: **todo:** `ELInput` -> `ElInput`. The "L" should be lowercase...

src/lab/search-input/search-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * A controlled search input component with:
+ * - A leading search icon (visible when the field is empty).
+ * - A clear button (visible when the field has a value).
+ * - Support for custom sizes and disabled state.
+ */
+export const SearchInput: FC<SearchInputProps> = ({ inputSize = 'medium', onSearch, isDisabled = false, ...rest }) => {
+  const [value, setValue] = useState('')
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    if (isDisabled) {
```
kurtdoherty, line 36: **question:** why are we doing this when the input is disabled? Disabled inputs won't fire change events...

src/lab/search-input/search-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * A controlled search input component with:
+ * - A leading search icon (visible when the field is empty).
+ * - A clear button (visible when the field has a value).
+ * - Support for custom sizes and disabled state.
+ */
+export const SearchInput: FC<SearchInputProps> = ({ inputSize = 'medium', onSearch, isDisabled = false, ...rest }) => {
+  const [value, setValue] = useState('')
```
kurtdoherty, line 33: **question:** why are we managing the value's state internally like this?

kurtdoherty (commented):
Thanks @DanishAli-Reapit. There's a few suggestions, but mostly interested in the questions I've asked 👍

src/lab/search-input/styles.ts
DanishAli-Reapit: Updated, thanks

src/lab/search-input/styles.ts
DanishAli-Reapit: It's there in the Figma.

image

src/lab/search-input/search-input.tsx
DanishAli-Reapit: Added and used the CSS to hide it.

```
&[type='search']::-webkit-search-cancel-button {
  -webkit-appearance: none;
  appearance: none;
}
```

src/lab/search-input/search-input.tsx
kurtdoherty: `useRef` + `mergeRefs` is one option, yes.

But that won't necessarily resolve the main problem I see with your current implementation: currently, you're forcing consumers to rely on `onSearch` if they want to control the input's value and still have the built-in clear button function correctly (because clearing only calls `onSearch`, it doesn't fire an `onChange`). A lot of form libraries like React Hook Form and Formik tend to rely on `onChange`, so I'm not sure it's appropriate for us to rely on `onSearch` the way you currently are.

src/lab/search-input/styles.ts
kurtdoherty: Ah, yeah I think that's just a mistake in Figma. We don't want to be visually truncating input text.

src/lab/search-input/styles.ts
kurtdoherty: Again, I think this is just a mistake in Figma. We don't want to be visually truncating input text.

src/lab/search-input/search-input.tsx
kurtdoherty: In the spirit of "experimental", you could proceed with the current implementation if you need to move on, I would just rename `onSearch` to `unstable_onSearch` and clearly document in the prop's JS Doc that it will likely be removed when the component is put in core.

kurtdoherty (approved):

src/lab/search-input/search-input.tsx
kurtdoherty: @DanishAli-Reapit I think you missed my request in the comment above about renaming `onSearch` and documenting it will likely be removed in future.

PR #708, opened 2025-08-25
fix: #704 `Accordion` disclosure marker now hidden, custom icon now r…

### LINKED ISSUE #704: bug: v5 Accordion component issue on IOS
**Describe the bug**
I've tried to use the new v5 Accordion component and noticed additional left triangle icon, and the right arrow always showing collapsed state despite the accordion already expanded

**Expected behaviour**
the left triangle icon to not shown, and the button-icon to show expand/collapse state correctly

**Screenshots**

Image

**Additional context**

tested on IOS Chrome & Safari 15

kurtdoherty:
Fixes #704

- `:open` pseudo-class is not supported by Safari, so we need to use a forgiving selector like `:is`.
- `::marker` is styled with `display: none`. Typically, the marker is hidden automatically when the summary element has a non-list-item display mode, but this doesn't seem to be the case for older versions of Safari.

### DIFF
--- a/src/core/accordion/summary/styles.ts
+++ b/src/core/accordion/summary/styles.ts
@@ -20,6 +20,11 @@ export const ElAccordionSummary = styled.summary`
     outline: var(--border-width-double) solid var(--colour-border-focus);
     outline-offset: var(--border-width-default);
   }
+
+  /* Ensure the user-agent disclosure marker is hidden */
+  &::marker {
+    display: none;
+  }
 `
 
 export const ElAccordionSummaryTitle = styled.div`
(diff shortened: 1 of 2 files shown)


kurtdoherty:
**note:** I'm unable to test Safari 15 via the Xcode simulator, so unsure if the `::marker` changes in this PR actually result in them being correctly hidden. Generally speaking, Safari 15 does not have wide usage anymore so it's not a key area of support for us (Safari Mobile 15 ~0.33% of market according to https://www.stetic.com/market-share/browser/)

PR #715, opened 2025-08-26
feat: Add new `TableRowPrimaryAction`

kurtdoherty:
### Context

  - Part 1: Primary action for table rows (this PR)
  - Part 2: Primary data for table body cells
  - Part 3: Double-line layout for cells
  - Part 4: Body cell
  - Part 5: Body row
  - Part 6: Table body

### This PR

- Adds `TableRowPrimaryAction` and `TableRowPrimaryActionButton`.

**note:** These components render as an anchor and button respectively, but ones whose "hit area" will expand to fill the bounding box of the closest, relatively-positioned ancestor, which, by default, will be the table row.

Screenshot 2025-08-27 at 9 37 54 am
Screenshot 2025-08-27 at 9 37 59 am

### DIFF
--- a/src/core/table/primary-action/styles.ts
+++ b/src/core/table/primary-action/styles.ts
@@ -1,0 +1,30 @@
+import { css } from '@linaria/core'
+import { font } from '#src/core/text'
+
+export const elTableRowPrimaryAction = css`
+  display: inline-grid;
+  align-items: center;
+  border: none;
+  margin: 0;
+  padding: 0;
+
+  ${font('sm', 'medium')}
+
+  background: transparent;
+  color: var(--colour-text-primary);
+  text-decoration: none;
+
+  outline: none;
+
+  &::after {
+    content: '';
+    display: block;
+    position: absolute;
+    inset: 0 0 0 0;
+  }
+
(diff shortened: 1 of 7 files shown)


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/f5a3e4a0-30ae-431b-b8b9-12366607391b?refs=6871de3ba61860262d26f759cf11a4ff15ef2cd6&source=prChecksComment)

src/core/table/primary-action/styles.ts
```diff
@@ -0,0 +1,30 @@
+import { css } from '@linaria/core'
+import { font } from '#src/core/text'
+
+export const elTableRowPrimaryAction = css`
+  display: inline-grid;
+  align-items: center;
+  border: none;
+  margin: 0;
+  padding: 0;
+
```
kurtdoherty, line 10: **note:** the default font style is the "primary" styled outlined in the design spec. Since this component is for the "primary" action of the row, it should always be used in the row header cell, which is mean to use the medium weight font.

PR #716, opened 2025-08-27
feat: Add new `TableCellPrimaryData`

kurtdoherty:
### Context

- We have table atoms in `@reapit/elements/lab/table`, but these don't facilitate the level of nuance required by the DS. For example:
  - They are pinned to semantic table elements like `td`, `tr` and so on, but we may need the flexibility of using them in a div-based DOM structure instead.
  - They do not provide any solution for the primary row action (which is meant to _appear_ like the row itself is interactive).
  - They do not provide any solution for rendering row header cells.
  - They require column widths to be defined at the cell-level of the table rather than once at the table-level.
- We want to enhance the capability of our table atoms when implementing "official" core versions of them (i.e. the table atoms that will be available via `@reapit/elements/core/table`.
- The first chunk of this work will be done over a number of PRs and be focused on atoms involved in the table's body:
  - Part 1: #715 
  - Part 2: Primary data for table body cells (this PR)
  - Part 3: Double-line layout for cells
  - Part 4: Body cell
  - Part 5: Body row
  - Part 6: Table body

### This PR

- Adds `TableRowPrimaryData`. This component assists with the layout concerns shared by single-line cells and the first row of double-line cells.

Screenshot 2025-08-27 at 10 19 49 am
Screenshot 2025-08-27 at 10 19 55 am
Screenshot 2025-08-27 at 10 20 00 am

PR #719, opened 2025-08-27
feat: Add new `TableBodyRow`

kurtdoherty:
### Context

  - Part 1: #715 
  - Part 2: #716 
  - Part 3: #717 
  - Part 4: #718 
  - Part 5: Body row (this PR)
  - Part 6: Table body

### This PR

- Adds `TableBodyRow`. This component handles the layout requirements of a row within the table's body (classically, the `<tr>` element).
- Updates `TableBodyCell` to ensure its content is vertically centred.
- Updates `TableRowPrimaryAction` to ensure its focus outline is layered above subsequent sibling rows.

Screenshot 2025-08-27 at 10 20 15 pm
Screenshot 2025-08-27 at 10 20 22 pm
Screenshot 2025-08-27 at 10 20 27 pm

### DIFF
--- a/src/core/table/body-cell/styles.ts
+++ b/src/core/table/body-cell/styles.ts
@@ -6,6 +6,8 @@ import { font } from '#src/core/text'
 // supported by the TableBodyCell component.
 export const elTableBodyCell = css`
   display: grid;
+  align-items: center;
+  justify-content: inherit;
 
   ${font('sm', 'regular')}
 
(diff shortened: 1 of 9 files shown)


src/core/table/body-cell/styles.ts
```diff
@@ -6,6 +6,8 @@ import { font } from '#src/core/text'
 // supported by the TableBodyCell component.
 export const elTableBodyCell = css`
   display: grid;
```
kurtdoherty, line 8: **note:** `align-items` was missed in #718

PR #721, opened 2025-08-28
feat: Add new `TableRowMoreActions`

kurtdoherty:
### Context

  - Part 1: #715 
  - Part 2: #716 
  - Part 3: #717 
  - Part 4: #718 
  - Part 5: #719 
  - Part 6: #720 
  - Part 7: More actions menu (this PR)

### This PR

- Adds `TableRowMoreActions`. This component implements the custom "more actions" button and related menu for use in table rows.
- Updates `TableBodyCell` to handle block padding instead of `TableBodyRow` so that the focus outline of interactive elements within the cell (like `TableRowMoreActions`) doesn't get clipped/hidden due to the table cell's overflow handling.

Screenshot 2025-08-28 at 10 48 48 am
Screenshot 2025-08-28 at 10 48 55 am
Screenshot 2025-08-28 at 10 48 59 am


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/dbc6083e-7f27-4af1-b33f-fa4b1ab43df8?refs=6fafe1e7a4444c75df251b2f51edd919c9b71ded&source=prChecksComment)

src/core/table/body-row/body-row.stories.tsx
```diff
@@ -32,6 +34,12 @@ const meta = {
             <TableBodyCell>10 Hay St, Melbourne 3100</TableBodyCell>
             <TableBodyCell>Data</TableBodyCell>
             <TableBodyCell>Data</TableBodyCell>
```
kurtdoherty, line 36: **note:** updating some stories to demo the more actions component

src/core/table/primary-action/styles.ts
```diff
@@ -20,13 +21,13 @@ export const elTableRowPrimaryAction = css`
     content: '';
     display: block;
     position: absolute;
```
kurtdoherty, line 23: **note:** scouting a stylelint issue...not sure why it only just surfaced now.

src/core/table/constants.ts
```diff
@@ -0,0 +1,9 @@
+/** The z-index of the table row's primary action. */
```
kurtdoherty, line 1: **note:** added these new constants for the z-index values.

PR #725, opened 2025-08-28
feat: Add new `TableCellSortButton`

kurtdoherty:
### Context

- The first chunk of this work focused on atoms involved in the table's body and was completed over a number of PRs: #715, #716, #717, #718, #719 and #720.
  - Part 1: Sort button (this PR)
  - Part 2: Header cell
  - Part 3: Header row
  - Part 4: Table head

### This PR

- Adds `TableCellSortButton`.

Screenshot 2025-08-28 at 10 59 23 pm
Screenshot 2025-08-28 at 10 59 30 pm
Screenshot 2025-08-28 at 10 59 36 pm

### DIFF
--- a/src/core/table/sort-button/__tests__/sort-button.test.tsx
+++ b/src/core/table/sort-button/__tests__/sort-button.test.tsx
@@ -1,0 +1,47 @@
+import { render, screen } from '@testing-library/react'
+import { TableCellSortButton } from '../sort-button'
+
+test('renders as a button element', () => {
+  render(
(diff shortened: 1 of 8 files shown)


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/53ffdd21-9a05-463e-9798-57b4016d266c?refs=e8124463e6e8f38b36501770fbd970a3e4a244ac&source=prChecksComment)

PR #726, opened 2025-08-28
feat: Add new `TableHeaderCell`

kurtdoherty:
### Context

- The first chunk of this work focused on atoms involved in the table's body and was completed over a number of PRs: #715, #716, #717, #718, #719, #720 and #721.
  - Part 1: #725 
  - Part 2: Header cell (this PR)
  - Part 3: Header row
  - Part 4: Table head

### This PR

- Adds `TableHeaderCell`.

Screenshot 2025-08-29 at 8 36 54 am
Screenshot 2025-08-29 at 8 37 01 am
Screenshot 2025-08-29 at 8 37 05 am

### DIFF
--- a/src/core/table/__story__/table-wrapper.tsx
+++ b/src/core/table/__story__/table-wrapper.tsx
@@ -17,6 +17,12 @@ export function buildTableWrapper(placement: ChildPlacement): FC<{ children: Rea
       width: '100%',
     }
 
+    const subgridStyles: CSSProperties = {
+      display: 'grid',
+      gridColumn: '1 / -1',
+      gridTemplateColumns: 'subgrid',
+    }
+
     switch (placement) {
       case 'body':
         return <table style={tableStyle}>{children}</table>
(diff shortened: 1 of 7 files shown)


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/f018cea6-616f-4f12-91c5-1d0e5edf25f9?refs=043f5dcff85585a19d5302726d63b1a34656046f&source=prChecksComment)

src/core/table/__story__/table-wrapper.tsx
```diff
@@ -17,6 +17,12 @@ export function buildTableWrapper(placement: ChildPlacement): FC<{ children: Rea
       width: '100%',
     }
```
kurtdoherty, line 19: **note:** need these temporarily while the proper header row and table head components are built

PR #727, opened 2025-08-28
feat: Add new `TableHeaderRow`

kurtdoherty:
### Context

- The first chunk of this work focused on atoms involved in the table's body and was completed over a number of PRs: #715, #716, #717, #718, #719, #720 and #721.
  - Part 1: #725 
  - Part 2: #726 
  - Part 3: Header row (this PR)
  - Part 4: Table head

### This PR

- Adds `TableHeaderRow`.

Screenshot 2025-08-29 at 9 10 15 am
Screenshot 2025-08-29 at 9 10 20 am

### DIFF
--- a/src/core/table/__story__/table-wrapper.tsx
+++ b/src/core/table/__story__/table-wrapper.tsx
@@ -1,5 +1,6 @@
 import { TableBody } from '../body'
 import { TableBodyRow } from '../body-row'
+import { TableHeaderRow } from '../header-row'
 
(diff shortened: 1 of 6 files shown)


rpt-uk-github:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/df26dbf8-1ae4-4176-9aa4-d29945459786?refs=23ee088ed31e311f02e40388529fa6a7db865e5c&source=prChecksComment)

PR #795, opened 2025-09-18
chore: Pin dependencies and use explicit immutable installs

kurtdoherty:


.github/workflows/playwright.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           node-version: '22.x'
 
       - name: Install dependencies
```
kurtdoherty, line 39: **note:** while yarn defaults to immutable installs in CI, we want to be explicit about it.

